PR #913, opened 2015-01-28
 CORS in rest api

ghosthamlet:
add CORS header in rest api header, enable in site config file

### DIFF
--- a/doc/manuals/site-anatomy.rst
+++ b/doc/manuals/site-anatomy.rst
@@ -166,6 +166,29 @@ The following options can be configured:
   then those menus and modules will not be installed unless the new module
   performs those operations.
 
+``{service_api_cors, false}``
+  default is disabled, for cross domain api, see more detail: http://www.html5rocks.com/en/tutorials/cors/
+
+``{'Access-Control-Allow-Origin', "*"}``
+  default is all origin. meaningful if service_api_cors is true, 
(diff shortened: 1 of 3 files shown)


mworrell:
@mmzeeman can you have a look at this one please?

mworrell:
We are checking if this shouldn't be part of webzmachine, as there are also CORS request headers.

Url for some documentation/reading http://www.html5rocks.com/en/tutorials/cors/

Adding @arjan to the discussion.

arjan:
Sorry for the very late reply. This patch looks reasonable to me like this.

Two comments:
* Rename the option to `service_api_cors`
* It would be nice to have the `{service_api_cors, true}` option documented in [site-anatomy.rst](https://github.com/zotonic/zotonic/blob/5b3bdad831a88031badcb53fc847fe8c1a3f10a8/doc/manuals/site-anatomy.rst site-anatomy.rst), so it is clear which headers are set by activating this.

* And maybe for later: make each header configurable (allowed hosts, request methods, etc)

ghosthamlet:
i have renamed the option, added document to site-anatomy.rst,  but your link is broken. 
the other header configuration (allowed hosts etc) i think we can just expose to user with original header name and value, as there are many options there.    many users may use the service_api_cors only, but few advanced users may set the others, they know what they do, so i think it aslo not very necessary to validate the others value in set_cors_header.

arjan:
@ghosthamlet I finally got around to it becase I needed it for my own project. I did some small changes but overall it looked fine. Thanks for the contribution!

PR #921, opened 2015-02-27
Allow configuration of db creation and installation

CyBeRoni:
I'd much prefer if Zotonic left my database alone if it's not to its liking, this allows to set that so by setting dbcreate and dbinstall to false in the global config file. Defaults are to both create and install whatever needed.

### DIFF
--- a/priv/zotonic.config.in
+++ b/priv/zotonic.config.in
@@ -39,6 +39,12 @@
    {dbport, 5432},
    {dbhost, "localhost"},
 
+%%% By default, Zotonic will create a postgres database for you if it doesn't
+%%% already exist, and install tables in it. Uncomment the options below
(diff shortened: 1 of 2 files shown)


mworrell:
Good idea!  I've added a small comment, as it might be good to add a warning that the db/schema is not present and the system is not going to create it.

arjan:
Once these options are documented (in the default `zotonic.config` file, as commented out options), this is good to merge.

CyBeRoni:
Added.

CyBeRoni:
Also the warning.

mworrell:
Looks good now, merging...  thanks Marco!

mworrell:
Also merging into 0.12.x

PR #932, opened 2015-03-28
Change property from tooltip to title.

imagency:


mworrell:
I am a bit hesitant to merge this as the `codemirror` plugin is maintained elsewhere. So if we update form upstream we might loose this change again.

See also: https://github.com/codemirror/codemirror

@arjan

arjan:
I agree, this change belongs upstream. Please do a PR there, and we'll integrate it once it's merged upstream.

imagency:
Hello Marc, hello Argan!
It seems you are right, I download the latest version of codemirror plugin for tinyMCE from here:
http://www.avoid.org/codemirror-for-tinymce4/
and there is no such error.
May be it is time to refresh this plugin in zotonic project?

PR #933, opened 2015-04-01
Fix page block links

ddeboer:
I can remember that this worked before, but I don’t understand how. Anyway, this seems to fix the old behaviour.


mworrell:
I guess there used to be some javascript catching these '#<number>' links.

Thanks for the fix.

PR #936, opened 2015-04-04
Allow to override pidfile location with env var

CyBeRoni:
So I can put i in /run where it belongs


mworrell:
Looks good, and good reason to make it configurable.

CyBeRoni:
Can this be backported to 0.12.x too?

mworrell:
Done!

PR #942, opened 2015-04-20
Decode JSON body in service calls

ddeboer:
This is a proposal to automatically decode JSON content bodies in service API calls.

### DIFF
--- a/modules/mod_base/controllers/controller_api.erl
+++ b/modules/mod_base/controllers/controller_api.erl
@@ -179,14 +179,59 @@ to_json(ReqData, Context) ->
 
 
 process_post(ReqData, Context) ->
-    Module = z_context:get(service_module, Context),
-    case Module:process_post(ReqData, Context) of
-        ok ->
-            {true, ReqData, Context};
-        Result ->
-            api_result(ReqData, Context, Result)
+    case handle_json_request(ReqData, Context) of
+        {error, _Reason} ->
+            api_result(ReqData, Context, {error, syntax, "invalid JSON in request body"});
+        {ok, Context1} ->
+            Module = z_context:get(service_module, Context),
+            case Module:process_post(ReqData, Context1) of
+                ok ->
+                    {true, ReqData, Context1};
+                Result ->
+                    api_result(ReqData, Context1, Result)
+            end
+    end.
+    
+%% @doc Handle JSON request bodies.
+-spec handle_json_request(#wm_reqdata{}, #context{}) -> {ok, #context{}} | {error, string()}.
+handle_json_request(ReqData, Context) ->
+    case wrq:get_req_header("content-type", ReqData) of
+        "application/json" ++ _ ->
+            decode_json_body(ReqData, Context);
+        _ ->
+            {ok, Context}
     end.
 
+%% @doc Decode JSON request body.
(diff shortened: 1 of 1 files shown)


modules/mod_base/controllers/controller_api.erl
```diff
@@ -180,11 +180,28 @@ to_json(ReqData, Context) ->
...
+    
+    %% If Content-Type is application/json, try to decode content body.
+    Context1 = case wrq:get_req_header("content-type", ReqData) of
+        "application/json" ->
+            {ReqBody, _RD} = wrq:req_body(ReqData),
+            {struct, Body} = mochijson2:decode(ReqBody),
+            
+            lists:foldl(
+                fun({Key, Value}, Context2) ->
+                    z_context:set_q(z_convert:to_list(Key), Value, Context2)
```
ddeboer: Does it make sense to use `set_q` for this? My reason is that it then doesn’t matter whether a client submits form data or a JSON content body, as handling either of those responses will be (almost) the same.

ddeboer:
@arjan Quick tip: comment on the PR (under ‘Files changed’) instead of individual commits that may change because of squashing etc.

modules/mod_base/controllers/controller_api.erl
```diff
@@ -180,11 +180,28 @@ to_json(ReqData, Context) ->
 
 process_post(ReqData, Context) ->
     Module = z_context:get(service_module, Context),
-    case Module:process_post(ReqData, Context) of
+    
+    %% If Content-Type is application/json, try to decode content body.
+    Context1 = case wrq:get_req_header("content-type", ReqData) of
+        "application/json" ++ _ ->
```
ddeboer: Changed to support UTF-8.

ddeboer:
> What about arrays? Those will decode as a list, not as a tuple.

Good question: which `set_q` key should we use for lists? Should we either skip them or set some generic `body` key?

mworrell:
Maybe just check if the decoded value is a `{struct, ...}` and if so, then set the `q` arguments?

Maybe we should modify webzmachine - so that `content_types_accepted` is also called for 'normal' POST requests.

ddeboer:
Now only setting params if the JSON body is an object; squashed commits.

> Maybe we should modify webzmachine - so that content_types_accepted is also called for 'normal' POST requests.

In addition to this PR, or do you mean to move the JSON body decoding logic into webzmachine instead?

mworrell:
I meant as part of a more general discussion.

We want to replace webmachine with something else - which is easier to use and works with another http library.

ddeboer:
What still needs to be done here?

ddeboer:
Can this be merged? If not, could you please specify what to you would prefer to see changed here?

witeman:
Why not using jsx(https://github.com/talentdeficit/jsx) instead of mochijson2, it is able to decode json binary into lists or maps.

Best regards,
Witeman

ddeboer:
I was just following the examples in Zotonic itself (such as `controller_api.erl`). 

@mworrell Do you prefer to use jsx instead of mochijson?

arjan:
I don't think we have `jsx` as a dependency yet in Zotonic (at least not directly).
Zotonic use lists everywhere anyway, so this is fine for me. I'll do a final review it tonight and merge it.

ddeboer:
Okay, thanks @arjan!

PR #946, opened 2015-05-03
Fix#891

AlainODea:
- Directly source zotonic_setup in each command script


mworrell:
Thanks!

AlainODea:
My pleasure Marc.  Thank you :)

PR #955, opened 2015-05-27
Fix typo in _block_view_survey_thurstone.tpl

pguyot:
Looks like a copy-paste issue


mworrell:
That is indeed an interesting typo ... thanks for the fix!

PR #956, opened 2015-05-27
Handle &# entities when splitting markers

pguyot:
split_marker/2 previously did not handle &#; entities (typically with a single quote ').
Rewrote the function to properly skip &#.


mworrell:
Nice catch :+1:

PR #957, opened 2015-05-28
Fix multi & single HTML structure for thurstone survey element

pguyot:
Judging from Bootstrap documentation, each checkbox line should be in a &lt;div class="checkbox"&gt; and each radio button line should be in a &lt;div class="radio"&gt;.


mworrell:
@arjan to me this patch seems to be correct, can you check as well (as you did the semi-automatic bs3 conversion)

arjan:
Looking good to me :)

PR #958, opened 2015-05-28
Fix modal signup

ddeboer:
Signup didn’t work when using the modal dialog, probably because delegate was not supplied. This PR sets the delegate and makes it configurable.


mworrell:
Looks good to me, merging!

mworrell:
I will merge into `acl-content-groups` as well.  I assume that is the branch you are currently using?

ddeboer:
That’s right, Marc. Thanks!

mworrell:
Merged master and pushed.

PR #959, opened 2015-05-28
Fix new rsc dialog

trigeek38:
Flushed out the ability to save [object, pred] pair when creating new rsc using the dialog_new_rsc action.


trigeek38:
Recalling pull request

PR #960, opened 2015-05-28
Fix dialog_new_rsc action to save [object, predicate] pairs

trigeek38:
This allows users to pass objects variable to dialog_new_rsc action to be able to save object, predicate pairs at creation time, such as assigning current user as author.


mworrell:
I think we could also use `get_all_values` to get all the objects, but then the other templates already assume a single `objects` variable, so this fits to the current implementation. So that could be something for the future.

Thanks!

PR #961, opened 2015-05-29
mod_admin:  action_admin_dialog_new_rsc fix

trigeek38:
Changed Objects to proplists:get_all_values(object, Args).
This will allow calling the action like below ...

{% button text="New Topic"
    action={ dialog_new_rsc
             cat='topic'
             tabs_enabled = ["new"]
             nocatselect = "true"
             redirect = `topics`
             object = [m.acl.user, `author`]
             object = [421, `subject`]
    }
%}

This also fixes and introduced bug when 'objects' was not presented.  Around line 91 of the action

    [m_edge:insert(Id, Pred, m_rsc:rid(Object, Context), Context) || [Object, Pred] <- Objects]

would crap out and break the stream.  The rsc would save, but the subsequent actions would fail, ie close_dialog etc.


trigeek38:
I searched the modules/*/templates/ directories  and it did not seem like this change will affect any other templates

mworrell:
Looks good, and is consistent with the other arguments.

PR #977, opened 2015-07-01
Allow post-login redirect to be passed to logon modal

ddeboer:
Would be great if this made it into 0.13. :wink:

Didn’t add the param to the `{% with %}` section because it didn’t add much in my opinion. If you prefer, I can add it in.


mworrell:
Will also merge it into 0.13

mworrell:
Cherry-picked into the 0.13

PR #980, opened 2015-07-07
Make module and site includes path-independent

ddeboer:
Previously, `include_lib()` had to point to a (relative) path. With this fix, you can just do `include_lib("module_name/include/include_name.hrl")` and it will work no matter where your sites_dir and modules_dir are located.


mworrell:
:+1:

PR #981, opened 2015-07-07
Added dutch translations

fredpook:


mworrell:
Thanks!  Will also cherry-pick into the 0.13.

PR #982, opened 2015-07-07
Make post-logon actions dynamic

ddeboer:
Currently, the redirect after logon is hard-coded. This commit:
* adds a logon_actions notification for the post-logon actions
* defaults to redirect (the current behaviour) for BC

### DIFF
--- a/modules/mod_authentication/controllers/controller_logon.erl
+++ b/modules/mod_authentication/controllers/controller_logon.erl
@@ -26,7 +26,7 @@
 -export([get_rememberme_cookie/1, set_rememberme_cookie/2, reset_rememberme_cookie/1]).
 
 %% Convenience export for other modules.
--export([logon/2, reminder/2, expired/2, reset/2]).
+-export([logon/2, logon/3, reminder/2, expired/2, reset/2]).
 
 %% Convenience export for other auth implementations.
 -export([send_reminder/2, lookup_identities/2]).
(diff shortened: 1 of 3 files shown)


modules/mod_authentication/controllers/controller_logon.erl
```diff
@@ -489,4 +492,13 @@ get_by_reminder_secret(Code, Context) ->
     end.
 
 
+%% @doc Get actions that should be wired after successful logon
+-spec get_post_logon_actions(list(), #context{}) -> list().
+get_post_logon_actions(WireArgs, Context) ->
+    case z_notifier:foldl(#logon_actions{args=WireArgs}, [], Context) of
+        [] ->
+            [{redirect, [{location, cleanup_url(get_ready_page(Context))}]}];
```
ddeboer: Returning the default here instead of in an `observe_logon_actions` because I think if you want custom actions, you probably don’t want the redirect as well (it will be included because of the `foldl`).

modules/mod_authentication/controllers/controller_logon.erl
```diff
@@ -257,7 +261,7 @@ reset(Args, Context) ->
                 undefined ->
                     throw({error, "User does not have an username defined."});
                 Username ->
-                    ContextLoggedon = logon_user(UserId, Context),
+                    ContextLoggedon = logon_user(UserId, [], Context),
```
ddeboer: Should reset also fully support passing `WireArgs`?

modules/mod_authentication/controllers/controller_logon.erl
```diff
@@ -209,14 +209,18 @@ event(#z_msg_v1{data=Data}, Context) when is_list(Data) ->
 
 %%@doc Handle submit data.
 logon(Args, Context) ->
+    logon(Args, [], Context).
+
+-spec logon(list(), list(), #context{}) -> #context{}.
+logon(Args, WireArgs, Context) ->
```
ddeboer: Is there a naming convention for arguments defined for a wire in a template?

mworrell: Mostly we use Params. This is clear enough though :)

mworrell:
Hi David, did you check this?  
I assume you use this as a post action from a custom logon template/dialog.
Did the actions run correctly?
 
I am wondering because of the change of session/page processes during the authentication of a new user. Though thinking of it, the logon form uses a HTTP POST for the postback to allow setting http-only cookies, so the actions in the Context result of this logon action will be transported back on the return trip, together with the new cookies.

ddeboer:
Yes, I did. :wink: You do need to pass any extra params that your to the wire.

So to make this work (have some `action` arg):

```erlang

observe_logon_actions(#logon_actions{args=Args}, Acc, _Context) ->
    Actions = proplists:get_all_values(action, Args),
    case length(Actions) of
        0 -> Acc;
        _ -> [Actions|Acc]
    end.
```

You need to override the default logon wire:

```erlydtl
{% wire id="logon_form" type="submit" postback={logon} delegate="mod_authentication" %}
```

with this:

```erlydtl
{% wire id="logon_form" type="submit" postback={logon action=action} delegate="mod_authentication" %}
```

mworrell:
@mmzeeman @arjan Looks ok to me, are you also ok with this change?

mmzeeman:
Looks fine to me too.

arjan:
Looks great, tested it and it seems to work. Merged and backported!

arjan:
b5542a663099a2ae14792707658dda0ba89fc8e4

PR #983, opened 2015-07-08
Add knightlab audio plugin to the sanitize whitelist

fredpook:
The inline audio player of http://soundcite.knightlab.com/ is popular with our clients. It actually is pretty cool and it's from the Northwestern University who's timeline js we also use. So I think we can whitelist it.


arjan:
done!

PR #990, opened 2015-07-16
Fix opening admin dialog by removing call to dialogReposition()

ddeboer:
Open admin (connection) dialogs now throws a JS error. This is because `dialogReposition()` was removed from the public API in ad6ff53. Anyway, it does not need to be called anymore, as the dialog is already repositioned when opening it.

/cc @ArthurClemens


fredpook:
:+1:

mworrell:
Good to remove stuff that is not needed (esp. if throwing errors...)

Thanks!

cc @ArthurClemens

mworrell:
Also cherry-picked into 0.13.x

PR #1018, opened 2015-09-07
Added support for UNIQUE constraint in psql column definitions

millenaarg:
I needed to add a UNIQUE constraint to a column in a table defined with z_db:create_table/3.


arjan:
This looks great! Will be merged later today.

millenaarg:
I just noticed in line 804 the list:flatten was not in my last commit. I'll update because now the unique argument is ignored.

PR #1019, opened 2015-09-08
Added support for UNIQUE constraint in psql column definitions

millenaarg:
The variable for the unique constraint was not used in column_spec in the previous pull request. Fixed.


mworrell:
Thanks!  And sorry for the merge-delay...

PR #1032, opened 2015-09-14
added -noshell flag to zotonic-debug to make it launchd compatible (Fixes zotonic/zotonic#1031)

aseidlitz:
See issue zotonic/zotonic#1031 for details

### DIFF
--- a/src/scripts/zotonic-debug
+++ b/src/scripts/zotonic-debug
@@ -30,4 +30,4 @@ require_zotonic_not_running
 
 cd $ZOTONIC
 
-$ERL -env ERL_MAX_PORTS $MAX_PORTS +P $MAX_PROCESSES +K $KERNEL_POLL -pa $PA $NAME_ARG $NODENAME@$NODEHOST -boot start_sasl $(find_config_arg erlang.config) $(find_config_arg zotonic.config) -sasl errlog_type error -s zotonic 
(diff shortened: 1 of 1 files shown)


mworrell:
I was wondering if the Launchd shouldn't be started with the *start* argument instead of *debug*.  The *start* argument will launch a detached beam process.

aseidlitz:
Yes, processes managed by Launchd must not daemonize (fork, detach, etc.) Here's the reference: https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLaunchdJobs.html#//apple_ref/doc/uid/10000172i-SW7-104738

arjan:
I don't think we should pass `-noshell` to `zotonic debug`, as that defeats its purpose. The idea is to run Zotonic in the foreground for debugging purposes, while providing an erlang REPL. `-noshell` effectively disables the REPL...

A better idea would be to modify the `zotonic start` command to add a `--keep-in-foreground` (`-k`) option.

CyBeRoni:
A non-debug foreground option would also be very useful for Linux with systemd and Solaris SMF.

arjan:
PR's welcome :)

mmzeeman:
Or should we go for something like this? https://github.com/ShoreTel-Inc/erld/wiki

CyBeRoni:
If you are using a model where the server process forks, then those points are absolutely what should be followed (especially the 'don't return until the service has started successfully' part). 

But the more modern service starters (Darwin Launchd, Linux SystemD, Solaris SMF, Ubuntu Upstart) typically prefer to have daemons *not* fork, instead disconnecting stdin/out/err for them (and redirecting the latter two to a log file, often) and allowing for better process supervision.

mmzeeman:
Should we wait for the third edition of "Advanced Programming in the UNIX environment" ;-)

CyBeRoni:
I have an alternative patch for this, see #1080.

mworrell:
Fixed with #1080

PR #1064, opened 2015-10-22
Add modules model docs

ddeboer:


doc/ref/models/model_modules.rst
```diff
@@ -1,4 +1,15 @@
 
 .. include:: meta-modules.rst
 
-.. todo:: Not yet documented.
+Access information about which :ref:`modules <manual-modules>` are installed
+and which ones are active.
+
+To test if a module is activated, for instance mod_signup::
+
+    {% if m.modules.active.mod_signup %}
```
ddeboer: There is also `.enabled`. What, exactly is the difference?

doc/ref/models/model_modules.rst
```diff
@@ -1,4 +1,15 @@
...
+        {# Do things that depend on mod_signup #}
+    {% endif %}
+
+To print a list of all active modules::
+
+    {{ m.modules.all|pprint }}
```
ddeboer: This seems to print modules that have been activated at least once (but may afterwards have been deactivated again). Why is this?

mworrell:
Thanks!

PR #1073, opened 2015-10-30
Typo inside is_allowed check mod_import_csv.erl

loetie:
There is a small typo inside this file which checks if user is allowed to use mod_import_cvs, this should be mod_import_csv.


mworrell:
Thanks!

PR #1082, opened 2015-11-16
Allow overriding the configuration file location

CyBeRoni:
The default locations are not appropriate for all platforms and situations, so allow setting the configuration directory using an environment variable ZOTONIC_CONFIG_DIR. 

Note: if the required files are not available there, Zotonic will fail to start (which is the expected behaviour for me at least; starting with the wrong config may be harmful to your environment.)

### DIFF
--- a/src/scripts/helpers/zotonic_setup
+++ b/src/scripts/helpers/zotonic_setup
@@ -78,7 +78,7 @@ function find_config {
 export -f find_config 
 
 function find_config_arg {
-    [ ! -z "$ZOTONIC_CONFIG_DIR" ] && { echo "-config ${ZOTONIC_CONFIG_DIR}/$file"; return 0; }
+    [ ! -z "$ZOTONIC_CONFIG_DIR" ] && { echo "-config ${ZOTONIC_CONFIG_DIR}/$1"; return 0; }
 
     file=$(find_config $1)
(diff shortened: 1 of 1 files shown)


CyBeRoni:
Yes, I just saw that too.

CyBeRoni:
Ah, I notice I'm clobbering another variable; needs a little more work than a simple one-line fix.

CyBeRoni:
This should do; if you set ZOTONIC_CONFIG_DIR, it takes that at face value and if no config exists; fails. If not, then the previous behaviour is maintained.

CyBeRoni:
Man, with all the weird things Zotonic does to get running there are a lot of different ways to break it. I think I've caught them all now.

src/scripts/helpers/zotonic_setup
```diff
@@ -78,6 +78,8 @@ function find_config {
 export -f find_config 
 
 function find_config_arg {
+    [ ! -z "$ZOTONIC_CONFIG_DIR" ] && { echo "-config ${ZOTONIC_CONFIG_DIR}/$1"; return 0; }
+
```
ddeboer: Why is this needed? The `$ZOTONIC_CONFIG_DIR` check is already in `find_config`, which is called directly below.

PR #1088, opened 2015-11-20
Fix query export

ddeboer:
It seems search params have changed (`query_id` instead of `id`), so exporting a query threw an error.


mworrell:
Thanks!

mworrell:
Also cherry-picked into master

PR #1096, opened 2015-11-27
[WIP] Add support for responsive images through srcset

ddeboer:
This PR adds support for true [responsive images](http://blog.cloudfour.com/responsive-images-101-part-5-sizes/) through HTML’s [img srcset](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-srcset) attribute. With `srcset`, one can specify multiple image variants, leaving it to the browser to select the right size based on the current viewport.

Zotonic already supports multiple image sizes by defining a mediaclass in multiple `mediaclass.config` files (text, phone, tablet, desktop). These are selected, however, based not on the browser’s resolution (as `srcset` is) but on the User-Agent header.

Putting this up as a RFC. To do:

- [x] support `sizes` attribute
- [x] more documentation

### DIFF
--- a/doc/manuals/media-files.rst
+++ b/doc/manuals/media-files.rst
@@ -38,6 +38,27 @@ The image URL will have a checksum embedded in it so that when the
 contents of the media class is changed, all images which use that
 media class will be regenerated to reflect the new media class.
 
+Responsive images
+.................
+
+To provide images in multiple `responsive sizes`_, use the ``srcset`` attribute::
+
+    %% templates/mediaclass.config
+
+    [
+        {"masthead", [
+            {width, 1600},
+            {height, 900},
+            {crop, center},
+            upscale,
+            {quality, 85},
+            {srcset, [
+                {"640w", [{quality, 50}]},
+                {"1200w", [{quality, 100}]}
+            ]},
+            {sizes, "100vw"}
+        ]}
+    ].
 
 ImageMagick conversion options
 ..............................
@@ -82,3 +103,5 @@ selection mechanism.
 
 
 .. seealso:: :ref:`tag-image`, :ref:`tag-media`, :ref:`manual-lookup-system-ua`
+
+_responsive sizes: https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-srcset
--- a/doc/manuals/media-files.rst
+++ b/doc/manuals/media-files.rst
@@ -54,11 +54,34 @@ To provide images in multiple `responsive sizes`_, use the ``srcset`` attribute:
             {quality, 85},
             {srcset, [
                 {"640w", [{quality, 50}]},
-                {"1200w", [{quality, 100}]}
+                {"1200w", []},
+                {"2x", []}
             ]},
             {sizes, "100vw"}
         ]}
     ].
+
+An ``{% image id mediaclass="masthead" %}`` tag in your template will output::
+
+    <img src='image-default.jpg'
+        sizes='100vw'
+        srcset='image-640.jpg 640w, image-1200.jpg 1200w, image-2400.jpg 2x' />
+
+Each ``srcset`` value is either a `width descriptor`_ or a pixel density
+descriptor.
+
+* A width descriptor of ``640w`` will resize the image to a width of 640 pixels.
+* A pixel density descriptor of ``2x`` will resize the image to 2 times the
+  original media class width value, so 2 * 1600 = 3200.
+
+By default, each srcset image will inherit all properties from the parent
+media class. So, in the example above, the 640w image will have a reduced
+quality value of 50 while the 1200w image will inherit its parent’s value of 85.
+
+So you can override the automatically determined width of 3200 for the 2x
+descriptor by adding::
+
+    {"2x", [{width, 2000}]}
 
 ImageMagick conversion options
 ..............................
@@ -105,3 +128,4 @@ selection mechanism.
 .. seealso:: :ref:`tag-image`, :ref:`tag-media`, :ref:`manual-lookup-system-ua`
 
 _responsive sizes: https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-srcset
+_width descriptor: https://html.spec.whatwg.org/multipage/embedded-content.html#image-candidate-string
--- a/doc/ref/tags/tag_image.rst
+++ b/doc/ref/tags/tag_image.rst
@@ -37,6 +37,8 @@ The following arguments/filters can be specified:
 |mediaclass          |The media class of the image. See                           |mediaclass="thumb"  |
 |                    |:ref:`manual-media-classes`.                                |                    |
 +--------------------+------------------------------------------------------------+--------------------+
+|srcset              |Provide the image in multiple `responsive sizes`_           |srcset=["150", "450"]|
++--------------------+------------------------------------------------------------+--------------------+
 |background          |The background color for transparent image parts. See       |background="white"  |
 |                    |ImageMagick colors for how to specify the RGB color.        |                    |
 +--------------------+------------------------------------------------------------+--------------------+
@@ -107,3 +109,5 @@ The following arguments/filters can be specified:
 See also :ref:`manual-media-classes` for some options that are only available in `mediaclass` files.
 
 .. seealso:: :ref:`tag-image_url`, :ref:`manual-media-classes` and :ref:`tag-media`.
+
+.. _responsive sizes: https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-srcset
--- a/doc/ref/tags/tag_image.rst
+++ b/doc/ref/tags/tag_image.rst
@@ -107,5 +107,3 @@ The following arguments/filters can be specified:
 See also :ref:`manual-media-classes` for some options that are only available in `mediaclass` files.
 
 .. seealso:: :ref:`tag-image_url`, :ref:`manual-media-classes` and :ref:`tag-media`.
-
-.. _responsive sizes: https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-srcset
--- a/doc/ref/tags/tag_image.rst
+++ b/doc/ref/tags/tag_image.rst
@@ -37,8 +37,6 @@ The following arguments/filters can be specified:
 |mediaclass          |The media class of the image. See                           |mediaclass="thumb"  |
 |                    |:ref:`manual-media-classes`.                                |                    |
 +--------------------+------------------------------------------------------------+--------------------+
-|srcset              |Provide the image in multiple `responsive sizes`_           |srcset=["150", "450"]|
-+--------------------+------------------------------------------------------------+--------------------+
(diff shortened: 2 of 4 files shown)


doc/ref/tags/tag_image.rst
```diff
@@ -37,6 +37,8 @@ The following arguments/filters can be specified:
 |mediaclass          |The media class of the image. See                           |mediaclass="thumb"  |
 |                    |:ref:`manual-media-classes`.                                |                    |
 +--------------------+------------------------------------------------------------+--------------------+
+|srcset              |Provide the image in multiple `responsive sizes`_           |srcset=["150", "450"]|
```
ddeboer: Is this list format a good way to provide the arg? Or is something like `srcset=150 srcset=450` preferred?

src/support/z_media_tag.erl
```diff
@@ -199,16 +199,49 @@ tag({filepath, Filename, FilePath}, Options, Context) ->
...
         end.
 
+with_srcset(TagOptions, Filename, Options, Context) ->
+    SrcSet = case proplists:get_value(srcset, Options) of
+        undefined ->
+            case proplists:get_value(mediaclass, Options) of
+                undefined ->
+                    undefined;
+                MediaClass ->
+                    {ok, Props, _Hash} = z_mediaclass:get(MediaClass, Context),
```
ddeboer: Give precedence to `{% image srcset=…` and fall back to any srcset defined in `mediaclass.config`:

```
    {"landscape-large", [
        {width, 750},
        {srcset, [150, 500]}
    ]}
```

mworrell:
I am not sure if this is the best solution. @mmzeeman and I thought that having multiple mediaclasses, one for each size would be better.

I was also thinking how this could be made compatible with the upcoming client-hints https://github.com/igrigorik/http-client-hints

mworrell:
@ArthurClemens do you have any ideas in this area?

ArthurClemens:
Is the idea that all specified image sizes are first generated on the server, and then one picked by the browser based on the viewport or screen resolution?

ddeboer:
>  @mmzeeman and I thought that having multiple mediaclasses, one for each size would be better.

But (as I mentioned above) the ‘sizes’ as they work in Zotonic’s template lookup system are based on the User-Agent header, not the browser’s resolution (or viewport size). 

> Is the idea that all specified image sizes are first generated on the server, and then one picked by the browser based on the viewport or screen resolution?

Exactly: provide multiple images and leave it up to the browser to select the right one based on its viewport. For an introduction, see (e.g.) http://blog.cloudfour.com/responsive-images-101-part-4-srcset-width-descriptors/.

ArthurClemens:
> Exactly: provide multiple images 

That is not a direct answer (the browser part I took for granted). Are the specified image sizes a trigger to generate those image variants on the server?

ddeboer:
> Are the specified image sizes a trigger to generate those image variants on the server?

Based on the `srcset` attribute supplied, I [create a URL](https://github.com/zotonic/zotonic/pull/1096/files#diff-52fe97d5f55be45b863d4bed3b6dfcfaR237) to an image in that size. So yes, adding a size in the mediaconfig or the template will cause an extra image size to be generated on the server.

ArthurClemens:
In that case this is a good enhancement.

About the syntax, why not stay close to the html syntax? For instance `srcset="400w, 800w, 1600w"`, then parse that string to a list of integers. Less burden on the dev writing templates.

mworrell:
There are a couple of things I am wondering and which are holding me back for a merge:

 * Why not use different mediaclasses for the different sizes?  I can imagine that you want different cropping, compression, height etc.

 * Adding an extra argument makes the image url generation slower, most optimal is just a mediaclass (no checksum needs to be calculated)

 * How can we let srcset work together with the HTTP Client Hints? (They are added to Chrome right now, and I think Firefox is working on them)

mworrell:
And of course - the mediaclass format is flexible - we can change/extend it when needed.

ddeboer:
Okay, I now see three alternatives.

1. We add `srcset` support, moving the variants into `mediaclass.config`, making them fully-fledged mediaclass elements (so not just width but other properties like crop are valid as well) and disallow setting `srcset` variants in template files. Something like:

Something like:

```
%% mediaclass.config

    {"landscape-large", [
        {width, 750},
        {srcset, [
            {"landscape-large-s": [
                {width, 150},
                crop

            ],
            {"landscape-large-m": [
                {width, 500}
            ]
        ]}
    ]}
```

Perhaps we want to leave out the mediaclass names (`landscape-large-s`, `landscape-large-m`) and autogenerate them instead.

2. Alternatively, we jump right to HTTP client hints (in the hope that IE and Safari will start [supporting them soon](http://caniuse.com/#feat=client-hints-dpr-width-viewport). When using client hints, the `srcset` [don’t need to be manually set](https://github.com/igrigorik/http-client-hints#device-pixel-ratio-and-viewport-based-selection). So either we still add manual `srcset` support to give template developer full control over which variants (including crop etc.) are offered to clients or we let Zotonic take care of the image negotiation by binning the advertised values (e.g., 100px, 200px, 500px, 1000px).

3. But I would combine the two solutions above:

* Add manual `srcset` variants support (defined by template developer in `mediaclass.config`) with a fallback to defaults that we define in Zotonic.
* When browser supports client hints, pick the right the variant from `mediaclass.config` (falling back to Zotonic defaults) on the server and render that in the `` tag.
* When browser does not support client hints, render all variants in `srcset` so the browser can pick the right variant.

ArthurClemens:
Sounds good to me. I would indeed let the mediaclass names be auto-generated, or leave them out: 

```
{srcset, [
        [
            {width, 150},
            crop
        ],
        [
            {width, 500}
        ]
    ]}
```

mworrell:
Auto-generated sounds good, that might also make it possible to do the reverse lookup from url+size to correct mediaclass. In that way we can support the http client hints.

The mediaclass server and indexer could preprocess all mediaclasses and build the correct lookup structures.

mworrell:
One of the options of srcset is also the pixel density, would be nice to support that as well. It is either a size of a pixel density argument.

Though we could also support the pixel density by naively multiplying the sizes of the image.

mworrell:
Maybe something like:

```
{"somemediaclass", [
    {option, value},
    {srcset, [
         {"2x", [...]},
         {"300w", [...]}
    ]}
]}
```

This would then create variations of the mediaclass to be used in different contexts. From this it is also possible to derive the different `srcset` variations for the template.

For the http-client-hints we can lookup the "best fitting" variant matching the mediaclass and serve that variant.

ArthurClemens:
Any progress on this?

ddeboer:
@ArthurClemens No, I haven’t yet found the time to finish this.

doc/manuals/media-files.rst
```diff
@@ -38,6 +38,27 @@ The image URL will have a checksum embedded in it so that when the
...
+            {width, 1600},
+            {height, 900},
+            {crop, center},
+            upscale,
+            {quality, 85},
+            {srcset, [
+                {"640w", [{quality, 50}]},
+                {"1200w", [{quality, 100}]}
+            ]},
+            {sizes, "100vw"}
```
ddeboer: Are we okay with this syntax?

ddeboer:
Pushed an update:
* now supports the `100w` width as well as the `2x` density syntax
* removed the possibility to set `srcset` in the templates; you can now only do so in the `mediaclass.config`

Still to be done:
* add possibility to override parent mediaclass properties (as shown in the doc section above).

doc/manuals/media-files.rst
```diff
@@ -38,6 +38,50 @@ The image URL will have a checksum embedded in it so that when the
...
+
+Each ``srcset`` value is either a `width descriptor`_ or a pixel density
+descriptor.
+
+* A width descriptor of ``640w`` will resize the image to a width of 640 pixels.
+* A pixel density descriptor of ``2x`` will resize the image to 2 times the
+  original media class width value, so 2 * 1600 = 3200.
+
+By default, each srcset image will inherit all properties from the parent
+media class. So, in the example above, the 640w image will have a reduced
```
ddeboer: Is this inheritance desired behaviour? If the parent defines an `upscale` or `crop` how do you disable that again, if you want to?

ArthurClemens: You would need a bare parent that only implements the common image attributes.
To make this flexible it would be useful to reference other configs:

```
{"masthead", [
    {width, 1600},
    {height, 900},
    {srcset, [
        {"640w", "masthead-small"},
        {"1200w", "masthead-large"},
        {"2x", []}
    ]}
]},
{"masthead-small", [
    {crop, center},
    {quality, 85}
]},
{"masthead-large", [
    {quality, 50}
]}

```

ddeboer: We decided above to have anonymous names for srcset candidates, so perhaps it would be better to inline their props into the parent mediaclass element, as [shown in my example](https://github.com/zotonic/zotonic/pull/1096#discussion-diff-48092781R56). 

@mworrell Is there a way to disable e.g. upscale again when you have enabled it in the parent media class? If not, do you agree on having inheritance from the basic parent?

mworrell:
I kind of like the proposal of @ArthurClemens to reference other mediaclasses in the `srcset`. That would make the mapping from the generated filename to the definition straight forward. Everything gets simpler with such a *redirect* method.

It would also make it possible to reuse the definitions in other places/contexts.

ddeboer:
And do we then still want the inheritance? I think in this solution `masthead-large` should be a full media class definition that is independent from `masthead`.

mworrell:
I am still in dubio about the *reference to other mediaclass* versus the *variation of the current mediaclass*.

The *variation* is easier to understand (especially for someone making templates). Especially with your automatic behavior for `2x` and `1234w` syntaxes.  It is easier to understand what happens.

And we don't need to make it easy on the machine — pre-computing all the variations into special mediaclasses is easy enough to do.

For the question about disabling the `upscale` after it has been enabled. That is not possible right now, but I don't see any problems adding something like `{upscale, false}`

@ArthurClemens What do you think?

ddeboer:
Though I usually favour convention over configuration, we could allow both setting properties and referencing other media classes. So:

* the [docs above](https://github.com/zotonic/zotonic/pull/1096#discussion-diff-48092781R77) still apply: props inherit from parent media class
* alternatively, you can reference another media class.

ArthurClemens:
I think the 2 way approach offers the flexibility for real world problems. The combination of inheritance and composition is similar to [Less extend](http://lesscss.org/features/#extend-feature-extend-inside-ruleset).

ddeboer:
Discussed today with @mworrell:

* We will support both props inheritance and referencing existing media classes from srcset.
* In case props are used, an (implicit) media class is added to the list (e.g., `masthead-640w`).
* `z_mediaclass:reindex_files/2` will fold over user agents and the full list of media classes and look for srcset in them. When expanding the media classes, the referenced media class (implicit or not) is embedded in the `srcset` argument. This speeds up future lookups.

PR #1110, opened 2015-12-16
Added config key for content group when user registers

DorienD:
We have added a config key for content group so we can decide in which content groups new users are placed.

### DIFF
--- a/modules/mod_signup/mod_signup.erl
+++ b/modules/mod_signup/mod_signup.erl
@@ -220,7 +220,7 @@ ensure_identity(Id, {Type, Key, IsUnique, IsVerified}, Context) when is_binary(K
 
 props_to_rsc(Props, IsVerified, Context) ->
     Category = z_convert:to_atom(m_config:get_value(mod_signup, member_category, person, Context)),
-    ContentGroup = z_convert:to_atom(m_config:get_value(mod_signup, content_group, default_content_group, Context)),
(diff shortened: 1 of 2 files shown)


arjan:
It would be nice if this feature were documented (in `doc/ref/modules/mod_signup.rst`), for the rest it looks fine

DorienD:
Done!

mworrell:
Does this work well if the default content group is not present. For example in systems without mod_content_group enabled.

I think it might be better to set the default to `undefined`

modules/mod_signup/mod_signup.erl
```diff
@@ -220,10 +220,12 @@ ensure_identity(Id, {Type, Key, IsUnique, IsVerified}, Context) when is_binary(K
 
 props_to_rsc(Props, IsVerified, Context) ->
     Category = z_convert:to_atom(m_config:get_value(mod_signup, member_category, person, Context)),
+    ContentGroup = z_convert:to_atom(m_config:get_value(mod_signup, content_group, default_content_group, Context)),
```
mworrell: I propose to have `undefined` as the default content group. As that works better in situations where `mod_content_groups` is not enabled.  And `default_content_group` is the default for 'undefined' content groups anyway.

DorienD: Allright! I will change this

mworrell:
Also cherry-picked into the 0.13.

Thanks!

PR #1112, opened 2015-12-17
Fix template not found error

ddeboer:
_signup_config.tpl is part of mod_signup. When that module is disabled, an "[info] include: could not find template "_signup_config.tpl" message is printed. Apparently, even though the include is within an if case that is never reached, Zotonic still tries to find the template. Making the include optional fixes that.


mworrell:
Thanks!

mworrell:
Also merged 1230837 into master

Issue #1084, opened 2015-11-17
m_value should be called more often


Dirklectisch:
The mandatory function for template models m_value should be called more often. More often than not the model record is passed to filters or template tags without conversion. This can be easily reproduced by using the m_rsc model which defines that function as follows:

````erlang
%% @doc Transform a model value so that it can be formatted or piped through filters
-spec m_value(#m{}, #context{}) -> undefined | any().
m_value(#m{value=undefined}, _Context) ->
    undefined;
m_value(#m{value=Id}, Context) ->
    get_visible(Id, Context).
```

I would expect that the function is called when passing #m records to filters or printing from templates. Unfortunately this doesn't work as expected. Here are some examples. 

Printing from a template results in a stringily error:

````
{{ m.rsc[1] }}
````

m_value is not applied when passing through a filter :

````
{{ m.rsc[1]|pprint }}
````

Currently the work around is to just return values instead of records directly from the m_find_value function. This works but isn't ideal if you want to build more advanced models.

ddeboer:
Fixed in #1114.

PR #1114, opened 2015-12-18
mod_base: Issue #1084: Force model queries into values more often

Dirklectisch:
I have added the to_value conversion to several places so that the examples
from issue #1084 work correctly. In addition to that I have added a make_value
filter so that you can not only print the value but also use it in a more
elaborate expression.


mworrell:
Looks good, thanks!

Will add docs for the new filter later.

PR #1115, opened 2015-12-20
add a encoder fun for the external mqtt client

witeman:
there is no encoder fun of the mqtt payload published by pubzub(via mod_mqtt.erl).  There would an error, when an external mqtt client connects to zotonic emqtt broker and subscribes to a topic to which the pubzub published.


mworrell:
@mmzeeman Can you have a look at this one?

PR #1137, opened 2016-01-11
Document catinclude argument for moreresults

### LINKED ISSUE #1053: Document 'catinclude' option for moreresults action.
Hi!

We use the action 'moreresults' to paginate trough our results, but! It does not do a catinclude. To give it some more context:

I have a query page that shows all the events with the template list-item-vertical.event.tpl. Initially i show 6 results and the load more button (action moreresults). I click on the button and it loads 6 more results, but! It loads the list-item-vertical.tpl instead of the list-item-vertical.event.tpl. 

~ Dorien

ddeboer:
Fix #1053.

### DIFF
--- a/doc/ref/actions/action_moreresults.rst
+++ b/doc/ref/actions/action_moreresults.rst
@@ -3,9 +3,9 @@
 
 Show more results of the current search query inline on the page.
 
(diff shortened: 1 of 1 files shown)


mworrell:
Thanks!

PR #1142, opened 2016-01-12
doc: Document release schedule

ddeboer:
As decided today with @arjan @mmzeeman and @mworrell. 

/cc @CyBeRoni


mworrell:
Perfect, as we discussed.

Issue #1143, opened 2016-01-14
Why does zotonic_setup export functions?


pguyot:
zotonic_setup exports all functions available from other scripts.
However, it is sourced from other scripts and therefore functions are available anyway.

Could the `export -f` lines be removed from the script?

The issue with exporting functions is that as a workaround for Shellshock exploit, several operating systems did patch bash (including FreeBSD) to prevent auto-importing of functions, thus generating warnings that eventually break zotonic scripts in specific environment (if LC_\* is set).

This is the output with the export -f:

```
sudo -u zotonic /home/zotonic/zotonic/bin/zotonic shell
bash: Skipping importing function definition for `get_user_sites_dir': --import-functions required.
bash: Skipping importing function definition for `find_enabled_sites': --import-functions required.
bash: Skipping importing function definition for `get_user_modules_dir': --import-functions required.
bash: Skipping importing function definition for `require_zotonic_running': --import-functions required.
bash: Skipping importing function definition for `enablesite': --import-functions required.
bash: Skipping importing function definition for `find_config': --import-functions required.
bash: Skipping importing function definition for `major_version': --import-functions required.
bash: Skipping importing function definition for `require_zotonic_not_running': --import-functions required.
bash: Skipping importing function definition for `find_config_arg': --import-functions required.
sh: environment corrupt; missing value for BASH_FUNC_get_user_modules_dir
sh: environment corrupt; missing value for BASH_FUNC_get_user_modules_dir
pong
Zotonic is not running. You need to start Zotonic first to use this command.
```

Without export -f lines, the script works (zotonic is indeed running).

PR #1144, opened 2016-01-14
Remove unnecessary export -f statements (#1143)

pguyot:
export -f statements seem unnecessary as zotonic_setup is sourced by other scripts, not executed as a child.


mworrell:
@arjan Can you have a look at this?

mworrell:
Rebased in #2156

PR #1145, opened 2016-01-14
Fix issue with ffprobe returning unicode text

pguyot:
ffprobe can return UTF-8 JSON that is then interpreted in os:cmd and returned as a string with unicode characters. z_convert:to_binary is simply iolist_to_binary for lists and does not work with integers > 255.

Not sure z_convert:to_binary/1 should not be fixed instead.


mworrell:
In the master we are working on having real unicode lists and utf-8 binaries, this is a good solution till then.

Thanks!

mworrell:
Also cherry-picked into the 0.13.x

PR #1163, opened 2016-01-25
mod_import_csv: Fix CSV import for CR files

### LINKED ISSUE #1160: Importing CSV file with CR line breaks doesn’t work
When importing a CSV File with CR line breaks (`\r`) all rows are erroneously concatenated into one list so no data gets imported. No error is thrown either. When I convert my file to LF (`\n`) line separators, the import works fine. 

So: should we look for and automatically convert non-LF line separators (also CRLF) to LF or document that only LF line breaks work?

ddeboer:
Fix #1160.


mworrell:
Looks good to me.  Merging.

mworrell:
Cherry-picked a9550a1 into master.

PR #1173, opened 2016-02-01
Support managed ACL rules

### LINKED ISSUE #1146: Manage ACL rules in manage_schema
Especially if you have a large and complex set of ACL rules that must be the same on all environments, managing them through the web interface is not convenient. So @mworrell and I came up with the following.

1. ACL rules are added as a key to `manage_schema`.
2. All managed ACL rules are marked thus in the database tables (a `managed_by=mod_abc` property).
3. When managed ACL rules are loaded, first all existing managed rules are deleted. This makes the set of rules that you describe in `manage_schema`:
  * *declarative*, that is, it describes the desired state of ACL settings.
  * *complete*: if you want to remove an ACL rule, just remove it from `manage_schema`. 
4. All managed ACL rules are greyed out in the web interface and cannot be customised there. This way, they are clearly distinguished from rules that have been added though the web interface.
5. The export/import functionality for ACL rules use Erlang tuples instead of blobs.

To be discussed:

Point 3 above implies a slight inconsistency with the other `manage_schema` types such as resources and categories: you can only *add, not remove* those.

cc @dirklectisch @fredpook

ddeboer:
Fix #1146.

* Disable managed ACL rule editing in admin
* Support both module and rsc rules
* Add documentation

### DIFF
--- a/doc/ref/modules/mod_acl_user_groups.rst
+++ b/doc/ref/modules/mod_acl_user_groups.rst
@@ -8,7 +8,7 @@ This module adds rule based access control.
  * Content groups are arranged in an hierarchy
  * User groups are arranged in an hierarchy
 
-Rules are defined between the user groups and the content groups.
+ACL rules are defined between the user groups and the content groups.
 Per rule the following properties can be set:
 
  * User group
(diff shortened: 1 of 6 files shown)


ddeboer:
Updated to make use of #1179.

doc/ref/modules/mod_acl_user_groups.rst
```diff
@@ -23,3 +23,75 @@ If all is accepted, then the *edit* version van be published.
...
+2. in your site’s or module’s code using
+   :ref:`module versioning <guide-modules-versioning>`_.
+
+While editing a simple set of ACL rules in the web interface is easier for end
+users, developers may prefer to manage more complex rules in code. Code-based
+rules have two important advantages. First, they are equal between all
+environments (such as development, acceptance and production). Secondly, when
+developing and deploying new features, ACL rules and code often belong together.
+By defining the rules in your code, you can commit and store them along with the
+feature code.
```
ddeboer: This is my point of view. Do you agree?

doc/ref/modules/mod_acl_user_groups.rst
```diff
@@ -23,3 +23,75 @@ If all is accepted, then the *edit* version van be published.
...
+
+    manage_schema(install, Context) ->
+        #datamodel{
+            %% your resources...
+            data = [
+                {acl_rules, [
+                    %% A resource ACL rule is defined as {rsc, Properties}.
+                    {rsc, [
+                        {acl_user_group_id, acl_user_group_members},
+                        {actions, [<<"link">>]},
```
ddeboer: It seems that just `[link,view]` doesn’t work.

doc/ref/modules/mod_acl_user_groups.rst
```diff
@@ -23,3 +23,75 @@ If all is accepted, then the *edit* version van be published.
...
+        %% code to upgrade from 1 to 2
+        ok;
+
+Compile the code and restart your module to load the managed rules.
+
+The set of rules defined in ``manage_schema/2`` is *declarative* and *complete*.
+That is to say, you declare the full set of rules that you wish to define. Any
+changes or deletions that you make to the rules in your code, will propagate to
+the site’s rules. The protect the set’s completeness, managed rules cannot be
+altered in the web interface.
```
ddeboer: Is this clear enough?

doc/ref/modules/mod_acl_user_groups.rst
```diff
@@ -23,3 +23,75 @@ If all is accepted, then the *edit* version van be published.
...
+        manage_schema/2
+    ]).
+
+    %% .... more code here...
+
+    manage_schema(install, Context) ->
+        #datamodel{
+            %% your resources...
+            data = [
+                {acl_rules, [
```
ddeboer: Do we agree with this structure?

modules/mod_acl_user_groups/templates/_admin_acl_rule_row.tpl
```diff
@@ -13,6 +13,10 @@
     {% endif %}
     <form id="{{ #ruleform }}" class="row acl-rule-row acl-rule-{{ rule.id }} {% if rule.is_block %}is_block{% endif %}" method="post" action="postback">
 
+        {% if rule.managed_by %}
+            <fieldset disabled="disabled">
+        {% endif %}
```
ddeboer: I just disable the inputs in the interface, which is of course very soft deletion protection. Do we want to harden this in the `m_acl_rule` module?

mworrell:
Also cherry-picked into master

ddeboer:
Yeah, thanks for merging! :wink:

mworrell:
Merge fest :+1:

PR #1179, opened 2016-02-02
Add data model notification

### LINKED ISSUE #1178: Notified datamodel
When working on #1173 I found that we need a separation between `manage_schema` handling in Zotonic’s core on the one hand and modules that can handle certain types of datamodel entries (such as acl rules) on the other.

As discussed with @mworrell today, we propose to:
* extend `#datamodel{}` with an extra record field `notify` or something similar
* `z_datamodel` throws a first notification with the content of the `notify` list as an argument
* modules observe the notification to persist the `notify` datamodel entries.

ddeboer:
Fix #1178.

### DIFF
--- a/include/zotonic_notifications.hrl
+++ b/include/zotonic_notifications.hrl
@@ -642,6 +642,11 @@
         state :: term()
     }).
 
+%% @doc Notify modules of a data model entry defined in manage_schema
+-record(manage_data, {
+    module :: atom(),
+    props :: list()
+}).
 
 % Simple mod_development notifications:
 % development_reload - Reload all template, modules etc
--- a/doc/developer-guide/modules.rst
+++ b/doc/developer-guide/modules.rst
@@ -350,6 +350,14 @@ the first time, no upgrade functions are called, but only the
 ``install`` clause. The upgrade functions exist for migrating old
 data, not for newly installing a module.
 
+Data model notification
+^^^^^^^^^^^^^^^^^^^^^^^
+
+In the ``#datamodel`` record you can manage categories, predicates, resources,
+media and edges. You can also set the ``data`` property, which will send out
+a first ref:`notification <guide-notification>`_. To subscribe to that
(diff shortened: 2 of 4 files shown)


include/zotonic_notifications.hrl
```diff
@@ -642,6 +642,11 @@
         state :: term()
     }).
 
+%% @doc Notify modules of a data model entry defined in manage_schema
+-record(managed_data, {
+    module :: atom(),
+    props :: list()
+}).
```
mworrell: Maybe call it either `datamodel_data` or `manage_data` (more active)?
I like a reference to the `#datamode{}` here, as it makes it easier to understand what this notifications does.

Also, maybe also include the whole `#datamodel{}` here (and datamodel version), as it might be useful for the manage routines to know what is happening.

/cc @arjan  @mmzeeman

ddeboer: Changed to `manage_data`. 

> Also, maybe also include the whole #datamodel{} here (and datamodel version)

You mean the `mod_schema` version? We don’t have that in the place where we send the notification (`z_datamodel`).

ddeboer:
@mworrell wrote:

> I like a reference to the #datamodel{} here, as it makes it easier to understand what this notification does.

I can add it to the notification `@doc`. Is that sufficient?

> Also, maybe also include the whole #datamodel{} here (and datamodel version), as it might be useful for the manage routines to know what is happening.

You mean the `mod_schema` version? We don’t have that in the place where we send the notification (`z_datamodel`).

mworrell:
@ddeboer You can keep it simple for now.

@arjan are you ok with the names of the notifications etc?

ddeboer:
Can we merge this?

mworrell:
:+1: And also into master

mworrell:
Cherry-picked into master.

mworrell:
And thanks again!  :smile:

PR #1184, opened 2016-02-10
Increase resource URI length

### LINKED ISSUE #1156: Make rsc.uri column bigger
Any objections to this? It’s currently 250 chars, but when working with linked data we sometimes encounter longer URIs.

ddeboer:
Fix #1156. This of course works only for newly created sites, but that seems sufficient for now.

See also http://stackoverflow.com/a/417184.

### DIFF
--- a/src/install/z_install.erl
+++ b/src/install/z_install.erl
@@ -103,7 +103,7 @@ model_pgsql() ->
     (
       id serial NOT NULL,
       name character varying(80) NOT NULL DEFAULT ''::character varying,
-      uri character varying(250) NOT NULL DEFAULT ''::character varying,
+      uri character varying(2048) NOT NULL DEFAULT ''::character varying,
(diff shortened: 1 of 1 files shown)


mworrell:
2K is still (just) below the max key size in postgresql B-tree indices. And as most entries will have much shorter urls (or none) I think this is ok.

If many entries will have a longer keys then we might want to add an index using a (shal1) checksum of the uris.

mworrell:
:+1: to merge into 0.13 and master

mworrell:
BTW talking about resource uris. We do sanitize the incoming uris, maybe we also want to normalize them?  (Thinking of upper/lower case, escapes, etc)

ddeboer:
Cherry-picked as 5993714ba870621fbecaf7a6bc31529c4821bcf4 into master.

PR #1186, opened 2016-02-10
mod_admin_identity: Add config param for new user category

### LINKED ISSUE #1154: Make new user category configurable
The category that new users created through Authentication > Users are placed in, should be configurable. It is now [hard-coded](https://github.com/zotonic/zotonic/blob/master/modules/mod_admin_identity/actions/action_admin_identity_dialog_user_add.erl#L65).

Something similar to [mod_signup.member_category](http://zotonic.com/docs/latest/ref/modules/mod_signup.html#config-setting-the-category-for-new-users), but that only works for the signup process.

ddeboer:
Fix #1154.

### DIFF
--- a/doc/ref/modules/mod_admin_identity.rst
+++ b/doc/ref/modules/mod_admin_identity.rst
@@ -1,13 +1,25 @@
 
 .. include:: meta-mod_admin_identity.rst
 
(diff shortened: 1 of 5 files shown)


mworrell:
I also cherry-picked this into master.

PR #1188, opened 2016-02-12
doc: Improve access control chapter

ddeboer:
* Combine authentication and authorization into one chapter


mworrell:
David, can you check the build failures?
Otherwise we just move it to 0.15

PR #1189, opened 2016-02-14
mod_acl_user_groups: Improve wording and add missing Dutch translations 

ddeboer:
* rename 'block' to 'deny'


mworrell:
Looks good.  I am not sure about the "Keep" in the load rules dialog. But if that is confusing then we can always change it back to "Cancel".

mworrell:
Also cherry-picked into 0.13.x

PR #1191, opened 2016-02-16
[zotonic_status] Added the name of the site in the template.

mgpld:
While working with many versions of a site, I found handy to
be able to distinguish which url is which module.
(Nothing prevent the user to configure an already existing url
inside a new site, thus making multiple site with the same url)


arjan:
Thanks, I incorporated it, with a few tweaks.

PR #1192, opened 2016-02-17
Add scripts to do 'z:m()' from the command line; wait for zotonic to be started.

CyBeRoni:
This helps automated deployments; they could be used like zotonic wait && zotonic make.

### DIFF
--- a/bin/zotonic
+++ b/bin/zotonic
@@ -58,6 +58,6 @@ elif [ ! -f "$ZOTONIC_SCRIPTS/zotonic-$1" ]; then
     echo Command not found: $1 1>&2
     exit 127
 else
-    exec $ZOTONIC_SCRIPTS/zotonic-$1 ${@:2}
+    exec $ZOTONIC_SCRIPTS/zotonic-$1 "${@:2}"
 fi
 
--- a/doc/ref/cli/index.rst
+++ b/doc/ref/cli/index.rst
@@ -28,11 +28,17 @@ Currently, the following subcommands are implemented:
 ``zotonic restart``
   Restart the background Zotonic server instance.
 
+``zotonic wait [timeout]``
+  Wait ``timeout`` seconds (default: 30) for Zotonic to be started, then return. 
+
 ``zotonic update``
   Update the server.  Compiles and loads any new code, flushes caches and rescans all modules.
 
 ``zotonic shell``
   Get an EShell on the running instance. See :ref:`guide-cli-shell`.
+
+``zotonic rpc``
+  Send an RPC request to the running Zotonic instance. Example: `zotonic rpc "zotonic ping"`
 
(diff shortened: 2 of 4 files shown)


ddeboer:
Do we need `zotonic-make` if we already have `zotonic-compile`? I prefer to have a separate `zotonic-flush` instead: after all, make = compile + flush.

CyBeRoni:
zotonic-compile runs a function directly in an erlang interpreter, while this zotonic-make tells a running zotonic to do it using rpc. 

I'd prefer to have the flush built-in myself but I could modify it to support an argument and call z:c() in absence of such a thing, which doesn't flush.

arjan:
Why not collapse all of this into one command?
Refactor `zotonic compile` so that it follows the following logic:

 * zotonic not running: perform the "current" `zotonic compile` function
 * zotonic running: perform "wait" and then the RPC call.

I see the case for a separate `zotonic flush` call though.

CyBeRoni:
> Why not collapse all of this into one command?

I'm not personally the largest fan of do-it-all commands. I usually prefer to string smaller things together when automating. You know, the unix way.

CyBeRoni:
That said, the zotonic-make that's in there now could also be renamed zotonic-rpc and just be an interface to call functions directly. Then I would use something like zotonic wait && zotonic rpc "z m"

arjan:
Yeah, that makes more sense to me. Having both a `zotonic compile` and `zotonic make` command makes things very confusing.

CyBeRoni:
David: 0.15? I need this literally two weeks ago.

CyBeRoni:
Changed to make a generic "zotonic rpc" command.

ddeboer:
> David: 0.15? I need this literally two weeks ago.

I understand, but we’re still discussing the best course of action here while the 0.14 release branch was created yesterday.

CyBeRoni:
This PR was made two weeks ago.

arjan:
This looks good, only thing that's missing is documentation.
Please add these two new commands [here](https://github.com/zotonic/zotonic/blob/master/doc%2Fref%2Fcli%2Findex.rst), then I'll merge it.

CyBeRoni:
Added.

ddeboer:
Cool, please squash commits!

arjan:
I'll do it

CyBeRoni:
Don't worry, I just squashed the shit out of this.

arjan:
It's merged now to master but I dont know to which release branch to merge it to.
@ddeboer can you explain?

CyBeRoni:
David tells me 0.13.x will be replaced next week with 0.14 so that will do for me.

ddeboer:
Cherry-picked into 0.14.0. Just because I’m in a good mood. :wink:

arjan:
yay :-)

Issue #1202, opened 2016-02-29
Leap year error during static files request.   


LilDash:
Zotonic crashed with this error:  `{badarg,[{erlang,universaltime_to_localtime,[{{2017,2,29},{2,13,1}}],[]}`

The code :  `{{{Y+1,M,D},HIS}, ReqData, State};` in the controller_file.erl(line 105)  makes leap year error during static files request.

mworrell:
Wow, that its embarrassing. Thanks for the fix!

mworrell:
A hotfix release was made to handle this issue https://github.com/zotonic/zotonic/releases/tag/0.13.8

mworrell:
@LilDash Thanks again for reporting and fixing this.

LilDash:
my pleasure : )

PR #1203, opened 2016-02-29
Fix the leap year error(issue : #1202)

LilDash:
Fix the leap year error during static files request.


mworrell:
We released a hotfix 0.13.8 release for this issue.

See https://github.com/zotonic/zotonic/releases/tag/0.13.8

Issue #1148, opened 2016-01-17
Show 403 page instead of redirect to the login page on access-denied


alvaropag:
Using the option {acl, {edit, my_page_id}} in the dispatch rules causes a redirect to the login page when a user that don't have permission to edit a page try to do it. I'd like to be able to intercept this redirection and show some template/message telling the user he doesn't have permission to execute this action.

Talking with @mworrell via email he suggested 

-------------------------------------------------------------------------------------------

Recently I was also thinking that we should be able to show the same template but then with an error message.

Or at least not change the url.

For other errors we now use a controller_error, which doesn’t change the url and displays a different html page.
On that page we can then just add the “log on” button and some explanation.

Looking at the code the problem is that the z_acl:wm_is_authorized/2 function wants to redirect.
We might want to add an option to throw/return an access denied error - which can then call controller_error.

It is in this code:

> %% @doc Convenience function, check if the current user has enough permissions, if not then
> %% redirect to the logon page.
> -spec wm_is_authorized(boolean() | acl(), #context{}) -> webzmachine:reply().
> wm_is_authorized(true, Context) ->
>     ?WM_REPLY(true, Context);
> wm_is_authorized(false, Context) ->
>     wm_is_authorized(false, logon, Context);
> wm_is_authorized(ACLs, Context) when is_list(ACLs) ->
>     wm_is_authorized(ACLs, logon, Context).

And here:

> wm_is_authorized(false, Redirect, Context) ->
>     ContextLocation = wm_set_location(Redirect, Context),
>     ?WM_REPLY({halt, 302}, ContextLocation);

=============================================================
So, what is the best alternative? Redirect to controller_error? Show the same page with an error? Redirect to a custom template?

mworrell:
I think it should be ok to now *not* redirect but show a 403-like page, with the option to log on. This page can be changed to do the actual redirect.

If Javascript redirects then we can also keep the hash-part of the url in tact during the login (as we don't see the hash-part of the url on the server).

Pulling in some others.

@ddeboer @mmzeeman @arjan

mmzeeman:
Sounds like a plan. When somebody is already logged in it doesn't make sense to show the logon page anyhow.

mworrell:
Indeed, good point. It is quite confusing to show the login page if somebody is already logged on. At least in a template we can tell what is wrong and offer the possibility to go back or log on as a different user.

ddeboer:
:+1: on the fix

PR #1208, opened 2016-03-07
Add 0.14.0 release notes

ddeboer:
@mworrell Any [main changes](https://github.com/zotonic/zotonic/compare/0.13.8...release-0.14.0) that I failed to mention?

### DIFF
--- a/doc/developer-guide/releasenotes/rel_0.14.0.rst
+++ b/doc/developer-guide/releasenotes/rel_0.14.0.rst
@@ -1,0 +1,110 @@
+.. _rel-0.14.0:
+
+Release 0.14.0
+==============
+
+Welcome to Zotonic 0.14.0, released on 7 March, 2016.
+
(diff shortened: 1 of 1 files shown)


mworrell:
Maybe this one, as it changes the behavior:

 * core: show 403 page with option to login instead of redirecting to login controller. Fixes #1148

/cc @mmzeeman

mworrell:
Also good to mention that deps/depcache needs to be updated for the fix *Fixed status page login button.*

ddeboer:
How can devs update `depcache`? Do they need to delete the directory?

ddeboer:
> core: show 403 page with option to login instead of redirecting to login controller. Fixes #1148

Added.

mworrell:
yeah!  :tada: :balloon: :beers:

ddeboer:
Go ahead and Tweet! :wink:

PR #1213, opened 2016-03-11
[WIP] Add support for responsive images through srcset

ddeboer:
Replaces #1096, which was based on 0.13.x.


mworrell:
How much WIP is this?  What needs to be done?

mworrell:
I see that Client Hints never made it past Chrome.  https://caniuse.com/#search=hints

So using `srcset` is the way to go.

mworrell:
Superseded by #2703

PR #1214, opened 2016-03-13
Duplicated Variable 'Module' in controller_api

witeman:
This error would cause the controller_api unable to function.  It complains like below error:
(reason {{case_clause,<<"something">>},[{controller_api,allowed_methods,2,[{file,".../zotonic/modules/mod_base/controllers/controller_api.erl"},{line,65}]}

### DIFF
--- a/modules/mod_base/controllers/controller_api.erl
+++ b/modules/mod_base/controllers/controller_api.erl
@@ -63,7 +63,7 @@ allowed_methods(ReqData, Context) ->
                  undefined ->
                      case z_convert:to_binary(z_context:get(method, Context1)) of
                          <<>> -> Module;
-                         Module1 -> Module1
+                         Mtd -> Mtd
                      end;
                  Mtd -> z_convert:to_binary(Mtd)
              end,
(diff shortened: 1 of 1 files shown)


modules/mod_base/controllers/controller_api.erl
```diff
@@ -63,7 +63,7 @@ allowed_methods(ReqData, Context) ->
                  undefined ->
                      case z_convert:to_binary(z_context:get(method, Context1)) of
                          <<>> -> Module;
-                         Module -> Module
+                         Module1 -> Module1
```
ddeboer: For clarity, I would call this variable `Method1` instead.

witeman: Yes, it should be.  But I am just following the original naming.

mworrell:
Thanks!

mworrell:
Looks like this was already fixed in the 0.x, or it is an error introduced during the *binaryfication*. Thanks for the fix :+1:

witeman:
Welcome

PR #1219, opened 2016-03-22
Added Metric: Number of Open TCP Connections.

tahteche:
Added metric to display the number of open TCP connections in the mod_admin_statistics module.

### DIFF
--- a/modules/mod_admin_statistics/templates/admin_statistics.tpl
+++ b/modules/mod_admin_statistics/templates/admin_statistics.tpl
@@ -11,8 +11,8 @@
     <div class="panel-heading">System</div>
     <div class="panel-body">
         {% for id, name in [
-                ["erlang-stats-system_info-process_count", "Process Count"], 
-                ["erlang-stats-statistics-run_queue", "Run Queue"], 
+                ["erlang-stats-system_info-process_count", "Process Count"],
+                ["erlang-stats-statistics-run_queue", "Run Queue"],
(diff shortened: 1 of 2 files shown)


modules/mod_admin_statistics/templates/admin_statistics.tpl
```diff
@@ -70,7 +75,10 @@ function to_human(value) {
 {% endjavascript %}
 
 {% javascript %}
-    pubzub.subscribe("erlang/stats/#", function(topic, msg) {
+    pubzub.subscribe("erlang/stats/#", function(topic, msg) {   
+        //print topics to log to see all topics available under erlang/stats/#
+        console.log(topic, msg);
```
mworrell: Can you remove this console.log?
On certain browsers (IE) it is not available and will crash the Javascript.

tahteche: Thanks for the heads-up. Its removed.

mworrell: :+1:

arjan:
This looks great to me. @mmzeeman could you also check this out? This is part of @tahteche 's proposal for his GSoC 2016 project.

mmzeeman:
Yes, this is nice work. I have also read the proposal which is also good. I can merge the pull after I'm done with an upgrade which is scheduled around 23:00

Verstuurd vanaf mijn iPhone

mmzeeman:
Thanks :+1:

PR #1222, opened 2016-03-25
change the mqtt app start order

feng19:
after emqtt app started,but the zotonic_sup not start,so the z_sites_dispatcher is not start yet,this time the mqtt client connect to the server and emqtt will report this error:
=ERROR REPORT==== 25-Mar-2016::11:00:37 ===
** Generic server emqtt_auth terminating
** Last message in was {check,<<"333@test">>,<<"test">>}
** When Server state == {state,emqtt_auth_zotonic,[]}
** Reason for termination == 
** {noproc,{gen_server,call,[z_sites_dispatcher,{get_host_for_domain,"test"}]}}

after many time,emqtt app is down

### DIFF
--- a/src/zotonic_app.erl
+++ b/src/zotonic_app.erl
@@ -42,11 +42,12 @@ start(_Type, _StartArgs) ->
     inets:start(httpc,[{profile,zotonic}]),
     zotonic_deps:ensure(),
     ensure_started(mimetypes),
-    ensure_started(emqtt),
     ensure_started(gproc),
     ensure_started(jobs),
     z_tempfile_cleanup:start(),
-    zotonic_sup:start_link().
+    R = zotonic_sup:start_link(),
+    ensure_started(emqtt),
+    R.
 
 %% @spec stop(_State) -> ServerRet
 %% @doc application stop callback for zotonic.


arjan:
Why are you / who is connecting to the emqtt before the startup process is completed?
emqtt is a dependency of Zotonic, not the other way around, so emqtt needs to be started before Zotonic does. But I agree, `emqtt_auth_zotonic` should be made resilient instead of crashing the server.

feng19:
I use app connecting to server, so the connect timeing is out of control. Emqtt is a dependency of zotonic, Yes! It seems to me, zotonic depend on emqtt, beacause if zotonic is not started, emqtt can't use.

arjan:
But this is only temporary, in the first seconds that the zotonic is still starting up, right?

mworrell:
I think we should start emqtt before zotonic, as zotonic will make use of MQTT. Maybe we should delay/queue all calls to a site if zotonic (or the site) is still starting up.

mworrell:
So maybe emqtt_auth should suspend the process till the sites dispatcher (and sites) are up and running.

We have the same problem where the http server is up but the sites aren't up yet. So we might be able to generalize this *suspend till started* problem?

feng19:
Yes! Thank you @arjan  and @mworrell

PR #1225, opened 2016-03-29
Automatically publish managed ACL rules

ddeboer:


ddeboer:
@mworrell As far as I can see, publishing the rules (often) shouldn’t cause any trouble.

mworrell:
Yes, it is ok to publish - though with many content groups we have seen that it can take a bit of time (but that is a different problem which we can tackle separately)

ddeboer:
In that case, feel free to merge this.

mworrell:
Thanks!

PR #1227, opened 2016-03-30
Add a whitelist of IP addresses that can use the default 'admin' password to login into sites.

mworrell:
Often the admin password for development sites is "admin".
It happens quite often that these sites are exposed to the Internet, which risks total access to the server.

This merge request adds a IP address whitelist for IP addresses that are allowed to use the password "admin" for logging into a site using the "admin" account.

There is a new Zotonic configuration in the `zotonic.config`:

```
%%% IP whitelist, used for accessing sites with a default "admin" password
   %% {ip_whitelist, "127.0.0.0/8,10.0.0.0/8,192.168.0.0/16,172.16.0.0/12,::1,fd00::/8"},
```

Per default the local loopback interface and local network addresses are whitelisted.

### DIFF
--- a/doc/ref/configuration/site-configuration.rst
+++ b/doc/ref/configuration/site-configuration.rst
@@ -94,7 +94,7 @@ The following options can be configured:
   This overrides the default list of modules installed by the
   skeleton.
 
-.. versionadded:: 0.10
+.. versionadded:: 0.16
    To inherit the list of modules from a skeleton, add a ``{skeleton,
    <name>}`` and it will install the list of modules from that skeleton
    as well.
@@ -102,6 +102,13 @@ The following options can be configured:
   The list of installed modules will be updated on each site start,
   e.g. when you add a module to the ``install_modules`` list, it will
   be installed automatically when you restart the site.
+
+``{ip_whitelist, "127.0.0.0/8,10.0.0.0/8,192.168.0.0/16,172.16.0.0/12,::1,fd00::/8"}``
+  List of TCP/IP addresses and their netmasks.
+  The admin user password *admin* will only be accepted if logging in
+  from a host matching the whitelisted IP addresses. This for protecting 
+  development systems that are exposed to the Internet.
+  This can also be configured in the :ref:`guide-configuration`.
 
 ``{smtphost, "..."}``
   Hostname you want e-mail messages to appear from. See :ref:`guide-email`.
--- a/priv/zotonic.config.in
+++ b/priv/zotonic.config.in
@@ -104,7 +104,7 @@
    {password, "%%GENERATED%%"}
 
 %%% IP whitelist, used for accessing sites with a default "admin" password
-   %% {ip_whitelist, "127.0.0.0/8,10.0.0.0/8,192.168.0.0/16,172.16.0.0/12,::1,fd00::/8"},
+   %% {ip_whitelist, "127.0.0.0/8,10.0.0.0/8,192.168.0.0/16,172.16.0.0/12,169.254.0.0/16,::1,fd00::/8,fe80::/10"},
 
 %%% Inet request backlog, increase when requests are being dropped.
    %% {inet_backlog, 500},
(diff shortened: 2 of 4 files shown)


ddeboer:
I’m in favour of this PR. It is important, however, that we document it well. Can you push some documentation in this PR?

mworrell:
I was looking where to document it and am a bit uncertain about the best chapter/section.

arjan:
Documentation indeed, plus a better name for the config option; `ip_whitelist` is too generic, since this is only a protection for the admin

mworrell:
@arjan Later I would like to re-use the same whitelist for other functionalities. These are basically trusted IP addresses.

ddeboer:
> I was looking where to document it and am a bit uncertain about the best chapter/section.

In the reference, we have http://zotonic.com/docs/0.13/ref/configuration/site-configuration.html and http://zotonic.com/docs/0.13/ref/configuration/zotonic-configuration.html.

mworrell:
@ddeboer yes, needs documentation in both, as this can be overruled per site.

arjan:
I would not use one whitelist for multiple functionalities; there might be different uses for each whitelist.
(e.g. API access, admin access, et cetera); 
The whitelist-checking code could be generic of course, but the actual config value should rather not.

mworrell:
I am not sure about the whitelists, it is good to have the ability to differentiate between whitelists, but not good to be forced to maintain multiple lists for simple use cases. Maybe we can differentiate later, if we need to add a whitelist and use the ip_whitelist as the default fallback?

ddeboer:
@arjan What about adding this as is for now? We can always introduce different whitelist later.

CyBeRoni:
Can this be extended to the zotonic status page?

CyBeRoni:
By the way I think you meant fe80::/10 instead of fd00::/8. The former is link-local addressing, the latter is for ULAs.

mworrell:
The fe80::/10 should indeed be added, if we also add 169.254.0.0/16. Which makes sense in ad-hoc networks, which are local by default.

Regarding the ULA, it is basically the same as the ipv4 private addresses, isn't it?
https://en.wikipedia.org/wiki/IPv6_address#Unique_local_addresses

> fc00::/7 — Unique local addresses (ULAs) are intended for local communication. They are routable only within a set of cooperating sites.[21] The block is split into two halves, the upper half (fd00::/8) is used for "probabilistically unique" addresses in which a 40-bit pseudorandom number is used to obtain a /48 allocation. This means that there is only a small chance that two sites that wish to merge or communicate with each other will have conflicting addresses. No allocation method for the lower half of the block (fc00::/8) is currently defined. These addresses are comparable to IPv4 private addresses (10.0.0.0/8, 172.16.0.0/12 and 192.168.0.0/16)

mworrell:
For the zotonic_status page, here we are indeed coming to the point @arjan made to have multiple whitelists...

I am thinking what the best solution is:

 1. Multiple independent whitelists
 2. Different whitelists, with the `ip_whitelist` as an ultimate fallback
 3. One list, with tagging per block what is allowed/disallowed

I am for option 2.

As we are now only using this during log on we don't need to be efficient with matching ip addresses. Only if we are going to whitelist IP addresses for requests we need efficient matching (for which there are some solutions).

CyBeRoni:
> Regarding the ULA, it is basically the same as the ipv4 private addresses, isn't it?

There's not really such a thing in IPv6. They're non-routable so I suppose it's mostly safe to include them.

ULAs are meant for non-internet-connected networks that also will never be connected to the internet.

mworrell:
I propose to merge this and add a separate issue to add an ip-whitelist for the zotonic_status site.

All ok?

arjan:
:+1: 
Separate whitelists with a single fallback whitelist is a good solution.

mworrell:
Good, then we merge this and refactor into a more generic solution.

ddeboer:
Thanks, @mworrell!

mworrell:
Also cherry-picked into the master.

Issue #1229, opened 2016-03-30
Updating a resource requires insert permission on its category


ddeboer:
Because of [`can_insert_category`](https://github.com/zotonic/zotonic/blob/master/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl#L185). Why is that check there? Is it because we often don’t differentiate between insert and update?

This is similar to #1125.

mworrell:
We have to check where it is used and might need to differentiate between insert and update (those are indeed different actions in the ACL, so should be checked different)

ddeboer:
So we need to add some `can_update_category` or `can_category`, right?

mworrell:
I will check how this is used and tune it.

ddeboer:
Awesome!

mworrell:
@ddeboer can you check branch  issue-1229-acl-update-check ?

ddeboer:
Will do, please open a PR, as that makes commenting easier.

mworrell:
See https://github.com/zotonic/zotonic/pull/1232

mworrell:
This has been merged.

PR #1232, opened 2016-03-31
Fix a problem where an user can't update their own 'person' resource.

mworrell:
This fixes issue #1229

### DIFF
--- a/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
+++ b/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
@@ -43,6 +43,7 @@
         can_insert_category/2,
         can_insert_category/3,
         can_insert_category_collab/2,
+        can_update_category/3,
         can_rsc_insert/3,
         can_move/3,
         default_content_group/2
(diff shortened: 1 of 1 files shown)


ddeboer:
Persons can now correctly update themselves. This fix breaks manage_schema for ACL rules, however. Will report back when I know more.

modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
```diff
@@ -40,6 +40,7 @@
 -export([
         can_insert_category/2,
         can_insert_category/3,
+        can_update_category/3,
```
ddeboer: This function is called nowhere. Is that intentional?

ddeboer:
This worked on 0.x but fails on this branch:

```erlang
{ok, Id} = m_rsc:insert([{category_id, page_menu}], z_acl:sudo(C)).
m_rsc:update(Id, [{title, "iets"}], z_acl:sudo(C)).
** exception throw: {{error,eacces},
                     [{m_rsc_update,update_transaction_fun_props,3,
                                    [{file,"src/models/m_rsc_update.erl"},{line,365}]},
                      {z_db,'-transaction1/2-fun-0-',3,
                            [{file,"src/db/z_db.erl"},{line,141}]},
                      {timer,tc,2,[{file,"timer.erl"},{line,179}]},
                      {z_db,with_connection,3,
                            [{file,"src/db/z_db.erl"},{line,206}]},
                      {z_db,transaction,3,[{file,"src/db/z_db.erl"},{line,95}]},
                      {m_rsc_update,update_transaction,3,
                                    [{file,"src/models/m_rsc_update.erl"},{line,303}]},
                      {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,661}]},
                      {shell,exprs,7,[{file,"shell.erl"},{line,684}]}]}
     in function  m_rsc_update:update_result/3 (src/models/m_rsc_update.erl, line 349)
```

ddeboer:
By the way, we really need functional tests (#1235) to automatically detect regressions like these.

mworrell:
@ddeboer the `observe_acl_rsc_update_check` can also restrict properties and fill in defaults. It is also run inside the update transaction. It is a last stage check before the actual update.

ddeboer:
Thanks for the fix. 

Currently, `observe_acl_rsc_update_check` duplicates a lot of checks that are also done in `observe_is_allowed`, which increases the potential for bugs (as we saw here). Do we need to do the same checks again or can we re-use the logic from `is_allowed`?

mworrell:
Maybe we can refactor, but let's do that in another issue, after the collaboration groups are merged as well.

ddeboer:
Fair enough.

PR #1235, opened 2016-04-03
tests: Add testsandboxdb site

### LINKED ISSUE #1234: Site testsandboxdb is gone?
I see some tests still [referring to testsandboxdb](https://github.com/zotonic/zotonic/search?utf8=%E2%9C%93&q=testsandboxdb). That site, however, (no longer?) exists. Should we re-add it? Testing against a site with a database connection seems very useful to me, as a lot of Zotonic functionality depends on `z_db` calls and will fail if no database is available. 

/cc @arjan

ddeboer:
* Run tests against the database

TODO:
- [x] Update Travis config
- [x] Read db config from `/etc/zotonic`/`~/zotonic`.
- [x] Add `z_notifier:await`.

Fix #1234.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ otp_release:
   - 18.0
   - 17.4
   - R16B
+services:
+  - postgresql
+before_script:
+  - psql -c 'CREATE USER zotonic; ALTER USER zotonic CREATEDB;' -U postgres
 install: echo "hi" # stub command; otherwise travis runs 'rebar get-deps'
 script: make all edocs && bin/zotonic runtests
 sudo: false
--- a/modules/mod_acl_user_groups/tests/mod_acl_user_groups_tests.erl
+++ b/modules/mod_acl_user_groups/tests/mod_acl_user_groups_tests.erl
@@ -1,0 +1,29 @@
+%% @doc Tests for mod_acl_user_groups
+-module(mod_acl_user_groups_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+-include("zotonic.hrl").
+
+person_can_edit_own_resource_test() ->
+    Context = z_context:new(testsandboxdb),
+    ok = z_module_manager:activate_await(mod_content_groups, Context),
+    ok = z_module_manager:activate_await(mod_acl_user_groups, Context),
+
+    %% Person must be able to edit person category
+    m_acl_rule:replace_managed(
+        [
+            {rsc, [
+                {acl_user_group_id, acl_user_group_anonymous},
+                {actions, [view, update]},
+                {is_owner, true},
+                {category_id, person}
+            ]}
+        ],
+        testsandboxdb,
+        z_acl:sudo(Context)
+    ),
+
+    {ok, Id} = m_rsc:insert([{category, person}], z_acl:sudo(Context)),
+
+    UserContext = z_acl:logon(Id, Context),
+    {ok, Id} = m_rsc:update(Id, [{title, "Test"}], UserContext).
--- a/src/scripts/zotonic-runtests
+++ b/src/scripts/zotonic-runtests
@@ -55,7 +55,7 @@ case "$1" in
         echo
 
         # Run the tests
-        $ERL +P 10000000 +K true -pa $PA $NAME_ARG ${NODENAME}_testsandbox@$NODEHOST -boot start_sasl -sasl errlog_type error -s zotonic -eval "timer:sleep(8000),init:stop(case eunit:test([$ALL],[]) of error -> 1; ok -> 0 end)"
+        $ERL +P 10000000 +K true -pa $PA $NAME_ARG ${NODENAME}_testsandbox@$NODEHOST $(find_config_arg zotonic.config) -boot start_sasl -sasl errlog_type error -s zotonic -eval "timer:sleep(8000),init:stop(case eunit:test([$ALL],[]) of error -> 1; ok -> 0 end)"
         EXIT=$?
         ;;
 esac
--- a/doc/ref/directory-structure.rst
+++ b/doc/ref/directory-structure.rst
@@ -79,7 +79,7 @@ Erlang application::
 
   The directory containing sites which are internal to Zotonic. These
   are the ``zotonic_status`` site (see
-  :ref:`ref-status-site`), and the ``testsandbox`` site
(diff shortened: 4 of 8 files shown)


src/scripts/zotonic-runtests
```diff
@@ -55,7 +55,7 @@ case "$1" in
         echo
 
         # Run the tests
-        $ERL +P 10000000 +K true -pa $PA $NAME_ARG ${NODENAME}_testsandbox@$NODEHOST -boot start_sasl -sasl errlog_type error -s zotonic -eval "timer:sleep(8000),init:stop(case eunit:test([$ALL],[]) of error -> 1; ok -> 0 end)"
+        $ERL +P 10000000 +K true -pa $PA $NAME_ARG ${NODENAME}_testsandbox@$NODEHOST -boot start_sasl -sasl errlog_type error -s zotonic -config /etc/zotonic/zotonic.config -eval "timer:sleep(8000),init:stop(case eunit:test([$ALL],[]) of error -> 1; ok -> 0 end)"
```
ddeboer: TODO: remove hardcoded reference to config file. How do we find the user’s database credentials?

mworrell: Maybe from the Travis-CI config file (or just Travis-CI defaults)?

modules/mod_acl_user_groups/tests/mod_acl_user_groups_tests.erl
```diff
@@ -0,0 +1,30 @@
...
+                {acl_user_group_id, acl_user_group_members},
+                {actions, [view, update]},
+                {is_owner, true},
+                {category_id, person}
+            ]}
+        ],
+        testsandboxdb,
+        z_acl:sudo(Context)
+    ),
+    m_acl_rule:publish(rsc, z_acl:sudo(Context)),
```
ddeboer: Can be removed now #1225 has been merged.

modules/mod_acl_user_groups/tests/mod_acl_user_groups_tests.erl
```diff
@@ -0,0 +1,30 @@
...
+            {rsc, [
+                {acl_user_group_id, acl_user_group_members},
+                {actions, [view, update]},
+                {is_owner, true},
+                {category_id, person}
+            ]}
+        ],
+        testsandboxdb,
+        z_acl:sudo(Context)
+    ),
```
ddeboer: This fails on Travis with:

```
11:46:36.510 [error] [testsandboxdb] z_db error {error,error,<<"42P01">>,<<"relation \"acl_rule_rsc\" does not exist">>,[{position,<<"13">>}]} in query "DELETE FROM acl_rule_rsc WHERE managed_by = $1" with [testsandboxdb]
mod_acl_user_groups_tests: person_can_edit_own_resource_test (module 'mod_acl_user_groups_tests')...*failed*
in function z_db:'-q/4-fun-1-'/5 (src/db/z_db.erl, line 289)
in call from timer:tc/2 (timer.erl, line 181)
in call from z_db:with_connection/3 (src/db/z_db.erl, line 206)
in call from m_acl_rule:delete_managed/2 (modules/mod_acl_user_groups/models/m_acl_rule.erl, line 516)
in call from m_acl_rule:replace_managed/3 (modules/mod_acl_user_groups/models/m_acl_rule.erl, line 264)
in call from mod_acl_user_groups_tests:person_can_edit_own_resource_test/0 (modules/mod_acl_user_groups/tests/mod_acl_user_groups_tests.erl, line 13)
**throw:{error,{error,error,<<"42P01">>,
              <<"relation \"acl_rule_rsc\" does not exist">>,
              [{position,<<"13">>}]}}
  output:<<"">>
```

@mworrell This looks like a timing issue, where the ACL module hasn’t been (completely) started yet. Is this correct? If so, is there a way to wait for the module to be completely initialised?

mworrell: That is indeed a problem where the module is requested to start, but as it is starting async, it isn't finished starting when the test is running.

We might want to add something like `z_module_manager:await(ModuleName, Context)`.

.travis.yml
```diff
@@ -3,6 +3,10 @@ otp_release:
   - 18.0
   - 17.4
   - R16B
+services:
+  - postgresql
+before_script:
+  - psql -c 'create user zotonic; ALTER USER zotonic CREATEDB;' -U postgres
```
ddeboer: Let’s change this to use a separate schema, not database.

ddeboer:
@mworrell Will add the `z_notifier:await` so we can wait for the modules to have started completely.

mworrell:
Added the `z_notifier:await` in c483cde6c7749a8b172f4f744d2a2be760d5227d
You can also use: `z_module_manager:activate_await/2`

ddeboer:
Wicked!

ddeboer:
Now using `activate_wait` but the build is still failing. @mworrell Any ideas?

mworrell:
@ddeboer will check

arjan:
In #1254 I have just added database to the general testsandbox site. Why should we have 2 testsandbox sites?

mworrell:
@arjan Good question.  The only reason I can come up with is if we want to be able to run Zotonic without a database, or do quick functional checks without db checks.

The former is not something we actively support. The latter might be useful if working on non-db parts of the system.

ddeboer:
I see two advantages to having a separate db-enabled test site:

- It will make it explicit that some test requires a database connection; isolated function tests can just rely on the site without a database
- In the long run, we may want to reset the database before each separate database test so we can isolate them. Obviously, this slows down things, so the more tests can run without a database, the better.

mworrell:
@ddeboer can you check with fe08682bc52c06a19c3f0ceeab0024e24169a3f5 merged in?

mworrell:
The above commit should fix the activate_await/2 call.

ddeboer:
> @ddeboer can you check with fe08682 merged in?

Rebased on 0.x.

ddeboer:
Awesome, now green. Thanks, @mworrell! 👍 

Squashed commits, so ready to merge.

ddeboer:
:sparkles: 🎩

mworrell:
Also merged into master - small merge conflict with .travis.yml, doesn't seem to be a problem.

PR #1252, opened 2016-04-26
mod_acl_user_groups: Fix permission issue

arjan:
When content_group_id is not passed on insert, the insert fails when the user has no permissions to insert anything in any category *except* on a category where he is a collaboration group of.

With this fix, the content group id will be set to the first collaboration group where the user has insert rights on.


arjan:
As this is my first commit in a while I hop I understood the module's internals correctly ;-)

mworrell:
👍 Maybe also merge into the upcoming release?

mworrell:
Merged, will be part of 0.16

Issue #1139, opened 2016-01-12
Drop Erlang < 18 support


ddeboer:

ddeboer:
We now run Travis on Erlang >= 18 and our docs state Erlang 18 as requirement.

PR #1254, opened 2016-04-27
core: Switch to Rebar3

### LINKED ISSUE #1241: rebar3 support
[Rebar3](http://www.rebar3.org/) is becoming the de-facto standard for Erlang compilation, now that it's even going to be officially supported by the OTP team.

Things like dependency management are easier (and faster) with rebar3.

Should we move our compilation process to it as well?

ddeboer:
Fix #1241.


ddeboer:
Let’s find out whether this builds.

arjan:
I think this is now good to merge. @mworrell @ddeboer can you give it a spin?

.travis.yml
```diff
@@ -1,9 +1,9 @@
 language: erlang
 otp_release:
+  - 18.2.1
   - 18.0
-  - 17.4
-  - R16B
```
ddeboer: Dropping Erlang < 0.18 is fine on master: see #1139.

However, is the plan to merge this into 0.x as well? I would prefer so. Does rebar3 require Erlang 18?

priv/sites/testsandbox/config
```diff
@@ -10,7 +10,11 @@
  %% Hostname for the test sandbox. Do NOT change this! It will break the tests.
  {hostname, "localhost:8040"},
 
- {dbdatabase, none},
+ {dbhost, "127.0.0.1"},
+ {dbport, 5432},
+ {dbuser, "postgres"},
+ {dbpassword, ""},
+ {dbdatabase, "testsandbox"},
```
ddeboer: As there are no added (database) tests in this PR, I don’t see the benefit of changing the sandbox setup here. This PR is about rebar3, let’s leave an improved test setup to #1235.

rebar.config
```diff
@@ -12,36 +12,35 @@
...
+  {poolboy, "1.5.1"},
+  {epgsql, "3.1.1"},
+  {erlware_commons, "0.20.0"},
+  {erlang_localtime, "1.0.0"},
+  {gproc, "0.5.0"},
+  {parse_trans, "2.9.0"},
+  {proper, "1.1.1-beta"},
+  {recon, "2.3.1"},
+  {meck, "0.8.4"},
+  {edown, "0.7.0"},
```
ddeboer: Yes, finally we can lock depend on released versions instead of on master!

ddeboer:
Very cool, @arjan, and it builds on Travis! 🔥 

This now conflicts with master, so please rebase. If we want to merge this into 0.x as well, please squash commits as well.

priv/sites/testsandbox/config
arjan: Agreed

arjan:
I would rather have this on master only, I tried this with Erlang R16 but ran into a variety of problems.

ddeboer:
All right, one more reason to get 1.0 out quickly and ditch 0.x. :wink: @mworrell Do you agree?

And @arjan could you still resolve the conflict and squash commits?

arjan:
Working on it ;)

arjan:
Fixed in 7dd4e51

ddeboer:
Why did you close rather than merge this PR?

arjan:
Good question.. I always merge from the CLI, github didnt pick this up apparently

ddeboer:
The [usual flow](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit) when wrapping up a PR is to rebase (and solve any merge conflicts) squash commits (to get a linear history that can be easily cherry-picked).

mworrell:
@ddeboer indeed another reason to get the master up & running 👍 Am on it.

Issue #1150, opened 2016-01-19
Refactor z_ids module


mmzeeman:
This module generates list based random identifiers, numbers and ids used throughout zotonic modules.

The current implementation is a gen_server, which can become a bottleneck when a lot of ids are requested. With the upcoming binarification of zotonic we should generate id's and identifiers as binaries, and use ```crypto:randbytes/1``` so the gen_server can be removed. 

*Note: This is a breaking change*

ddeboer:
@willemdj Is this a good one for you to work one? Nice introduction to Zotonic. :wink:

willemdj:
I am still new to this, so I hope it is okay if I ask a couple of questions.

__Getting rid of the gen_server__
Under normal operational circumstances the functions in the module are all stateless, so they don't need a gen_server. 

Only when the server is started with the `fixed_seed` option or after execution of `z_ids:fix_seed()` the state becomes important: it changes the behavior of some of the functions, and some of those are no longer stateless.

I believe that it would be possible to change this approach in such a way that the API functions are executed without the gen_server under normal circumstances. This could be done by loading a different version of the module under test circumstances. Perhaps using a compiler directive, or using something like meck. Alternatively it could be based on m_config or aplication:get_env()

Questions: 
- is it important to keep the `fix_seed` functionality,
- if so, must the interface to use activate that functionality remain exactly as it is today, or would it be acceptable to achieve the same result in another way, for example by re-compiling and reloading the module.

__Efficieny and move to binary__
The exact requirements for each of the functions in the module are not clear to me. In order to rewrite the functions to return a binary value, it would be good to understand exactly which aspects of the current implementation are really required, and which are the results of implementation choices.

- Is it the expectation that `id` and `identifier` are unique? (in what context 
  or with what periodicity?) 
- Must they be random? Or is the random function used as a way to achieve uniqueness?
- Must the identifiers be exactly of the specified length, or is the length used as a way to guarantee a certain level of uniqueness?

(I would assume that the main requirement for session ids and ids in HTML etc is that they are unique, rather than random?)

I can imagine that it may be efficient  to use `erlang:unique_integer().` Whether that is acceptable or not depends on the exact requirements.

For example `integer_to_binary(erlang:unique_integer([positive]))` gives values that are unique (as long as the VM is not restarted) in a very efficient way, but they are not random, they contain only numbers and they are not of a fixed length (but they could be made to be, with certain limitations).

ddeboer:
Good questions!

[...]

PR #1270, opened 2016-05-10
core: Refactor z_ids module.

willemdj:
Removes a potential bottleneck by changing z_ids from a gen_server into a stateless module.

Improves the randomness by using crypto:strong_random_bytes/1.

Changes the return type of several functions from string to binary, in accordance with the general move to binary.

Makes the necessary changes in the modules that call functions in z_ids to deal with the new return type.

Solves #1150 (Refactor z_ids module).

### DIFF
--- a/modules/mod_mailinglist/models/m_mailinglist.erl
+++ b/modules/mod_mailinglist/models/m_mailinglist.erl
@@ -234,7 +234,7 @@ insert_recipient(ListId, Email, Props, WelcomeMessageType, Context) ->
 					  from mailinglist_recipient 
 					  where mailinglist_id = $1
 					    and email = $2", [ListId, Email1], Context),
-	ConfirmKey = z_ids:id(20),
+	ConfirmKey = binary_to_list(z_ids:id(20)),
 	WelcomeMessageType1 = case Rec of
 		{RecipientId, true, _OldConfirmKey} ->
 			%% Present and enabled
--- a/src/support/z_ids.erl
+++ b/src/support/z_ids.erl
@@ -18,59 +18,80 @@
 
 -module(z_ids).
 -author("Marc Worrell <marc@worrell.nl>").
--behaviour(gen_server).
+-include("zotonic.hrl").
 
-%% Length of session keys, used for the cookies, must be unique
 -define(ID_LENGTH,20).
 -define(OPTID_LENGTH,6).
 
-%% gen_server exports
--export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2, code_change/3]).
-
-%% id server exports
 -export([
-    start_tests/0,
-    start_link/0,
     unique/0, 
     id/0,
     id/1,
     identifier/0,
     identifier/1,
+    random_id/2,
     optid/1,
     sign_key/1,
     sign_key_simple/1,
     number/0,
-    number/1,
-    fix_seed/0
+    number/1
 ]).
 
--record(state, {is_fixed = false, unique_counter = 0}).
--include("zotonic.hrl").
+-type charset() :: 'az' | 'az09' | 'azAZ09'.
 
-start_tests() -> 
-    gen_server:start({local, ?MODULE}, ?MODULE, [[{fixed_seed,true}]], []).
-start_link() -> 
-    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+%%%--------------------------------------------------------------------------
+%%% API
+%%%--------------------------------------------------------------------------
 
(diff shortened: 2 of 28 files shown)


modules/mod_mailinglist/models/m_mailinglist.erl
```diff
@@ -202,7 +202,7 @@ recipient_delete1(RecipientProps, Quiet, Context) ->
 recipient_confirm(ConfirmKey, Context) ->
 	case z_db:q_row("select id, is_enabled, mailinglist_id from mailinglist_recipient where confirm_key = $1", [ConfirmKey], Context) of
 		{RecipientId, _IsEnabled, ListId} ->
-			NewConfirmKey = z_ids:id(20),
+			NewConfirmKey = binary_to_list(z_ids:id(20)),
 			z_db:q("update mailinglist_recipient set confirm_key = $2, is_enabled = true where confirm_key = $1", [ConfirmKey, NewConfirmKey], Context),
```
arjan: Can't NewConfirmKey be passed into the db as a binary?

willemdj: I was not sure about the effects. 

I don't know what conversions happen when you store and retrieve a value in the database. If a string is automatically converted to a binary when stored, then obviously there is no reason to convert it to string first, and the call to binary_to_list() can be removed.

In general I tried to "err on the safe side", and I added conversions to the original string type where it was not clear to me that it would be safe to omit them.

mworrell: It is converted to a binary when processed by the db layer.

In fact we are in the process of moving all string data to binary (utf8) values.

mworrell:
I think  we check the `binary_to_list/1` calls as part of the *binaryfication*
That will also remove many other from/to binary calls.

There are some minor problems with the edocs, please check the details of the Travis-CI build.

modules/mod_mailinglist/models/m_mailinglist.erl
willemdj: I'll remove the conversion then.

I noticed that there is also a problem with the documentation, I'll fix that as well.

src/support/z_ids.erl
```diff
@@ -18,72 +18,89 @@
...
+%%% API
+%%%--------------------------------------------------------------------------
 
-%% @doc Return an unique id to be used in javascript or html.  No randomness, just unique in the cluster.
+-spec unique() -> binary().
+%% @doc Return an unique id to be used in javascript or html.  No randomness,
+%% just unique for the current runtime system instance.
+%%
+%% Returns a binary of unspecified length starting with the character $t followed
+%% by a number of digits, example: <<"t123456">>.
```
ddeboer: edoc trips over `<<"…">>` (see Travis)

src/support/z_ids.erl
```diff
@@ -18,72 +18,89 @@
...
-identifier() -> 
-    gen_server:call(?MODULE, {identifier, ?OPTID_LENGTH}).
-
-identifier(Len) -> 
-    gen_server:call(?MODULE, {identifier, Len}).
-
+    random_id('azAZ09', Len).
+
+-spec identifier() -> binary().
+-spec identifier(Length::integer()) -> binary().
```
ddeboer: Seems multiple `-spec`s are not allowed. On Travis:

```
/home/travis/build/zotonic/zotonic/src/erlydtl/erlydtl_compiler.erl, function identifier/0: at line 1692: multiple @spec tag.
edoc: skipping source file '/home/travis/build/zotonic/zotonic/src/erlydtl/erlydtl_compiler.erl': {'EXIT',
                                                                                             error}.
```

ddeboer:
Great work, @willemdj! 👍 

Some small edocs problems on Travis. I added some comments on your code to point you in the right direction.

willemdj:
I solved the edocs problems - I didn't know it was not accepted by edoc to use this way of annotation, I got the example from some OTP sources...

I also removed two conversions to string where I was sufficiently confident that they are not required.

I hope the git handling is OK, I amended the earlier commit.

ddeboer:
Amending is perfect!

ddeboer:
Build now passes. @mworrell This looks good to me. The only thing missing is some tests, but we didn’t have them before anyway. So let’s merge this?

mworrell:
I am ok with merging!

ddeboer:
Awesome, @willemdj! 🌟 

Time for another challenge? :wink:

mworrell:
🍰 thanks! @willemdj

willemdj:
It still feels like a good way to get to know Zotonic.  So sure, if there
is something useful that needs to be done, let me know and I'll see what I
can do.

PR #1281, opened 2016-05-19
Switch to universal analytics using async method

### LINKED ISSUE #1249: Switch to Universal Analytics
Google has deprecated the old Analytics, so let’s [switch to Universal](https://developers.google.com/analytics/devguides/collection/analyticsjs/#the_javascript_tracking_snippet). I’m in favour of the [alternative script] that allows modern browsers to preload (but gives a slight penalty on IE 9 and below).

tahteche:
To use the User ID feature, the user will need to set it up from the Google Analytics backend.
https://support.google.com/analytics/answer/3123666?hl=en

Fixes #1249

### DIFF
--- a/modules/mod_seo/templates/_html_head.tpl
+++ b/modules/mod_seo/templates/_html_head.tpl
@@ -43,16 +43,15 @@
 {% endif %}{% endwith %}
 {% if m.acl.user /= 1 and not notrack %}
 	{% with m.config.seo_google.analytics.value as ga %}{% if ga %}
-	<script type="text/javascript">
-		var _gaq = _gaq || [];
-		_gaq.push(['_setAccount', "{{ ga }}"]);
-		_gaq.push(['_trackPageview']);
-		{% if m.acl.user %}_gaq.push(['_setCustomVar',1,'User','True',1]);{% endif %}
-		(function() {
-			var ga = document.createElement('script'); ga.type = 'text/javascript'; ga.async = true;
-			ga.src = ('https:' == document.location.protocol ? 'https://ssl' : 'http://www') + '.google-analytics.com/ga.js';
-			var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
-		})();
+	<script>
+		window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
+		{% if m.acl.user %}
+			ga('create', "{{ ga|escapejs }}", 'auto', { userId: "{{ m.acl.user|escapejs }}" });
+		{% else %}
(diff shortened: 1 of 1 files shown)


modules/mod_seo/templates/_html_head.tpl
```diff
@@ -46,7 +46,7 @@
 	<script>
 		window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
 		{% if m.acl.user %}
-			ga('create', "{{ ga }}", 'auto', { userId: m.acl.user});
+			ga('create', "{{ ga }}", 'auto', { userId: {{ m.acl.user }} });
```
arjan: `m.acl.user` needs quoting otherwise it'll cause a javascript error. It's wise to also apply the `escapejs` filter. The entire line would become:

```
ga('create', "{{ ga|escapejs }}", 'auto', { userId: "{{ m.acl.user|escapejs }}" });
```

ddeboer:
Cool! Looks good to me. Could you please squash the commits?

tahteche:
I have quoted `m.acl.user` but when I use the escapejs filter only the admin section works as the front end shows a white page.

arjan:
Anything in the JS console?

mworrell:
@tahteche Is this ok to merge & test, or are you still experiencing some problems?

tahteche:
@mworrell Sorry for the absence, fell ill. The PR needs some rectification over using the `escapejs` filter before merging.

@arjan   All I get in the JS concole is a 500 (Internal Server Error) which flashes for less than a sec. Since I run Zotonic with `zotonic debug`, I get the log entry linked to below each time I refresh the page in the CLI:
http://pastebin.com/yFXfeaFB

mworrell:
Aha, apparently `escapejs` filter can't escape an integer.

```
escapejs(undefined, _Context) ->
    <<>>;
escapejs(Input, _Context) when is_binary(Input) ->
    z_utils:js_escape(Input);
escapejs(Input, _Context) when is_list(Input) ->
    z_utils:js_escape(Input).
```

@arjan seems to me that we don't need these guards and can just directy call `z_utils:js_escape/2` for any any value. What's your opinion?

arjan:
Well, since we know that `m.acl.user` is always an integer, we don't need the filter in the first place for this particular case. But removing the guards makes sense, provided that `z_utils:js_escape` does the right thing.

mworrell:
I will remove the guards.

tahteche:
@mworrell It now works with the `escapejs` filter. I think its now fine to be merged.

ddeboer:
Thanks @tahteche, awesome! 🌟

Issue #1131, opened 2016-01-08
Deprecate adminonly and simple_acl in favour of acl_user_groups


ddeboer:
There are now too many ACL options (especially for newcomers to the framework) to choose from. 

What about deprecating adminonly and simple_acl or at least strongly recommending to use acl_user_groups? Or does anyone still see good use cases for the older two ACL modules?

mworrell:
I am for moving te `mod_acl_simple_roles` out of the core.
We can do the same with the `mod_acl_adminonly` if (and only if) we have good defaults for that module and clear documentation.
Till that moment I am for keeping the `mod_acl_adminonly` in the core.

ddeboer:
> I am for moving te mod_acl_simple_roles out of the core.

Good idea. Let’s start there.

> for that module and clear documentation.

You mean defaults for acl_user_groups, right?

mworrell:
Yes, good defaults for the user groups,  content groups and some rules.

mmzeeman:
Do you mean by good defaults also a way to move your resources from simple_roles to user_groups?

mworrell:
@mmzeeman That could be a support function you can run manually?

mmzeeman:
Yes of course. We currently have groups via predicates. That would allow us to write an update routine.

ddeboer:
We can start adding default ACL rules now #1173 has been merged.

ddeboer:
What about the following set?

Group |  Content group | Category | Permissions
---------|------------------|------------|-----
anonymous | all | all | view
managers | all | all | view,insert,update,remove,link

Group |  Modules | Permissions
---------|------------------|------------
managers | mod_admin | use
managers | mod_admin_category | use
managers | mod_admin_identity | use
managers | mod_content_groups  | use
managers | mod_menu  | use

Do we agree about adding this set as fixtures? This does mean they will be read-only on the ACL screen and cannot be altered later. We could optionally place them in a separate module, but that module should then be enabled by default to support the idea of an out-of-the-box set of useful rules.

mworrell:
:+1: 

Also we need to fix #1211

mmzeeman:
What is the procedure for moving from acl simple roles to the new module?

ddeboer:
1. Release mod_acl_user_groups with sane defaults and add a deprecation warning when mod_acl_simple_roles is enabled (e.g. 0.15).
2. Remove mod_acl_simple_roles out of the core (e.g. 0.18: 3 months later?).

[...]

PR #1284, opened 2016-05-21
mod_acl_user_groups: Insert default ACL rules

ddeboer:
See #1131.

### DIFF
--- a/modules/mod_acl_user_groups/mod_acl_user_groups.erl
+++ b/modules/mod_acl_user_groups/mod_acl_user_groups.erl
@@ -69,8 +69,8 @@
 
 %% gen_server state record
 -record(state, {
-            site, 
-            is_rebuild_publish=true, 
+            site,
+            is_rebuild_publish=true,
             is_rebuild_edit=true,
             rebuilder_pid,
             rebuilder_mref,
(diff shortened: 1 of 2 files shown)


modules/mod_acl_user_groups/mod_acl_user_groups.erl
```diff
@@ -601,19 +601,22 @@ manage_datamodel(Context) ->
...
-                    {hascollabmanager, 
+                    {hascollabmanager,
                         [{title, {trans, [{en, <<"Manager">>},{nl, <<"Beheerder">>}]}}],
                         [{acl_collaboration_group, person}]
                     }
-                ]
+                ],
+            data = [
+                {acl_rules, acl_default_rules:get_default_rules()}
+            ]
```
ddeboer: This is the major change. Sorry about the large diff, but my IDE (correctly) strips all trailing spaces.

arjan: Aren't rules that are inserted programmatically causing that the rules cannot be edited in the UI anymore?

ddeboer: Correct. That’s why I’m keeping the set of default rules to the bare minimum, hopefully covering the 80%. Do you nonetheless see some problems with it?

Of course, if you really do not want these default rules, you can always `observe_manage_data(#manage_data{module = Module, props = {acl_rules, Rules}}, Context)` in your site and get rid of them or change them.

mworrell:
I am ok with this.  @fredpook and I are planning to add a `is_disabled` flag to the ACL rules, so that ACL rules can be temporarily disabled, which is very useful for schema-installed rules.

mworrell:
Also merged 25a3da9 into 0.x

ddeboer:
Thanks, @mworrell! 

@fredpook @mworrell Please create an issue for `is_disabled` so we can discuss whether it’s needed.

PR #1287, opened 2016-05-22
doc: Document mod_acl_user_groups

ddeboer:
Fix #1240.

PR #1289, opened 2016-05-23
mod_acl_adminonly: Remove the module from Zotonic core

ddeboer:
Requires #1287, #1284.

Ref #1131.


arjan:
I guess this is only for the 1.x release?
Also it might be good to add this to the upgrade notes, how to migrate from `mod_acl_adminonly` to `mod_acl_user_groups`.

ddeboer:
Indeed, although we can introduce this change earlier, it’s probably easiest to do this for 1.0. Will add upgrade notes.

ddeboer:
Added upgrade note.

doc/developer-guide/upgrading.rst
```diff
@@ -7,6 +7,14 @@ These notes list the most important changes between Zotonic
...
 major Zotonic version.
 
+Upgrading from 0.x to 1.0
+-------------------------
+
+ACL
+"""
+
+* mod_acl_adminonly was replaced by :ref:`mod_acl_user_groups`. To create users
+  that have access to the admin, add them to the ‘Managers’ user group.
```
ddeboer: This requires the default ACL rules for managers from #1284.

mworrell:
👍 this module has its run and should indeed be shelved.

PR #1297, opened 2016-05-30
docker: Install Erlang 18

### LINKED ISSUE #1294: Install Erlang R18 in Docker container
The current Dockerfile doesn’t work because it relies on Jessie’s stable Erlang package, which is on R17, while our master now requires R18.

- [x] add Docker package repo and install R18

ddeboer:
Zotonic master now requires Erlang 18.

Fix #1294.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,22 @@
 FROM debian
 MAINTAINER Andreas Stenius git@astekk.se
 
-RUN apt-get update                                                                                 && \
-    apt-get install -y erlang build-essential postgresql imagemagick exif wget git                 && \
+ENV DEBIAN_FRONTEND noninteractive
+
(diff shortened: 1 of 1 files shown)


ddeboer:
@kaos Would you like to have a look before we merge this?

kaos:
No, not really. I don't see any apparent issues with the Dockerfile, so go ahead :)

kaos:
Btw, any of you going to DockerCon in Seattle (it's now in June, 20-21)?
If so, we could meet up there.. ;)

mworrell:
(hijacking issue) @kaos No, maybe we should submit a talk to the Erlang conf in Stockholm though.

ddeboer:
@kaos No, I won’t be there.

@mworrell Unfortunately, it seems the CFP for Erlang Stockholm is over.

PR #1298, opened 2016-06-02
Inconsistent text placement in thurstone block

hoosiersoft:
The explanation text in the thurstone survey blocks is placed after the
input elements. This is inconsistent with the other block types that
place the explanation text immediately after the label (prompt).

This change results in more consistent admin and view appearance of
surveys.


mworrell:
Thanks!  Merging this into master for the upcoming 1.0.

PR #1299, opened 2016-06-03
Remove condition that makes it impossible for anonymous users to edit resources

Dirklectisch:
Even if you explicitly configure an ACL rule that enables editing.

### DIFF
--- a/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
+++ b/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
@@ -623,6 +623,12 @@ is_owner(insert_rsc, _Context) ->
 is_owner(Id, Context) ->
     is_owner(Id, m_rsc:p_no_acl(Id, creator_id, Context), Context).
 
+is_owner(Id, undefined, #context{user_id=undefined, session_id=SessionId} = Context) ->
+    case m_rsc:p_no_acl(Id, session_owner, Context) of
+        undefined -> false;
+        SessionOwner -> SessionId == SessionOwner
+    end;
+is_owner(_Id, _CreatorId, #context{user_id=undefined}) -> false;
 is_owner(Id, _CreatorId, #context{user_id=Id}) -> true;
 is_owner(_Id, CreatorId, #context{user_id=CreatorId}) -> true;
 is_owner(_Id, _CreatorId, _Context) -> false.
--- a/modules/mod_admin/filters/filter_temporary_rsc.erl
+++ b/modules/mod_admin/filters/filter_temporary_rsc.erl
@@ -50,15 +50,16 @@ task_delete_inactive(RscId, SessionId, Context) ->
     case is_unmodified_rsc(RscId, Context) of
         true ->
             case is_session_alive(SessionId, Context) of
-                true ->
+                false ->
                     lager:debug("[~p] Deleting unmodified temporary resource ~p", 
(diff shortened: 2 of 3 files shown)


arjan:
:+1:

Dirklectisch:
I made some more adjustments that make anonymous users owners of their temporary resource for the duration of their session.

modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
```diff
@@ -623,6 +623,12 @@ is_owner(insert_rsc, _Context) ->
 is_owner(Id, Context) ->
     is_owner(Id, m_rsc:p_no_acl(Id, creator_id, Context), Context).
 
+is_owner(Id, undefined, #context{user_id=undefined, session_id=SessionId} = Context) ->
+    case m_rsc:p_no_acl(Id, session_owner, Context) of
+        undefined -> false;
```
mworrell: By placing the session-id as a property in the resource, you are in fact publishing it to the whole world.
Another solution could be to have the resource-id in the session process and then check against the session state.

mworrell:
You can also create an anonymous user account to bind all edits during a particular session together.

An user-facing option could be to remember this anonymous account on the browser, for returning visits. Later the user can then change that anonymous account into a real account, without loosing history and access to the previous edits.

mworrell:
I think you need to be very careful about the use cases here. If you only allow for anonymously editing content during a particular session, then a small hiccup could disable access to the newly created page. That would force the user to re-create a new page.

modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
Dirklectisch: Ouch, I guess you are right about publishing the session-id to the whole world this way. I'll look into your suggestion of storing the RscId's in the session process.

Dirklectisch:
We do not count on the session lasting longer than the initial content insertion. It's just that to make that work two edits happen, first the insert of the temporary resource then the update with the content. For that second part we needed someway to keep the resource editable. Hence the session solution. The goal was not keeping the resource editable beyond that. (Otherwise I wouldn't have it clean up the session-id in the inactive-resource task).

We discussed having anonymous user accounts but that introduces all sorts of other questions into the equation. For example there needs to be an upgrade process and a solid defense against people trying to claim each others accounts. Although your points are completely valid.

Even if we do opt to take the anonymous user accounts route, I think it should at least be possible to post something to Zotonic in an anonymous publish once fashion if a developer chooses to.

mworrell:
I mostly handle this kind of anonymous submissions by temporarily sudo-ing as another user who has the correct permissions.

Dirklectisch:
That is certainly something we have been doing as well. The downside is that you will loose some of the benefits of the ACL system. Want to know if this is_editable? Better make sure your templates handle the anonymous case explicitly. It also requires the developer to drop down to the Erlang level, where else would the sudo-ing happen?

mworrell:
I agree.

Though I think it would be easier to have a real user around, otherwise you have to check the session state for every single resource without a creator.

mworrell:
Copied the changes to z_acl and filter_temporary_rsc.

In discussion with @Dirklectisch we decided to put the anonymous user checks in a separate module.

ddeboer:
Good job on fixing the old behaviour, @Dirklectisch and @mworrell!

Issue #1113, opened 2015-12-17
Add page blocks to full text search


ddeboer:
The content of page blocks is currently not searchable. Let’s make it so, perhaps with a slightly lower priority than the body text?

mworrell:
Maybe this can be checked in the context of #1076 ?

mworrell:
Putting this on the same milestone as #1076

ddeboer:
Agreed.

ddeboer:
Discussed today:

* only index page block title and body, with same priority (A/B) as parent resource
* embedded page blocks (that refer to other resources) that are inline should be indexed as well.

mworrell:
Proposal to make full text  indexing really flexible: add templates for rendering the texts that need to be indexed. The template can be made of four blocks/parts. Each part signifying a A...D priority. After rendering we can fetch the parts and build the full text index. If we use a catinclude for the template then it is easy for front end programmers to fine tune how different categories can be indexed.

In the template itself we can loop over the different languages provided by the resource. The pivoter will use the default language as pre-selected language for rendering (and sudo permissions to access the resource).

ddeboer:
I like it!

mworrell:
Note: also add a block that can be filled with a proposed re-indexing date. In this way an event can be automatically demoted after its end date.

fredpook:
+1

ddeboer:
We decided that @mworrell will create the template (`fulltext.tpl`). We’ll then discuss together what that template should exactly look like.

Issue #1076, opened 2015-11-04
More control over search ranking


ddeboer:
As discussed yesterday with @mworrell, @loetie and @fredpook (see our notes below), the way search results are currently ranked can be confusing for users:
- related pages are ranked higher than a page that is primarily about the search term when that search term appears more often on the related pages
- for results in categories such as news or blog, old items can be ranked above more recent and relevant ones
- results in categories that form the core business (such as subsidies) are ranked too low.

These issues (and related ones) can be solved by allowing more fine-grained control of the way Zotonic ranks search results (through PostgreSQL’s weights A/B/C/D). This can be done as follows.
1. A single resource is ranked higher when the `is_featured` property is set.
2. All resources in a category are ranked higher when the `is_featured` property is set on the category.
3. A new `search_rank` (?) fold notification allows even more detailed control. By observing this notification, sites can prefer more recent items over outdated ones. To work correctly for date-related ranking, this notification must be scheduled to run each day/week/month.

![img_2117 1](https://cloud.githubusercontent.com/assets/89267/10944450/8b8975ba-8319-11e5-85fd-af790740638e.jpg)

mworrell:
Note:

Maybe the re-indexing of old events (ie. in the past) could be run in batches? Like fetching date ranges for stuff that is between thresholds. This query could even be a configurable query.

If we run this query then we could get the set of pages to newly add to one of the buckets: important, normal, low, and not-important-at-all (need to come up with names, maybe just A, B, C and D).

By using buckets (and reindexing accordingly) we might be able to prevent big re-indexing actions.

The text of a resource is re-sorted into the A/B/C/D categories depending on the priority. This priority should also be saved with the resource, so that we can use it in more dynamic ways for filtering/sorting etc.

mworrell:
See the full text indexing proposal on #1113

fredpook:
See also: #1113

"Proposal to make full text indexing really flexible: add templates for rendering the texts that need to be indexed. The template can be made of four blocks/parts. Each part signifying a A...D priority. After rendering we can fetch the parts and build the full text index. If we use a catinclude for the template then it is easy for front end programmers to fine tune how different categories can be indexed.

In the template itself we can loop over the different languages provided by the resource. The pivoter will use the default language as pre-selected language for rendering (and sudo permissions to access the resource)."

ddeboer:
This can only and will be done on master.

ddeboer:
[...]

PR #1300, opened 2016-06-03
New template compiler

mworrell:
**Work in progress**

Integrating the new template compiler at https://github.com/zotonic/template_compiler

To do:

- [x] Add include sudo/anondo support
- [x] Integrate with z_filewatcher
- [x] Move erlydtl datetime module to new module
- [x] Remove erlydtl sub-directory
- [x] Remove specific erlydtl tests (they are now in the template_compiler repo)
- [x] Change all calls to erlydtl modules
- [x] Check calls to z_template.erl
- [x] Add debugging support to template_compiler
- [x] Add new filter tests (to be added as separate issue)


ddeboer:
Cool! 

You put this on the 1.0 milestone, so am I correct in assuming that it’s no prerequisite for (e.g.) #1076?

mworrell:
@ddeboer The template_compiler uses `maps` and `merl`. So it is tied to the Erlang 18. I think we should expedite the 1.0 release and after fixing the template_compiler add the cowboy integration (and most pressing issues for the 1.0)

ddeboer:
And can we do #1113 separately or has it now become dependent on 1.0 as well?

mworrell:
#1113 can be done independently.

mworrell:
This is now ready to be squashed and merged.

Added issue #1303 for the extra filter (and scomp) tests.

mworrell:
@ddeboer and @arjan can you run some sites with this branch to see if there are any obvious problems?

mworrell:
Almost ready: need to add sudo/anondo support for the include tags.

mworrell:
Added support for `sudo`, `anondo` and `z_language` context variables.

They can be supplied in *include*, *for*, *with* and value statements like `{{ .... with z_language='en' }}`

mworrell:
@ddeboer @arjan @mmzeeman Please check this branch with a sample project. It is now ready to be merged.

I didn't do any performance checks - any performance degradations should be solved later.

ddeboer:
Awesome work, @mworrell! 👊

mmzeeman:
Indeed, it works very nice... Wow.

PR #1310, opened 2016-06-13
docker: Create lite Zotonic container

### LINKED ISSUE #1295: Create slim Docker image
* Create slim image beside the current full one: the slim image should not include PostgreSQL.
* Split development (`build-essential` etc.) and production images
* Add Dockerfile to 0.x branch, too.

@CyBeRoni Any ideas how to create proper production images that don’t include build tools? Of course, to create a deployable container, we do need the tools.

ddeboer:
Fix #1295.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,34 +15,21 @@ ADD https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb .
 RUN apt-get clean \
     && dpkg -i erlang-solutions_1.0_all.deb \
     && apt-get update \
-    && apt-get install -y --no-install-recommends erlang build-essential ca-certificates postgresql imagemagick wget git \
-    && rm -rf /var/lib/apt/lists/* && \
-    useradd --system --create-home zotonic                                                         && \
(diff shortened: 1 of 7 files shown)


mworrell:
@kaos Andreas, can you have a look, as our in-house Docker expert? :)

kaos:
Looks good. 👍 for a "lite" runtime image where the DB has been moved out. That's the right move before using this in production.

ddeboer:
That's exactly the idea, @kaos!

Please don't merge yet as this is not completely finished. I'm still looking for a way to have multiple Dockerfiles but not clutter up the repo's root directory. We can define multiple Dockerfiles on Docker Hub, but if they are in a subdirectory their context is only that directory, not the whole repo.

ddeboer:
Added docs. @kaos and others, please review. I renamed the old `zotonic/zotonic` Docker image to `zotonic/zotonic-full` and removed Postgres from the `zotonic/zotonic` (slim) one.

kaos:
👍

PR #1318, opened 2016-06-21
Update config.in

mset:
Replacing mod_acl_adminonly with mod_acl_user_groups. mod_acl_adminonly has been removed.


ddeboer:
Good catch, cheers! 🍻

PR #1328, opened 2016-06-27
core: Move mod_acl_simple_roles out of the core into separate repo

### LINKED ISSUE #1131: Deprecate adminonly and simple_acl in favour of acl_user_groups

ddeboer:
Moved to https://github.com/zotonic/mod_acl_simple_roles.

Fix #1131.

Does everyone agree on removing mod_acl_simple_roles from the 0.x branch, too?


ddeboer:
@mmzeeman Do you agree if we apply this change to 0.x, too? It means that mod_acl_simple_roles no longer comes as a Zotonic core module so you’ll need to clone it from its separate repo.

mmzeeman:
Fine by me.

I'm going to look at moving to the groups acl after the summer.

mworrell:
I am also ok with moving it out - we just need to communicate clearly and have this module available om the zmr

ddeboer:
Also cherry-picked into 0.x.

PR #1334, opened 2016-07-01
Rename the #context 'host' field to 'site'.

### LINKED ISSUE #1333: Rename `host` to `site`
We are mostly talking about sites, but in the context record the field is still called `host`.

Let's rename this field to `site` (in the 1.0).

In the req model we can keep an alias for backwards compatibility.

mworrell:
Fixes #1333

### DIFF
--- a/modules/mod_translation/mod_translation.erl
+++ b/modules/mod_translation/mod_translation.erl
@@ -56,7 +56,7 @@
 
 %% @doc Make sure that we have the i18n.language_list setting when the site starts up.
 init(Context) ->
-    case Context#context.host of
+    case Context#context.site of
         zotonic_status -> 
             ok;
         _Other ->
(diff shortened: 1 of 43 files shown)


modules/mod_translation/mod_translation.erl
```diff
@@ -56,7 +56,7 @@
 
 %% @doc Make sure that we have the i18n.language_list setting when the site starts up.
 init(Context) ->
-    case Context#context.host of
+    case Context#context.site of
```
arjan: Why not use `z_context:site(Context)` here? Makes future refactorings easier.

mworrell: Done, see 1f111e9

ddeboer:
Thanks!

PR #1342, opened 2016-07-06
Pivot resources using templates.

mworrell:
Issue #1076

This also adds optional re-pivoting on a moment defined in the pivot template.

- [x] documentation (also removed some notifications)

The following is **not** per default in mod_base:

- default pivot template for events (with repivot if date_end in the future, lower prio text if in past)
- higher prio texts if `is_featured` is set

You need to add your own pivot templates for this.


mworrell:
Do we want that Zotonic per default change the text handling in A..D for featured and upcoming/past events?  This could also be part of a separate module, which might contain more domain specific logic.

/cc @arjan @ddeboer

mworrell:
Assigning @ddeboer for checking this.

mworrell:
Ready to be merged.

/cc @arjan @ddeboer

PR #1345, opened 2016-07-08
admin: Add view dropdown menu

ddeboer:
Adds a view dropdown menu with an allincluded template so modules can add specific content types that they serve. This solves two problems:

* users want a ‘view’ button without the ‘save’, so they don’t change the last updated timestamp or accidentally update other data
* users want an easy way to acccess alternative views for the page. 

screen shot 2016-07-25 at 14 20 57

Fix https://github.com/driebit/ginger/issues/49.

### DIFF
--- a/modules/mod_admin/templates/_admin_view_types.tpl
+++ b/modules/mod_admin/templates/_admin_view_types.tpl
@@ -1,0 +1,8 @@
+{% block content_types %}
+    <li><a href="{{ page_url }}">{_ page _}</a></li>
+    {% for content_type, url in id|content_type_urls %}
+        <li><a href="{{ url }}">{{ content_type|content_type_label }}</a>
+    {% endfor %}
+{% endblock %}
+
+{% block extra %}{% endblock %}
--- a/modules/mod_base/controllers/controller_id.erl
+++ b/modules/mod_base/controllers/controller_id.erl
@@ -27,7 +27,8 @@
     previously_existed/2,
     moved_temporarily/2,
     content_types_provided/2,
-    see_other/2
+    see_other/2,
+    get_rsc_content_types/2
 ]).
 
 -include_lib("controller_webmachine_helper.hrl").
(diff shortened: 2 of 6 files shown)


ddeboer:
@mworrell @arjan @mmzeeman @ArthurClemens Do you like the above screenshot? :wink: If so, let’s get this into 0.19.

ArthurClemens:
I can recognise the motivation for adding this functionality. The View dropdown should also contain the default view state ("rendered page") ("as" can be left out).

But the step of adding one more thing to that area makes clear that it has grown beyond its original intent. The complexity of options has become overwhelming. I will add a separate issue for this.

mworrell:
@ArthurClemens Yes, I agree that area needs cleaning up.

Maybe the "view as" could be a dropdown combined with the view button?

ddeboer:
Updated to hook into controller_id’s `content_types_dispatch` notification. On the template level the content types list can be changed by overriding `admin_view_types.tpl`.

@mworrell Please review/merge.

modules/mod_base/controllers/controller_id.erl
```diff
@@ -88,21 +89,26 @@ see_other(ReqData, Context) ->
...
 get_content_types(Context) ->
     case z_context:get(content_types_dispatch, Context) of
         undefined ->
             {Id, Context1} = get_id(Context),
-            CT = z_notifier:foldr(#content_types_dispatch{id=Id}, [], Context1),
-            CT1 = case proplists:get_value("text/html", CT) of
-                    undefined -> [{"text/html", page_url}|CT];
-                    Prov -> [Prov|CT]
-                  end,
+            ?DEBUG(Id),
```
mworrell: Guess this one has to go :)

ddeboer: OOPS. Fixed. :wink:

modules/mod_admin/templates/_admin_view_types.tpl
```diff
@@ -0,0 +1,7 @@
+{% block content_types %}
+    {% for content_type, url in id|content_type_urls %}
```
mworrell: Maybe put the text/html version on top and suppress it in the loop below?  Then we can also give the top one an understandable title like @ArthurClemens suggested: "View as rendered page" (or something similar)

ddeboer: Ok, separated the page dispatch rule. I still prefer to call it ‘page’ over ‘rendered page’ as the term ‘rendered’ doesn’t mean so much to end users.

mworrell:
🌷 🎆

arjan:
doc build error: https://travis-ci.org/zotonic/zotonic/jobs/147237316

ddeboer:
Whoops, will fix it right now.

PR #1355, opened 2016-07-27
Replace Mochiweb/Webmachine with Cowboy/Cowmachine.

### LINKED ISSUE #533: Web server: replace MochiWeb
See #513  for discussion on this

mworrell:
Fixes #533

See also: https://github.com/zotonic/cowmachine 🐮 

- [x] Remove `mochiweb`/`webmachine` http server initialization
- [x] Add `cowboy` http server initialization
- [x] Remove `webmachine` from deps
- [x] Replace `#wm_reqdata{}` with type `mochiweb_req:req()`
- [x] Replace calls to `wrq` with calls to `cowmachine_req`
- [x] Update controllers to simpler version with one `Context` argument
- [x] Change service API to only use a `Context` argument
- [x] Binaryfication of arguments
- [x] Add web socket upgrade
- [x] Cookie setting (new code in cowboy2)
- [x] Get it compiling
- [x] Test, test, test, and test.

Proposal is to move the re-addittion of the access log (`z_access_log`, `z_stats`, `z_access_syslog` and request watchers) to another separate issue.


ddeboer:
Yeah, great start! :cow:

rebar.lock
```diff
@@ -4,6 +4,14 @@
   2},
  {<<"bert">>,{pkg,<<"bert">>,<<"0.1.0">>},0},
  {<<"cf">>,{pkg,<<"cf">>,<<"0.2.1">>},1},
+ {<<"cowboy">>,
+  {git,"git://github.com/ninenines/cowboy.git",
```
arjan: Why no hex.pm dependencies?

rebar.lock
```diff
@@ -4,6 +4,14 @@
...
+       {ref,"0ba3a9a22269d21b2962fec78c03e5671294d20d"}},
+  1},
+ {<<"cowlib">>,
+  {git,"https://github.com/ninenines/cowlib",
+       {ref,"8cf98c9f3d1292c1b08fca5ca99772ae912eaa86"}},
+  2},
```
arjan: aren't cowlib and ranch transitive dependencies of cowboy?

rebar.lock
mworrell: Indeed, we are using the master (2.0 development) version of cowboy.

rebar.lock
mworrell: They are, but apparently rebar lock likes to lock them anyway.

mworrell:
I propose to add the access logs and stats via a different issue.

ddeboer:
And it’s green! :clap:

@mworrell As you’ve just converted Zotonic’s internal controllers, would you mind adding a section to `upgrading.rst` on how users can convert any custom controllers they have? Something like:

1. remove the `ReqDate` argument
2. different way of getting request and setting response headers
3. remove `init` and `services_available` functions
4. ...?

mworrell:
@ddeboer Added docs, also updates some out of date information.
There is also a "Just enough webmachine" chapter, shall we update or remove that?

mworrell:
We should also check chapters like *directory structure* that are not updated with the rebar3 changes.

ddeboer:
> Added docs, also updates some out of date information.

Great!

> There is also a "Just enough webmachine" chapter, shall we update or remove that?

Let’s remove it.

mworrell:
Let's warn the mailing list before we merge, as this is an incompatibel and quite drastic change.

doc/developer-guide/upgrading.rst
```diff
@@ -16,8 +16,88 @@ ACL
...
+Other controller changes changes are:
+
+ * Content types are now binaries in `content_types_accepted` and `content_types_provided`
+ * Character sets are now binaries in `charsets_provided`
+ * Methods are now binaries in `allowed_methods` and `known_methods`
+ * Encodings are now binaries in `content_encodings_provided`
+ * The return value of `generate_etag` must be a binary
+
+
+
```
ddeboer: Thanks, this is a great start. I’ll probably slightly rewrite this later for better developer experience. :wink: I noted that in #1399.

mworrell: We might want to add an example controller in the documentation, showing all possible callbacks.

mworrell:
We have warned the mailing list and on twitter, so I think we can merge.

/cc @ArthurClemens @arjan @mmzeeman

ddeboer:
Let’s give everyone one more day to look at this; if we hear no objections, I’ll squash and merge this tomorrow.

ddeboer:
BOOM! 💣 🔥 🎆 🚒

PR #1356, opened 2016-07-28
core: For every resource id argument, allow resource name as well

### LINKED ISSUE #1357: Where resource id is passed, allow resource name as well
We should be able to use resource ids and names (atoms) interchangeably. This is not yet the case in quite some places.

Please add any places you find here.

ddeboer:
* Solved by calling rid/2

Fix #1357.

### DIFF
--- a/modules/mod_base/support/z_list_of_ids_filter.erl
+++ b/modules/mod_base/support/z_list_of_ids_filter.erl
@@ -22,6 +22,8 @@
 -spec filter(integer() | [integer()], fun((integer()) -> boolean()), any()) -> boolean() | [integer()].
 filter(Id, Fun, _Context) when is_integer(Id) ->
     Fun(Id);
+filter(Id, Fun, _Context) when is_binary(Id) ->
+    Fun(Id);
 filter(List, Fun, Context) ->
     lists:filter(Fun, erlydtl_runtime:to_list(List, Context)).
 
--- a/modules/mod_comment/models/m_comment.erl
+++ b/modules/mod_comment/models/m_comment.erl
@@ -78,19 +78,20 @@ m_value(#m{value=undefined}, _Context) ->
 
 
 %% @doc List all comments of the resource.
-%% @spec list_rsc(int(), Context) -> [ PropList ]
-list_rsc(RscId, Context) when is_integer(RscId) ->
+-spec list_rsc(m_rsc:resource(), #context{}) -> list().
+list_rsc(RscId, Context) ->
     F = fun() ->
-        z_db:assoc_props("select * from comment where rsc_id = $1 order by created asc", [RscId], Context)
+        z_db:assoc_props("select * from comment where rsc_id = $1 order by created asc", [m_rsc:rid(RscId, Context)], Context)
     end,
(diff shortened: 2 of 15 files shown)


src/models/m_rsc.erl
```diff
@@ -120,7 +120,7 @@ m_value(#m{value=Id}, Context) ->
     get_visible(Id, Context).
 
 %% @doc Return the id of the resource with the name
-% @spec name_to_id(NameString, Context) -> {ok, int()} | {error, Reason}
+-spec name_to_id(string(), #context{}) -> {ok, int()} | {error, Reason}.
```
mworrell: Maybe we should have a more generic type for resource names, as we accept: `atom() | binary() | string()`

ddeboer: `m_rsc` already has `resource_name`, so I’ll use that

modules/mod_base/support/z_list_of_ids_filter.erl
```diff
@@ -22,6 +22,8 @@
 -spec filter(integer() | [integer()], fun((integer()) -> boolean()), any()) -> boolean() | [integer()].
 filter(Id, Fun, _Context) when is_integer(Id) ->
     Fun(Id);
+filter(Id, Fun, _Context) when is_binary(Id) ->
+    Fun(Id);
```
ddeboer: Added this `binary` clause to get the doc [example with an integer](http://zotonic.com/docs/latest/ref/filters/filter_is_a.html#apply-to-a-single-resource-id) working.

ddeboer:
@mworrell Please review this one last time so we can merge this.

ddeboer:
Rebased to resolve merge conflicts.

modules/mod_comment/models/m_comment.erl
```diff
@@ -78,19 +78,20 @@ m_value(#m{value=undefined}, _Context) ->
 
 
 %% @doc List all comments of the resource.
-%% @spec list_rsc(int(), Context) -> [ PropList ]
-list_rsc(RscId, Context) when is_integer(RscId) ->
+-spec list_rsc(m_rsc:resource(), #context{}) -> list().
+list_rsc(RscId, Context) ->
     F = fun() ->
-        z_db:assoc_props("select * from comment where rsc_id = $1 order by created asc", [RscId], Context)
+        z_db:assoc_props("select * from comment where rsc_id = $1 order by created asc", [m_rsc:rid(RscId, Context)], Context)
```
ddeboer: I use `m_rsc:rid/2` in most places to retain the old behaviour when an non-existing resource id integer was passed. 

Alternatively, we may want to use `m_rsc:name_to_id_check/2`, which will crash early if the resource does not exist. Which do you prefer?

mworrell: In the case of fetching the list of comments for a nonexistent resource, it should be good to return the empty list. As a nonexistent resource has—by definition—no comments.

ddeboer: Okay, so I’ll keep `rid` in the other places as well, to not change the current behaviour too much.

ddeboer:
Though I’m not sure I’ve tackled all places where both a rsc id and rsc name should be valid as an argument, this is a start, so merging.

PR #1359, opened 2016-08-01
docs: add example configuration for SSL termination using nginx.

mworrell:
Also mention SNI in the https-support section.

Example config was supplied by @maltejk (Thanks!)

https://gist.github.com/maltejk/b7e1739b07f1bd5eb79d59146ea3c1dc


ddeboer:
Thanks!

Issue #499, opened 2013-01-13
Add support for language variations like zh-hk


arjan:
It would be good to have support for more specific locales like zh-hk but also en-us, en-gb, etc.

Lets have this ticket as a discussion on how to do that and what that means for the po files, fallback mechanism (?), etc.

ArthurClemens:
This branch contains a fairly efficient lookup of a language fallback. It is called when a translation cannot be found for the current language. In the current Zotonic, only `en` is considered as hardcoded fallback. The code changes make it possible to define a specific fallback per language, thus progressing one step towards region specific language settings like "fr-be” and "zh-tw”.

The major code change is that `#context.language` now contains a list of language code atoms, instead of a single language code. The first list item is the selected language. To keep other code working as the same as possible, `z_context:language/1` will still return the selected language.
Other code that directly read `Context#context.language` has been replaced with the `z_context:language/1` function call.
`z_context:set_language/2` accepts an atom and puts it in the list, but it will also take a list. That list will be set by `mod_translation` when a language has defined a fallback language.

Translation lookup takes one more lookup in case no translation is found. Fallback lookup is not recursive: only the fallback of the current language is explored. This also removes the need to program against infinite loops.
Currently only 1 fallback is considered, but it is built with a list of fallbacks in mind.
The ultimate fallback is `en` and that functions as it does today.

It looks like there is an obsolete function `z_trans.erl:lookup_fallback_language` that is called by `filter_language`.

mworrell:
Nice work! And we shouldn't be poking directly into the Context#context.language anyway.

:+1: for merging

ArthurClemens:
Things left to do on this issue:

1. Add more languages to `i18n/iso639.erl`, possibly use https://github.com/bdswiss/country-language/blob/master/data.json or http://www.localeplanet.com/icu/iso639.html
1. Change `i18n/z_trans:is_language` to relax the check on 2-letter codes.
1. It seems that the url can only have a 2-letter language code.
1. Check `.po` generation
1. Check filters/scomps

mworrell:
Ad 3: that is also handled by mod_i18n, it can be easily changed to also accept ``xy-...`` prefixes, waarbij ``xy`` then a valid ISO code is and we add a max length to the language variation.

Do we also want to be able to handle languages like ``x-klingon`` ?

ArthurClemens:
The Mozilla L10n project uses this guideline:

To summarize, these schemes are supported:

* ab
* ab-XY
[...]

PR #1371, opened 2016-08-07
Extended language codes

ArthurClemens:
Implements #499

- [x] Core functionality
- [x] Update add/remove language templates
- [x] Check generation of POT files
- [x] Complete language data
- [x] Handle (too) long list of languages
- [x] Update filter_is_rtl.erl to use language data
- [x] Update documentation


ddeboer:
This now conflicts with master. Please rebase (or merge master into this PR).

willemdj:
Hi,

I got a suggestion from the "mention-bot" that this might be something for me to look into. Since I was interested I had a look. Below follow a couple of comments and suggestions.

I must say that I do not have any particular expertise in this area, and I still do not know much about zotonic either. So probably I just scraped the surface a bit. 

In general this looks like a nice addition, and it has obviously been a lot of work.

Since I did not know much about language tags, I googled a bit and found this:

  https://www.w3.org/International/articles/language-tags/

This is probably a better source than what is mentioned in `languages.erl` (in the initial comment).  Note that the W3C article suggests that RFC 3066 is obsolete.

My comments are all related to the new file `languages.erl`:

- I don't think it is a good idea to use the name `languages.erl`: it is too
  general, and it may cause naming conflicts. A better name might be
  `z_languages.erl`.

- Where possible, the module should implement a number of functions that answer specfic
  questions for a certain language tag, rather than return the language data as a proplist. 

  That way the calling modules would not have to know about the structure of the data set. It would make it easier to change this structure in the future, if necessary (the proplist is probably not very efficient - using a  map or a table might be faster).
  
  For example, the function `is_rtl` that is defined in `filter_is_rtl.erl` should be in `languages.erl`.

- It is not 100% clear to me what happens if a combination of language and region is used that is not explicitly defined in the `languages` module.  I have the impression that in those cases there will be no match. This would mean for example that there is no match for a tag like `es-419` (Spanish as used in Latin America), while I would think that it would generally be preferable to match this to Spanish. In any case it is important to realize that the list will probably not be complete, and to clarify what will happen in those cases.

Regards,
Willem

ArthurClemens:
> https://www.w3.org/International/articles/language-tags/
> 
> This is probably a better source than what is mentioned in languages.erl (in the initial comment). Note that the W3C article suggests that RFC 3066 is obsolete.
> 
I agree, it is more recent. But the gist remains the same: keep it simple. Only use extensions when they add value.

> My comments are all related to the new file languages.erl:
> 
> I don't think it is a good idea to use the name languages.erl: it is too
> general, and it may cause naming conflicts. A better name might be
> z_languages.erl.
> 

Good suggestion.

> Where possible, the module should implement a number of functions that answer specfic
> questions for a certain language tag, rather than return the language data as a proplist.
> 
> That way the calling modules would not have to know about the structure of the data set. It would make it easier to change this structure in the future, if necessary (the proplist is probably not very efficient - using a map or a table might be faster).
> 
I am open for suggestions to make it more efficient. So far the entire list is only used for the interface when selecting a language, so that doesn't matter much, but it is also used by z_trans to check if the language code is valid, so that could be optimized.

The check itself should be moved to z_languages.


> For example, the function is_rtl that is defined in filter_is_rtl.erl should be in languages.erl.
> 

I agree.
> 
> It is not 100% clear to me what happens if a combination of language and region is used that is not explicitly defined in the languages module. I have the impression that in those cases there will be no match. This would mean for example that there is no match for a tag like es-419 (Spanish as used in Latin America), while I would think that it would generally be preferable to match this to Spanish. In any case it is important to realize that the list will probably not be complete, and to clarify what will happen in those cases.
> 
I didn't know about es-419. I will add that to the list. Then es will be the fallback language. 
This could be made smarter by parsing the tag and capture 'es' to use as fallback language.
For all cases where no fallback language is found, English is used.

cheers

Arthur

ArthurClemens:
I have made quite some changes to make the data less brittle for future changes. z_language now hides the language data and offers a `properties` function to provide a fixed set of data properties. If the language data structure changes in the future, the properties function will still return a consistent format. The functions `all_languages` and `main_languages` also use `properties`.

mworrell:
I don't think you need a two-way merge. Worst case people can throw away their config and then proceed. In general we only have upgrade functions, no downgrade functions.

ArthurClemens:
ok.

Any idea why Travis fails on `make all edocs`? The log says `sed: can't read ../src/erlydtl/erlydtl_scanner.erl: No such file or directory`

mworrell:
Did you check in that file? Shouldn't be checked in.

ArthurClemens:
No, it is not checked in.

ArthurClemens:
I could create a new branch from master and copy over the changes one by one to see if that doesn't fail.

ddeboer:
I’m going to look into this build failure.

ddeboer:
`sed: can't read ../src/erlydtl/erlydtl_scanner.erl: No such file or directory` happens on master, too. But as you can see [just below it](https://travis-ci.org/zotonic/zotonic/jobs/153396707), it doesn’t cause a failure: `The command "make -C doc/ stubs" exited with 0.`. 

Instead, the failure was caused by a wrong placement of the module `@doc` block. Please have a look at my commit above (ea04814) for the fix. :wink:

src/i18n/z_language.erl
```diff
@@ -0,0 +1,1328 @@
...
+    z_convert:to_atom(m_config:get_value(i18n, language, ?DEFAULT_LANGUAGE, Context)).
+
+
+%% @doc Check if the language code code is a valid language.
+-spec is_valid(Code::binary() | any()) -> boolean().
+is_valid(Code) when is_binary(Code) ->
+    Languages = all_languages(),
+    proplists:get_value(Code, Languages) /= undefined;
+is_valid(Code) ->
+    is_valid(z_convert:to_binary(Code)).
```
ddeboer: Why not accept binaries only instead of doing implicit type conversion? The calling code should convert to binary (if needed).

modules/mod_admin/lib/css/zotonic-admin.css
```diff
@@ -111,9 +115,6 @@ body.noframe > .admin-container {
 .admin-edit-media {
   border: 1px solid #fff;
 }
-iframe, video, audio {
-  max-width: 100%;
-}
```
ddeboer: Why was this removed?

modules/mod_translation/lib/css/mod_translation.css
```diff
@@ -0,0 +1,145 @@
+.dropdown-menu.mod_translation_menu-has-icons > li > a {
+  padding-left: 40px;
```
ddeboer: Please use an indent of 4 spaces (as defined in the `.editorconfig` file).

src/i18n/z_language.erl
ArthurClemens: Yes, we could do that.

modules/mod_admin/lib/css/zotonic-admin.css
ArthurClemens: That code is not part of a LESS file, so on recompile that code is removed from the CSS.

modules/mod_translation/lib/css/mod_translation.css
ArthurClemens: That file is generated. Presumably the LESS to CSS script does not use the editorconfig settings.

modules/mod_admin/lib/css/zotonic-admin.css
ArthurClemens: Fixed.

modules/mod_translation/lib/css/mod_translation.css
ArthurClemens: I cannot find an option to set the number of spaces or read from editorconfig. The recommended way for compression is to install the less plugin [clean-css](https://github.com/less/less-plugin-clean-css) and then call lessc with `--clean-css`. That puts all code on 1 lines.

ArthurClemens:
Finally green! Thanks for the help.

ddeboer:
Yeah! 🎆 

Just to summarise, there are still two todos open before we can merge this:

> - [ ] Check generation of POT files => bug, issue #1377
> - [ ] Check translation tag => feature, issue #1378

alvaropag:
I'm not working on #1378 but I can give a look at it (tomorrow is holiday here).

ArthurClemens:
Thanks @alvaropag 

I will also look at @ddeboer 's comment "Why not accept binaries only instead of doing implicit type conversion? The calling code should convert any non-binaries to binary."

alvaropag:
@ArthurClemens I thank you for letting me help. I'm still learning erlang/zotonic but I want to help improve it in any way I can. The correction I sent was very basic, I still want to look why some modules don't generate the translation/template folder (eg. mod_contact).

PR #1373, opened 2016-08-09
Remove perl dependency

CyBeRoni:
At some point someone thought it was a good idea to fix non-linux compatibility by using perl to do an in-place find-and-replace. Unfortunately that of course brings with it a dependency on perl which is not always available in minimal environments.

This patch re-replaces perl with sed and I believe does so in a POSIX-compatible way.

### DIFF
--- a/src/scripts/helpers/zotonic_setup
+++ b/src/scripts/helpers/zotonic_setup
@@ -40,8 +40,7 @@ export -f require_zotonic_not_running
 function enablesite {
     config="priv/sites/$1/config"
     expr="s/enabled, \(true\|false\)/enabled, $2/"
-    perl -e "$expr" -pi $config
-    #echo sed -e "$expr" -i $config
+    sed -e "$expr" -i.bak -e $config
     echo "$1 enabled: $2"
 }
 export -f enablesite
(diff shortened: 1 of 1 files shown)


src/scripts/helpers/zotonic_setup
```diff
@@ -40,8 +40,7 @@ export -f require_zotonic_not_running
 function enablesite {
     config="priv/sites/$1/config"
     expr="s/enabled, \(true\|false\)/enabled, $2/"
-    perl -e "$expr" -pi $config
-    #echo sed -e "$expr" -i $config
+    sed -e "$expr" -i.bak -e $config
```
ddeboer: Is `sed -i` portable?

arjan: I see that on [BSD](https://www.freebsd.org/cgi/man.cgi?query=sed) it's supported.

arjan: But shouldn't this line read: `sed -e "$expr" -i $config` ?

CyBeRoni: sed -i with an extension works on the platforms I've tried (sunos, osx, linux) but it don't see the 'enablesite' function actually being called anywhere, plus it hardcodes the sites dir as being under priv/ so I didn't give it as much thought as the main config file. I actually considered removing it entirely.

arjan: Agreed, it might be a leftover from long ago, you have my blessing to remove it :)

CyBeRoni:
Ok I removed that old enablesite function.

ddeboer:
Thanks, @CyBeRoni!

PR #1374, opened 2016-08-09
Use Alpine Linux for the Docker images

CyBeRoni:
This PR changes the Docker images from Debian to Alpine, greatly reducing their total size. 

I consider this a healthy base for a Docker image to be used in a production environment -- though for that use-case you will likely want to extend from it and add your own configuration, sites, modules and whatnot, rather than running it directly.

I've also re-worked (and re-named) the non-lightweight container including PostgreSQL to extend from the lightweight image. It adds postgres of course, and since it runs multiple processes, it uses s6 for process supervision.

The base image ships with an erlang.config which disables most logging to prevent rotated-but-never-deleted access logs from filling up disks left and right.


ddeboer:
Alpine ftw!

Dockerfile.heavy
```diff
@@ -0,0 +1,15 @@
+FROM zotonic/zotonic:latest
```
ddeboer: Why rename full to heavy?

CyBeRoni: "Full" implies the regular one is incomplete.

Dockerfile
```diff
@@ -1,35 +1,29 @@
-FROM debian
-MAINTAINER Andreas Stenius git@astekk.se
+FROM alpine:3.4
+MAINTAINER Marco Wessel <marco@driebit.nl>
```
ddeboer: Let’s either remove this or add a generic Zotonic e-mail address.

Dockerfile
```diff
@@ -1,35 +1,29 @@
...
 
-ADD https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb .
 
-RUN apt-get clean \
-    && dpkg -i erlang-solutions_1.0_all.deb \
-    && apt-get update \
-    && apt-get install -y --no-install-recommends erlang build-essential ca-certificates imagemagick inotify-tools libnotify-bin wget git \
-    && rm -rf /var/lib/apt/lists/*
+# Note: dumb-init and gosu are pulled from edge; remove that when upgrading to an alpine release that
+# includes those packages.
```
ddeboer: Why do we need gosu anyway?

Dockerfile.heavy
```diff
@@ -0,0 +1,15 @@
...
+
+# This Docker image extends the regular Zotonic image to add
+# a PostgresQL database and, since multiple processes are run,
+# process supervision. When run, it considers Zotonic the "main"
+# process, and exits if it does.
+
+RUN apk --no-cache add postgresql postgresql-client s6 \
+    && gosu postgres initdb -D /var/lib/postgresql -U zotonic
```
ddeboer: Can we do without any supervisor such as s6? Shipping the image with a supervisor feels overly opinionated.

docker/erlang.config
```diff
@@ -0,0 +1,76 @@
...
+    {[erlang, memory], {function, erlang, memory, [], value, []}, []},
+    {[erlang, system_info], {function, erlang, system_info, ['$dp'], value, [process_count]}, []},
+    {[erlang, statistics], {function, erlang, statistics, ['$dp'], value, [run_queue]}, []},
+    {[erlang, io], {function, erlang, statistics, [io], match, {{'_', input}, {'_', output}}}, []}
+   ]}
+ ]},
+
+ {mnesia, [
+    {dir, "/tmp/mnesia"}
+ ]},
```
ddeboer: Is the mnesia dir the only change from erlang.config.in? We may then want to `sed` for this file too.

Dockerfile
CyBeRoni: Better safe than sorry. Running as root is generally not a good idea, regardless of whether or not you're in a container. Linux cgroups is not yet (and probably won't ever be) 100% secure.

Dockerfile.heavy
CyBeRoni: No, because two processes need to be started and something needs to handle signalling and zombie reaping. S6 is small and simple.

docker/erlang.config
CyBeRoni: No.

ddeboer: So what are the other changes?

Dockerfile.heavy
ddeboer: Agreed as this adds s6 only to the heavy image, which is already bloated. 😝

docker/erlang.config
CyBeRoni: Read the description in the PR and find out!

ddeboer: I mean of course: which config keys have been changed? If it’s only a small number, why not sed them, analogously to `zotonic.config`?

Dockerfile
CyBeRoni: Also, this allows me to set file-system permissions that actually do something, in an effort to further increase security. My intention is, at some point, to make the zotonic directory as much as possible read-only for the zotonic user. (Currently priv is writable.)

docker/erlang.config
CyBeRoni: Some are removed (console log), others are commented out (access log). I suppose since both those lines could be removed instead of commented, that a sed /foo/d would work for them. 

Truthfully though I'm not a huge fan of sed'ing config files like these and do it for zotonic.conf only because values need to be inserted. That said I'm also not a fan of shipping config files so I guess I'll sed the file into compliance.

.dockerignore
```diff
@@ -2,3 +2,17 @@
...
+Dockerfile
+Dockerfile.*
+LICENSE
+Makefile
+prepare-release.sh
+*.md
+start.*
+TRANSLATORS
+zotonic_install
+priv/emacs
```
ddeboer: Better to invert this: specify the dirs we *don’t* want to ignore, using `!...`.

docker/docker-entrypoint.sh
```diff
@@ -1,11 +1,27 @@
...
+	sed -e s/%%GENERATED%%/${ZOTONIC_PASSWORD-changeme}/ \
+	    -e s,%%USER_SITES_DIR%%,/opt/zotonic/user/sites, \
+	    -e s,%%USER_MODULES_DIR%%,/opt/zotonic/user/modules, \
+	    -e s,%%USER_EBIN_DIR%%,/opt/zotonic/user/ebin, /opt/zotonic/priv/zotonic.config.in > /etc/zotonic/zotonic.config 
+
+	exec /usr/bin/gosu zotonic /opt/zotonic/bin/zotonic "$@"
 else
-    exec "$@"
+	exec "$@"
 fi
```
ddeboer: Spaces, not tabs please.

Dockerfile
```diff
@@ -1,35 +1,28 @@
...
+RUN apk add --virtual build-deps --no-cache ca-certificates wget curl make gcc musl-dev g++ git \
+        && apk add --no-cache bash imagemagick \
+	&& apk add --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ dumb-init gosu \
+           erlang erlang-inets erlang-compiler erlang-crypto erlang-mnesia erlang-ssl erlang-stdlib erlang-public-key erlang-tools erlang-dev erlang-asn1 erlang-syntax-tools erlang-eunit erlang-parsetools erlang-snmp erlang-sasl erlang-xmerl erlang-erl-interface \
+        && DEBUG=1 make \
+        && apk del build-deps
```
ddeboer: This must be changed because `dumb-init` and `erlang` have moved from testing to community.

Dockerfile
```diff
@@ -1,35 +1,28 @@
...
 
-ADD . /opt/zotonic
-WORKDIR /opt/zotonic
+RUN mkdir /etc/zotonic \
+	&& sed -f docker/erlang.config.sed priv/erlang.config.in > /etc/zotonic/erlang.config \
+	&& adduser -S -h /tmp -H -D zotonic \
+	&& chown -R zotonic /opt/zotonic/priv
```
ddeboer: No trailing spaces please: see our `.editorconfig`.

Dockerfile
ddeboer: While we want to remove build-deps for production containers, don’t we need to keep them for development containers so that changes in `user/{modules|sites}` can be compiled?

Dockerfile
```diff
@@ -1,35 +1,28 @@
...
-ENV DEBIAN_FRONTEND noninteractive
-
-# httpredir.debian.org/debian fails too much, so replace it with a fixed mirror.
-RUN echo \
-   'deb ftp://ftp.nl.debian.org/debian/ jessie main\n \
-    deb ftp://ftp.nl.debian.org/debian/ jessie-updates main\n \
-    deb http://security.debian.org jessie/updates main\n' \
-    > /etc/apt/sources.list
+ADD . /opt/zotonic
+WORKDIR /opt/zotonic
```
ddeboer: Better move this below package installation because of Docker’s build layer caching.

CyBeRoni: That doesn't work at all.

ddeboer: It works perfectly well, and makes the builds way faster. You only have to move `DEBUG=1 make \ && apk del build-deps` below this.

Dockerfile
CyBeRoni: This is designed to be a base for a production container. Development is out-of-scope.

CyBeRoni: Both dumb-init and gosu are still only in `edge`.

Dockerfile
CyBeRoni: I don't care if the build is faster, I care about the resulting image size which would balloon if I did that.

I will leave this as it is.

CyBeRoni:
> Replace spaces with tabs

Eh, I messed up the commit message but the reverse was done to the file.

mmzeeman:
👍 I was already looking into using Alpine for a bare metal server. Any experience with managing large numbers of incoming connections there. Ubuntu has such terrible defaults. Any difference with Alpine?

ddeboer:
> Eh, I messed up the commit message but the reverse was done to the file.

We get the idea. :wink: We’ll squash commits anyway.

Dockerfile
ddeboer: You’re right, built the two variants locally and it’s a 236 vs 400 MB difference.

.dockerignore
CyBeRoni: I disagree. I'm listing here explicitly the files that I *know* don't care about. Any files not listed (that may pop up in future commits and may be necessary for the system) I can't be sure about so I let them be included until such time as their usefulness can be ascertained.

ddeboer: There’s definitely something to be said for that more conservative viewpoint.

However, I still prefer a `.dockerignore` that using the negation pattern for three reasons:

1. One of the purposes of using Alpine is to bring down the image size as much as we can (under 194 MB now using negation).
2. The directory structure is relatively stable. Usually, files that are necessary for the build will end up in one of the already existing (and properly unignored) directories.
3. If some file turns up elsewhere (and it will sometime, of course), the Docker build failures will tell us clearly anyway.


I’ll switch to negation for now; if this turns out to be unmaintainable, I’ll revert back to your solution.

ddeboer:
This is almost ready to merge now!

@CyBeRoni The db setup doesn’t work out of the box yet for the heavy container:

```
$ docker run … runtests
…
user "zotonic" cannot connect to the 'postgres' database: econnrefused
```

So we need to add something [like this](https://github.com/zotonic/zotonic/blob/master/Dockerfile.full#L7-L13).

CyBeRoni:
I'll take a look at that later today or tomorrow.

ddeboer:
Alternatively, we ditch the full/heavy image #1420). In production setups, it’s silly to combine Zotonic and database in one container anyway. Or would one then miss the supervisor capabilities added here to the heavy image?

CyBeRoni:
> In production setups, it’s silly to combine Zotonic and database in one container anyway.

I don't necessarily agree with that. The idea with containers is usually to put a single *service* in a container, which Docker explains as a single *process*. That explanation has its merits but also certain drawbacks. Other container platforms tend to take a different approach, typically running more than one process. 

I have no intention of using the heavy image myself but I can certainly see it having use for others.

> Or would one then miss the supervisor capabilities added here to the heavy image?

The supervisor in the heavy image restarts postgres should it crash or otherwise exit, but stops the container entirely (and should do it cleanly) if zotonic crashes or otherwise exists so that it can be externally restarted. I'm not sure you can do that with docker natively.

ddeboer:
Thanks! 💥

.dockerignore
ddeboer: See #1426.

PR #1379, opened 2016-08-10
mod_base: Fix Content-Security-Policy for inline PDFs

ddeboer:
Inline content-disposed PDFs no longer render because of the Content-Security-Policy header recently added to `controller_file`. See for instance https://tudelft.openresearch.net/image/2016/8/8/ri_2016_minor_calendar_q1.pdf. This fixes that by allowing browser PDF plugins to render the file.

### DIFF
--- a/modules/mod_base/controllers/controller_file.erl
+++ b/modules/mod_base/controllers/controller_file.erl
@@ -161,6 +161,14 @@ set_allow_origin(ReqData) ->
 
 set_content_policy(#z_file_info{acls=[]}, ReqData) ->
     ReqData;
+set_content_policy(#z_file_info{acls = Acls, mime = <<"application/pdf">>}, ReqData) ->
+    case lists:any(fun is_integer/1, Acls) of
+        true ->
+            RD1 = wrq:set_resp_header("Content-Security-Policy", "object-src 'self'; plugin-types application/pdf", ReqData),
+            wrq:set_resp_header("X-Content-Security-Policy", "plugin-types: application/pdf", RD1);
+        false ->
+            ReqData
+    end;
 set_content_policy(#z_file_info{acls=Acls}, ReqData) ->
     case lists:any(fun is_integer/1, Acls) of
         true ->


modules/mod_base/controllers/controller_file.erl
```diff
@@ -8,17 +8,17 @@
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
 %% You may obtain a copy of the License at
-%% 
+%%
 %%     http://www.apache.org/licenses/LICENSE-2.0
-%% 
+%%
```
arjan: Would it be possible to strip the whitespace of all source files in one commit? Otherwise we'll be seeing these sanitary changes for a long time.

ddeboer: Good point, I’ll look into that.

ddeboer: #1380.

ddeboer:
@mworrell Please have a look. Should we release this as hotfix 0.19.2?

mworrell:
Good fix and an unexpected side effect of the security policy.

Maybe (in due time, if this happens more often) we need a list with mappings from content-type to CSP headers 😨

mworrell:
I think for Driebit it would be good to have this as a hot fix.

ddeboer:
Alright, please merge if okay.

> Maybe (in due time, if this happens more often) we need a list with mappings from content-type to CSP headers 

Yeah, I’m no CSP expert myself, either. More generally we may want to think about a way to *test* changes like this: a GET request on some PDF in the testsandboxdb site, for instance?

mworrell:
Yes, fetching files from our own running server sounds ok to me.

At some point we might run into the max execution time on Travis, at that moment we could remove the edocs build.

ddeboer:
Well, I like to catch doc build failures, too. There’s always the option to move the docs to a separate repo, or do some other smart things. :wink:

ddeboer:
Tagged as 0.19.2.

PR #1382, opened 2016-08-13
build: set all dependencies to https that's easier on proxies

sistema-correspondente:
Changed all dependencies to https to help when there's a proxy.


ddeboer:
Good one! 

For those interested: GitHub [recommends](https://help.github.com/articles/which-remote-url-should-i-use/#cloning-with-https-urls-recommended) cloning over HTTPS.

PR #1385, opened 2016-08-14
doc: Show todos on website

ddeboer:
We used to hide the doc todo blocks, but I agree with the [Producing OSS book](http://producingoss.com/en/getting-started.html#documentation) that encourages to 

> Label the areas where the documentation is known to be incomplete.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -4,7 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
-PAPER         = 
+PAPER         =
 BUILDDIR      = _build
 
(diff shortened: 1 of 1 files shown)


ddeboer:
Do you agree with showing the todo blocks in the docs publicly? If so, please merge this. 

It’s not ideal, of course, but as we still have gaps in the docs, it’s preferable to show readers that we’re aware of them.

mworrell:
@ddeboer these are the TODOs from the docs, and not from the Erlang source code, aren't they?  if they are from the docs then I am ok to show them.

ddeboer:
Yep, only the todos from the docs.

PR #1389, opened 2016-08-21
mod_authentication: Rename menu to 'External Services'

ddeboer:
Rename 'App Keys and Authentication Services' to 'External Services'.
It's part of the Authentication menu, so we can be more concise.


ArthurClemens:
I am for.

ddeboer:
@mworrell Ok with you, too?

Issue #1176, opened 2016-02-02
Strange sasl message at zotonic boot time


mmzeeman:
When zotonic starts the following message is logged.

```
11:00:46.961 [error] Unable to read boot script (start_sasl.script): {error,enoent}
```

and it looks like this also causes an extra directory named `log.zotonic001@Buka` to be created. 

Is there something wrong with the order things are started?

ddeboer:
Yeah, we’re seeing this to. I think @mworrell mentioned this was due to a dep on master. @mworrell Do you remember which one?

mworrell:
Yes, `setup`.  I also saw it mentioned on the Erlang mailinglists.

mworrell:
Not seeing this on master anymore, might have been fixed in the config or in `setup`.

PR #1390, opened 2016-08-22
Use erlpass for calculating password hashes

### LINKED ISSUE #1386: Use erlpass for calculating password hashes
For security reasons it is needed to upgrade our current password hash algorithm (salted sha1) to something more secure.

For backward compatibility old salted sha1 hashes need to be updated to bcrypt after a successful user login.

mmzeeman:
Fix #1386.

This improves the security of passwords stored by zotonic. It now uses bcrypt instead of salted sha1 password hashes. Passwords which currently use salted sha1 hashes are converted to bcrypt hashes after a user has successfully authenticated themselves.


src/models/m_identity.erl
```diff
@@ -542,22 +543,34 @@ get_rsc(Id, Type, Context) ->
...
-    {hash, Salt, Hash}.
+    {bcrypt, erlpass:hash(Pw)}.
 
 
+%% @doc Hash a password, with bcrypt
+%% @spec bcrypt_hash(Password) -> tuple
+
 %% @doc Compare if a password is the same as a hash.
 %% @spec hash_is_equal(Password, Hash) -> bool()
+hash_is_equal(Pw, {bcrypt, Hash}) ->
```
ddeboer: I think this duplication of `@doc` and `@spec` blocks is an accident? It’s causing the build failure on Travis.

mmzeeman: Oops.. didn't spot that one before committing.

ddeboer: No problem. If you fix it, I’m all in favour of merging this quickly so we can ship it with the next release. 

Perhaps we should add a test though, as any mistakes in this change will cause users to be unable to log in.

mmzeeman: Indeed, looking into that later today.

mmzeeman: BTW, when I run the tests locally I see a lot of warnings. I'm not sure if this works or not.

```
(zotonic001_testsandbox@Edo)1> 10:07:52.494 [info] No action enabled for "growl"
(zotonic001_testsandbox@Edo)1> 10:07:52.494 [error] [testsandboxdb] Could not start mod_authentication: Missing dependencies: [acl]
(zotonic001_testsandbox@Edo)1> 10:07:52.494 [info] No action enabled for "growl"
(zotonic001_testsandbox@Edo)1> 10:07:52.494 [error] [testsandboxdb] Could not start mod_admin: Missing dependencies: [authentication]
(zotonic001_testsandbox@Edo)1> 10:07:52.494 [info] No action enabled for "growl"
(zotonic001_testsandbox@Edo)1> 10:07:52.494 [error] [testsandboxdb] Could not start mod_admin_identity: Missing dependencies: [admin]
(zotonic001_testsandbox@Edo)1> 10:07:52.494 [info] No action enabled for "growl"
(zotonic001_testsandbox@Edo)1> 10:07:52.494 [error] [testsandboxdb] Could not start mod_admin_category: Missing dependencies: [admin]
(zotonic001_testsandbox@Edo)1> 10:07:52.495 [info] No action enabled for "growl"
(zotonic001_testsandbox@Edo)1> 10:07:52.495 [error] [testsandboxdb] Could not start mod_admin_predicate: Missing dependencies: [admin]
(zotonic001_testsandbox@Edo)1> 10:07:52.495 [info] No action enabled for "growl"
(zotonic001_testsandbox@Edo)1> 10:07:52.495 [error] [testsandboxdb] Could not start mod_admin_modules: Missing dependencies: [admin]
(zotonic001_testsandbox@Edo)1> 10:07:52.495 [info] No action enabled for "growl"
(zotonic001_testsandbox@Edo)1> 10:07:52.495 [error] [testsandboxdb] Could not start mod_admin_config: Missing dependencies: [admin]
```

And in the end...

```
(zotonic001_testsandbox@Edo)1>   All 23 tests passed.
(zotonic001_testsandbox@Edo)1> Terminating due to shutdown
(zotonic001_testsandbox@Edo)1> 
```

Is that normal?

ddeboer: As the tests still pass, I *think* you can ignore these errors for now. We also have them [on Travis](https://travis-ci.org/zotonic/zotonic/jobs/154162331): they seem to be caused by a boot order issue. Cf. #1176.

mmzeeman: The following three lines must be added before `mod_authentication` to the configs of the test sites.

```
mod_menu,
mod_content_groups,
mod_acl_user_groups,
```

mmzeeman: This was probably caused because `mod_acl_adminonly` was deprecated?

mmzeeman: Strange, that the tests work, because one of the tests being done in `testsandbox_basic_tests` is checking if mod_admin is active. :-/

mmzeeman: There are more problems with the tests. For instance, it looks like `m_rsc_db_tests` is only compiled, but not run. Also simply adding embedded unit-tests doesn't do anything either. To be able to test upgrading the hash algorithm I need a db tests. 

So I will only make tests for things I can test without a working database connection.

ddeboer: `testsandbox_basic_tests` doesn’t use testsandboxdb but testsandbox.

`*_db_tests` seems to be [excluded](https://github.com/zotonic/zotonic/blob/master/src/scripts/zotonic-runtests#L37) from the test run, probably because it requires a database. Is this the file that you were adding tests to?

mmzeeman: Thanks for finding this. 

No, I first embedded simple unit tests in `m_identity.erl` only to find out these kind of tests are not run. Then I added a new test module named `m_identity_db_tests.erl` to `src/models/tests`  because I also wanted to test the hash upgrading mechanism only to find out the tests implemented there are skipped.

ddeboer:
@mmzeeman Fixed in #1400. Please review.

src/models/m_identity.erl
```diff
@@ -542,22 +543,34 @@ get_rsc(Id, Type, Context) ->
...
 hash_is_equal(Pw, {hash, Salt, Hash}) ->
     NewHash = crypto:hash(sha, [Salt, Pw]),
     Hash =:= NewHash;
 hash_is_equal(_, _) ->
     false.
 
+
+%% @doc Check if the password hash needs to be rehashed.
+%% @spec needs_rehash(Hash) -> bool()
+needs_rehash({bcrypt, _}) ->
```
ddeboer: Please use `-spec needs_rehash(Hash) -> bool()` instead

src/models/m_identity.erl
```diff
@@ -542,22 +543,34 @@ get_rsc(Id, Type, Context) ->
...
+    erlpass:match(Pw, Hash);
 hash_is_equal(Pw, {hash, Salt, Hash}) ->
     NewHash = crypto:hash(sha, [Salt, Pw]),
     Hash =:= NewHash;
 hash_is_equal(_, _) ->
     false.
 
+
+%% @doc Check if the password hash needs to be rehashed.
+%% @spec needs_rehash(Hash) -> bool()
```
ddeboer: Please use -spec needs_rehash(Hash) -> bool() instead

src/models/m_identity.erl
ddeboer: I fixed the tests in #1393, so if you rebase this on master, you should be able to rebase this on master and then add the your tests to `src/tests`.

ddeboer: If you still have problems with running your test, please push it to this branch so I can have a look and help out.

mmzeeman: Going to work on it right now.

mmzeeman:
Committed, but of course it has conflicts because of the recent cowboy branch merge.

ddeboer:
Merged master into this; let’s see whether we get a green light. :wink:

src/models/tests/m_identity_tests.erl
```diff
@@ -0,0 +1,107 @@
...
+hash_is_equal_old_hash_test() ->
+    PasswordHash = old_hash("password"),
+
+    % Validating stored old password hashes should still work.
+    ?assert(m_identity:hash_is_equal("password", PasswordHash)),
+    ?assert(not m_identity:hash_is_equal("1234", PasswordHash)),
+
+    ok.
+
+check_password_no_user() ->
```
ddeboer: This is not run as it doesn’t have the `_test` function extension. Can you change that?

ddeboer:
Okay, merged master (including a fix to the tests after the Cowboy merge), so the failures shown here properly belong to m_identity_tests now.

src/models/tests/m_identity_tests.erl
```diff
@@ -0,0 +1,107 @@
...
+    ?assertEqual({error, nouser}, m_identity:check_username_pw("mr_z", "does-not-exist", C)),
+    ok.
+
+check_username_password_test() ->
+    C = z_context:new(testsandboxdb),
+    AdminC = z_acl:logon(?ACL_ADMIN_USER_ID, C),
+
+    MrYId = ensure_new_user("mr_y", "secret", AdminC),
+
+    ?assertEqual({ok, MrYId}, m_identity:check_username_pw("mr_y", "secret", C)),
```
ddeboer: This fails with `{error,nouser}`.

mmzeeman: Strange, the tests run over here. 

It looks like out of data output because of the module dependency warnings. I don't get those when I run the tests locally.

mmzeeman:
```
19:42:13.421 [info] DEBUG: m_identity_tests:51  [{1,1,<<"username_pw">>,<<"admin">>,true,false,undefined,{bcrypt,<<"$2a$12$m2m6y1ic9kS078J6nQyzeOka8vPKXYGwafH7QiEmyewSMeDIRxkYC">>},<<>>,<<>>,<<>>,{{2016,8,30},{19,41,59}},{{2016,8,30},{19,41,59}},undefined},{2,314,<<"username_pw">>,<<"mr_y">>,true,true,undefined,{bcrypt,<<"$2a$12$Acuirmi7xaAEBgdV3b8O5.YsF.PLZAiQUPTWmzuA856Y5eUlA1Ctu">>},<<>>,<<>>,<<>>,{{2016,8,30},{19,42,12}},{{2016,8,30},{19,42,12}},undefined}]
19:42:13.421 [info] DEBUG: m_identity_tests:52  [{1,1,<<"username_pw">>,<<"admin">>,true,false,undefined,{bcrypt,<<"$2a$12$m2m6y1ic9kS078J6nQyzeOka8vPKXYGwafH7QiEmyewSMeDIRxkYC">>},<<>>,<<>>,<<>>,{{2016,8,30},{19,41,59}},{{2016,8,30},{19,41,59}},undefined},{2,314,<<"username_pw">>,<<"mr_y">>,true,true,undefined,{bcrypt,<<"$2a$12$Acuirmi7xaAEBgdV3b8O5.YsF.PLZAiQUPTWmzuA856Y5eUlA1Ctu">>},<<>>,<<>>,<<>>,{{2016,8,30},{19,42,12}},{{2016,8,30},{19,42,12}},undefined}]
m_identity_tests: check_username_password_test...*failed*
in function m_identity_tests:'-check_username_password_test/0-fun-0-'/2 (/home/travis/build/zotonic/zotonic/_build/default/lib/zotonic/src/models/tests/m_identity_tests.erl, line 54)
in call from m_identity_tests:check_username_password_test/0 (/home/travis/build/zotonic/zotonic/_build/default/lib/zotonic/src/models/tests/m_identity_tests.erl, line 54)
**error:{assertEqual,[{module,m_identity_tests},
              {line,54},
              {expression,"m_identity : check_username_pw ( \"mr_y\" , \"secret\" , C )"},
              {expected,{ok,314}},
              {value,{error,nouser}}]}
  output:<<"">>
```

Weird, the check_username_pw returns nouser, but the identity table contains the user data.

@mworrell, do you have an idea what the problem might be? On my home and work machines the tests work just fine.

src/models/m_identity.erl
mworrell: I think we can do a single big update of specs?
Or at least let's add (and check) specs when we update Erlang modules.

BTW `bool()` should be `boolean()`

ddeboer: > Or at least let's add (and check) specs when we update Erlang modules.

Definitely! That’s why we’re nitpicking here on the spec definition that @mmzeeman wrote. :smile:

src/models/m_identity.erl
```diff
@@ -358,9 +364,9 @@ check_username_pw("admin", Password, Context) ->
             {error, password}
     end;
 check_username_pw(Username, Password, Context) ->
-    Username1 = z_string:trim(z_string:to_lower(Username)),
+    Username1 = ?DEBUG(z_string:trim(z_string:to_lower(Username))),
```
ddeboer: Interestingly, this value is a binary (not a ‘string’). Is that a problem?

mmzeeman: No, `mod_admin_identity` was running on my local machine for `testsandboxdb`. It was probably started. The schema of the test site should be dropped before running the tests.

mmzeeman:
@ddeboer The tests fail because this test needs mod_admin_identity -> mod_admin -> mod_authentication -> acl module. We currently only have mod_acl_user_groups, which needs mod_content_group and other things.

What happens is that the tests start running before all modules are started.

mmzeeman:
@ddeboer To fix this for now I could add a dummy acl module to testsandboxdb so I can start `mod_authentication` without all the extra deps. But even then this is a race condition for all tests. The tests should just wait until all modules are started.

mmzeeman:
@ddeboer At some point in time `mod_acl_user_groups` must be tested too I guess.

ddeboer:
`mod_acl_user_groups` is already [being tested](https://github.com/zotonic/zotonic/blob/master/modules/mod_acl_user_groups/tests/mod_acl_user_groups_tests.erl) (albeit minimally); you can use [`activate_await`]https://github.com/zotonic/zotonic/blob/master/modules/mod_acl_user_groups/tests/mod_acl_user_groups_tests.erl#L67) to start a module in the tests and wait for it to be started.

mmzeeman:
Check, didn't know that one. If that is the case then we just have to start mod_base I think, all the other modules have dependencies.

ddeboer:
True, but you can just leave the list of modules in `config` unchanged for now.

mmzeeman:
@ddeboer It is maddening.. The tests now run on a clean testsandboxdb schema, but... you still get warnings about mod_authentication not being started. The reason for this is that z_installer depends on mod_authentication... which needs an acl module... 

Why does the installer need mod_authentication?

ddeboer:
Please push, so I can follow along. :wink:

ddeboer:
Right, you’re talking about https://github.com/zotonic/zotonic/blob/master/src/install/z_installer.erl#L425. I guess this is a check that ensures that at least some authentication is going on in the site, to protect the data from being unintentionally publicly available.

@mworrell Is that correct? What about removing that check?

mmzeeman:
Ok, now we have a timeout... is this progress?

mmzeeman:
Trying if I can run the tests without installing content groups

priv/sites/testsandboxdb/config
```diff
@@ -18,8 +18,6 @@
 
     {modules, [
         testsandboxdb,
-        mod_base,
-        mod_authentication,
-        mod_admin
+        mod_base
```
ddeboer: Can you leave this unchanged for now?

ddeboer:
Possibly a Travis hick-up, as restarting the test makes it green.

priv/sites/testsandboxdb/config
mmzeeman: Yup, I'll undo that.

mmzeeman:
In order to minimize the deps needed to run the password check database tests I created a mock acl module.

mmzeeman:
Now it fails on one of the mod_acl_user_group tests. Probably a race condition.

ddeboer:
Yeah, we get that one now and then on master, too, so not related to this PR. Just restarted the test.

ddeboer:
Also cherry-picked into 0.x.

PR #1391, opened 2016-08-23
mod_import_csv: Move CSV import to Content menu

### LINKED ISSUE #1350: CSV import should not require mod_admin_config privileges
### Problem

The CSV import button is currently hidden away under menu System > Status.

* This is unclear: importing files has nothing to do with system status.
* Showing that screen requires `{use, mod_admin_config}` ACL privileges, which then gives unwanted access to flushing and other system operations as well.

### Proposal

1. Move the import CSV button to another place: under ‘Content’ or under ‘Modules’?
2. Only depend on `{use, mod_import_csv}`.

ddeboer:
Fix #1350.

* Don't require use mod_admin_config permission, but only check use mod_import_csv.
* Create separate 'Import content' page with a menu entry under 'Content'.
* Remove 'Import CSV' button from the Status page.
* Add Dutch translation.

### DIFF
--- a/modules/mod_import_csv/templates/_admin_import.tpl
+++ b/modules/mod_import_csv/templates/_admin_import.tpl
@@ -1,0 +1,15 @@
+{% extends "admin_base.tpl" %}
+
+{% block title %}{_ Import content _}{% endblock %}
+
+{% block content %}
+
+    <div class="admin-header">
+        <h2>{_ Import content _}</h2>
+    </div>
+
+    <div class="well">
+        {% all include "_admin_import_button.tpl" %}
+    </div>
+
+{% endblock %}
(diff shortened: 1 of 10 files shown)


modules/mod_import_csv/templates/_admin_import.tpl
```diff
@@ -0,0 +1,15 @@
...
+{% block title %}{_ Import content _}{% endblock %}
+
+{% block content %}
+
+    <div class="admin-header">
+        <h2>{_ Import content _}</h2>
+    </div>
+
+    <div class="well">
+        {% all include "_admin_import_button.tpl" %}
```
ddeboer: This is a way for other modules to add import buttons (e.g. Excel or WordPress import).

PR #1393, opened 2016-08-23
tests: Include db_tests in testrun

### LINKED ISSUE #1392: Don't exclude db tests

ddeboer:
Fix #1392.

* Make sure all tests are run with runtests, including db_tests.
* Fix previously excluded tests that now failed.
* Start minimum number of modules on testsandbox and remove mod_acl_adminonly from modules list.

### DIFF
--- a/src/models/tests/m_rsc_db_tests.erl
+++ b/src/models/tests/m_rsc_db_tests.erl
@@ -11,8 +11,9 @@ modify_rsc_test() ->
     AdminC = z_acl:logon(?ACL_ADMIN_USER_ID, C),
     CatId = m_rsc:rid(text, C),
 
-    ?assertThrow({error, eacces}, m_rsc:insert([{title, "Hello."}], C)),
-    ?assertThrow({error, eacces}, m_rsc:insert([{title, "Hello."}, {category_id, CatId}], C)),
+%%    ?assertThrow({error, eacces}, m_rsc:insert([{title, "Hello."}], C)),
+    ?assertException(throw, {error, eacces}, m_rsc:insert([{title, "bla"}], C)),
+%%    ?assertThrow({error, eacces}, m_rsc:insert([{title, "Hello."}, {category_id, CatId}], C)),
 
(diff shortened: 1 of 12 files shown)


ddeboer:
We now have some [tests failing](https://travis-ci.org/zotonic/zotonic/jobs/154411566). It looks like they have been disabled for quite some time. @mworrell Have they worked before?

ddeboer:
Wouldn’t that also run the `erlydtl` tests that we want to ignore?

mmzeeman:
Why do we want to ignore tests?

mmzeeman:
The stable branch doesn't have eunit tests in erlydtl.. That was probably removed earlier. 

The test spec {application, zotonic} just runs the tests found in zotonics ebin directory. Maybe we should add {application, z_stdlib} too.

ddeboer:
Agreed, let’s change the default for `$ALL` to `{application, zotonic}`.

ddeboer:
The only problem with changing to `{application, zotonic}` is that it will also include tests in your custom code, including site tests. That’s rather inconvenient if you just want to test Zotonic.

src/models/tests/m_rsc_db_tests.erl
```diff
@@ -11,8 +11,8 @@ modify_rsc_test() ->
     AdminC = z_acl:logon(?ACL_ADMIN_USER_ID, C),
     CatId = m_rsc:rid(text, C),
 
-    ?assertThrow({error, eacces}, m_rsc:insert([{title, "Hello."}], C)),
-    ?assertThrow({error, eacces}, m_rsc:insert([{title, "Hello."}, {category_id, CatId}], C)),
+    ?assertThrow({{error, eacces}, _Trace}, m_rsc:insert([{title, "Hello."}], C)),
```
ddeboer: @mworrell I’m not sure this is what we want to return; wouldn’t just `{error, eacces}` be better?

mworrell: Yes, that trace is wrong.

Btw, I don't like that throw anyway, why don't we just return an error tuple?
I think @arjan had some ideas about this.

PR #1404, opened 2016-08-30
mod_acl_user_groups: make the is_owner check a notification.

### LINKED ISSUE #1403: Make acl_user_groups_checks:is_owner a notification
So client sites can have some custom logic as to when a user is a owner, for instance if the user has an author edge to the article in question.

mworrell:
Fixes #1403 

Also: if an user is member of a collab group, then the user can also see unpublished content.
The #acl_is_owner{} notification makes it possible to flexibly define the concept of content ownership.

### DIFF
--- a/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
+++ b/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
@@ -494,13 +494,16 @@ can_insert_with_ug(Cat, Context) ->
 %% @doc Check if the user can view/update/delete/link the resource
 can_rsc(undefined, _Action, _Context) ->
     false;
-can_rsc(Id, view, Context) when is_integer(Id) ->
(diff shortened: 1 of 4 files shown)


mworrell:
@ddeboer please check, and merge :)

modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
```diff
@@ -500,7 +506,10 @@ can_rsc(Id, view, Context) when is_integer(Id) ->
     UGs = user_groups(Context),
     (
         can_rsc_1(Id, view, CGId, CatId, UGs, Context)
-        andalso m_rsc:p_no_acl(Id, is_published_date, Context)
+        andalso (
+            is_collab_group_member(CGId, Context)
+            orelse m_rsc:p_no_acl(Id, is_published_date, Context)
+        )
```
ddeboer: But for non-collaboration group setups, wouldn’t this deny view permission on unpublished resources? What if you explicitly want members of (e.g.) the managers group to view unpublished resources?

PR #1419, opened 2016-09-01
core: Remove datamodel list fallback

ddeboer:
Only for 1.0.


mworrell:
Didn't even know that option still existed...

PR #1426, opened 2016-09-05
docker: Negate paths in .dockerignore to make image even smaller

ddeboer:
As discussed in https://github.com/zotonic/zotonic/pull/1374#discussion_r77437242.


kaos:
Neat idea. I like it, thanks :)

PR #1435, opened 2016-09-13
Add optional embedly authentication

mworrell:
### Description

Optionally add the embed.ly api key to the oembed request.=

### Checklist

- [x] documentation updated
- [x] add interface to set the Embedly key in the `Auth -> External Services` part

### DIFF
--- a/modules/mod_oembed/support/oembed_client.erl
+++ b/modules/mod_oembed/support/oembed_client.erl
@@ -113,10 +113,10 @@ code_change(_OldVsn, State, _Extra) ->
 
 do_discover(Url, State) ->
     UrlExtra = oembed_url_extra(State#state.context),
-    fixup_html(discover_per_provider(Url, UrlExtra, oembed_providers:list())).
+    fixup_html(discover_per_provider(Url, UrlExtra, oembed_providers:list(), State#state.context)).
 
 
-discover_per_provider(Url, UrlExtra, [Provider=#oembed_provider{}|Rest]) ->
+discover_per_provider(Url, UrlExtra, [Provider=#oembed_provider{}|Rest], Context) ->
     case re:run(Url, Provider#oembed_provider.url_re) of
         {match, _} ->
             case Provider#oembed_provider.callback of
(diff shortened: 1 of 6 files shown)


mworrell:
@ddeboer can you have a look and merge if you like?  I think all is ok, and tests do work.

src/models/tests/m_identity_tests.erl
```diff
@@ -42,7 +42,10 @@ check_password_no_user_test() ->
     ?assertEqual({error, nouser}, m_identity:check_username_pw("mr_z", "does-not-exist", C)),
     ok.
 
-check_username_password_test() ->
+check_username_password_test_() ->
+    {timeout, 20, fun() -> check_username_password() end}.
```
ddeboer: Was this necessary due to some timing issue?

mworrell: Yes, I guess the db can be a bit slow on Travis. The default timeout is 5 seconds, which gave random timeout failures.

modules/mod_oembed/support/oembed_client.erl
```diff
@@ -19,140 +19,78 @@
 -module(oembed_client).
 
 -author("Arjan Scherpenisse <arjan@scherpenisse.net>").
--behaviour(gen_server).
```
ddeboer: Gen_server asynchronicity was dropped. Will this impact performance in any way or where the `handle_casts` not useful anyway?

mworrell: It was only used to serialize requests, as only the call interface was used.
If you need something parallel or async then it is better to spawn yourself.

Besides that there was an issue where the gen_server was not restarted on failure, as it is not part of a supervisor hierarchy but just a linked process.

mworrell:
@ddeboer Thanks for checking and merging!

ddeboer:
Cool, thanks! This should also go to 0.x, right?

mworrell:
I think it is useful for Driebit, as the embed is used more often now.

PR #1439, opened 2016-09-14
Process postbacks async using sidejobs connected to the session.

mworrell:
### Description

With the introduction of Cowboy all web socket requests and replies are handled in a single process.  Data pushes are blocked whilst requests are handled.

This change introduces concurrent sidejobs for the handling of postback requests:

 * Max 10 running sidejobs per session
 * Max 10% of the process_limit  sidejobs per Zotonic instance
 * Max 50% of the process_limit  spawnjobs per Zotonic instance
 * `#msg_ack{}` is sent on receive of a request, so long running requests are not automatically retried
 * `#msg_ack{}` with a payload of `overload` is sent for back pressure if no side jobs can be started
 * Requests sent to the postback controller using a forced `ajax` transport are handled synchronously so that all cookies are correctly set (think logon code)
 * Session side jobs are linked to the session, if the session crashes then all the linked processes are stopped. Also, if the session is stopped, all linked processes are killed.

This breaks the assumption that all requests are handled in sequence.

### Checklist

Tests are quite hard, due to the fact that this code is called from the browser and interacts with the javascript in the browser. We need to define a way in which we can test this kind of web socket code without a browser.

I can't seem to find documentation about the various settings in `zotonic.config`, except in the file itself.


mworrell:
@mmzeeman @ddeboer please have a quick look to see if this is ok to merge.

modules/mod_base/controllers/controller_postback.erl
```diff
@@ -28,6 +28,7 @@
 
 -include_lib("zotonic.hrl").
 
+-define(SIDEJOB_TIMEOUT, 1000).
```
mworrell: This is how long we wait for sidejobs handling the request. Any late data could be sent using comet, this wait period **must** be adjusted to be shorter than the wait period of the comet connection.

mworrell: The comet flush is now set to: 

```
-define(COMET_FLUSH_DATA,  100).
```

So the `SIDEJOB_TIMEOUT` must be something like 90 msec.

mworrell: Changed the `SIDEJOB_TIMEOUT` to 90 msec.

mmzeeman: The comet connection is waiting a little bit before sending the first thing that arrives. Usually more arrives, and it is quite expensive to setup long-polling connections.

Is it possible to set longer timeouts for long tasks?

mmzeeman: Why must the sidejob timeout be lower than the comet timeout?

mworrell: Good question.  The sidejob places its result in the page_session queue, which then signals the comet process. In the mean time we also placed the message ack in the queue.  So if we wait longer than the comet-timeout, then the comet will transport the ack.  

So either we wait shortly and let the comet transport the result, or we wait longer and let the comet transport the ack.

mmzeeman: Aha... I see... but the ack just means that the message arrived, not that it was processed correctly.

mworrell: Yes, that is a difference between the previous method and this MR. Now the ack is only telling something about the transport, and basically says `ok` or `overload`.

For overload we also send back an ack for qos 0 messages. The client will not be able to resend the message but will back off for one second.

mmzeeman: Aha, ok.. so this is a timeout for deciding overload or ok? I thought it was a timeout to check for completion of a sidejob.

mworrell: No, it is not for overload, only for transporting the result of the sidejob within the same postback request.

mworrell:
@ddeboer ok to merge?

modules/mod_base/controllers/controller_connection_test.erl
```diff
@@ -58,13 +58,20 @@ provide_content(Context) ->
...
     Vars1 = [{pages, Combi} | Vars],
-
-    z_render:update(TargetId, #render{template = <<"_session_info.tpl">>, vars = Vars1}, Context).
+    z_render:update(TargetId, #render{template = <<"_session_info.tpl">>, vars = Vars1}, Context);
+
+%% Test code for transport, should be moved somewhere else.
+event(#z_msg_v1{msg_id=MsgId, data=Data}, Context) ->
+    case proplists:get_value(<<"cmd">>, Data) of
+        <<"sleep">> ->
+            ?DEBUG({MsgId, Context#context.session_id}),
```
ddeboer: Remove

ddeboer: @mworrell Almost ready :wink:

mworrell: Done

ddeboer:
Pretty cool!

PR #1442, opened 2016-09-21
docker: Fix duplicate zotonic.config file

ddeboer:
### Description

docker: Fix duplicate config files

* Because make was run before creating the config files, Zotonic generated default config files in   the default locations, that then became used  starting Zotonic.
* This PR adds a default zotonic.config that is copied before make. Bonus: less sed’ing necessary.

### Checklist

- [x] no BC breaks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,12 @@ FROM alpine:3.4
 ADD . /opt/zotonic
 WORKDIR /opt/zotonic
 
+COPY docker/zotonic.config /etc/zotonic/zotonic.config
+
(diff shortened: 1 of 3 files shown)


CyBeRoni:
This pre-loads a config file which I explicitly avoided originally because it requires maintaining a separate, duplicate, configuration file. Best to sed the zotonic.config.in instead, leaving the password to be modified by the entrypoint script.

ddeboer:
We can have a minimalistic, separate file without trouble, as Zotonic already has fallbacks for the config entries that we do not set.

PR #1450, opened 2016-09-23
Attempt to get rid of some dialyzer issues.

willemdj:
### Description

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/src/db/z_db.erl
+++ b/src/db/z_db.erl
@@ -88,6 +88,10 @@
 -compile([{parse_transform, lager_transform}]).
 
 -type table_name() :: atom() | string().
+-type parameters() :: list(tuple() | atom() | string()).
+-type sql() :: string() | iodata().
+-type props() :: proplists:proplist().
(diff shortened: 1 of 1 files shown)


willemdj:
@ddeboer, I did some further work, but the tests fail. Since I believe that I only changed type specs and removed some dead code, I suspect that the problem may be somewhere else. Perhaps it has been solved in the main branch. Can you redo what you did on Saturday (merge master back into this branch)? Then I will do a pull again and see how things go after that.

No doubt there is a way to accomplish this without your help, you can also try to teach me that, of course.

ddeboer:
Hey Willem, thanks for helping out!

You’re right: the test failure had nothing to do with the Dialyzer changes, and should now be solved on master. I merge your PR into https://github.com/zotonic/zotonic/pull/1370, so if you want to continue working on Dialyzer (please do!) then create a new branch from the `travis-dialyzer` branch.

Issue #1377, opened 2016-08-10
translation_scan is broken in master


ArthurClemens:
* `scan_file` expects binary extension while it receives a string. This is because z_module_indexer returns strings
* `insert` receives different input than it expects

alvaropag:
@ArthurClemens Did you fix it? If not I can give a look at it.

ArthurClemens:
Please go ahead.

alvaropag:
Ok!

ddeboer:
I merged zotonic/template_compiler#7. @alvaropag Do you want to continue prepping `mod_translation` for #1371?

alvaropag:
Yep, sorry, busy week, will send ASAP the pull request with the fixes.

ddeboer:
No problem, Álvaro! Please let us know if you get stuck or have any questions when fixing the translation_scan.

mworrell:
@alvaropag  I suspect the problem with the scan of `.erl` files is that we sometimes don't have the right include paths for the Erlang parser. This gives an error during the parsing of the source file.

There is a new parser in `syntax_tools` (Erlang 18+) that only parses and doesn't do any includes. We might want to give that one a try.

alvaropag:
Thanks @ddeboer ! 
@mworrell Nice, I didn't know about this new parser, I'll give a look at it this weekend. Which versions of erlang are we supporting?

mworrell:
In master we only support Erlang 18 and later.

alvaropag:
Ok, so it's safe to use it :D

ddeboer:
Just tried `translation_scan` and it’s (now) working fine. Can we close this issue or am I missing something?

PR #1459, opened 2016-10-01
core: Consolidate translation files

### LINKED ISSUE #858: Translations: easier translation by combining modules in one .po file
I propose to combine all .po files from the modules into a single .po file per language.

This file could be located in `~/priv/translations/xx.po` (where xx is the iso language code).

This would make it easier to translate the core Zotonic into different languages as there is an overlap between the different modules and managing a single .po file is easier than the current situation of a .po file per module.

Additionally: after generating a new .pot file we can automatically merge that .pot file into the language .po files.

Comments?

/cc @arjan @mmzeeman @kaos @michielklonhammer

ddeboer:
### Description

Fix #858.

* For core modules, consolidate their translation template into
  priv/translations/template/zotonic.pot, using msgcat.
* For core modules, consolidate their translations into
  priv/translations/{lang}.po.
* Remove duplicate translations.
* Trim translation labels from templates.
* Extract translations labels as binaries so fix labels containing Unicode characters from being skipped during .pot generation (ref #1377).
* Update Docker container to include gettext package, which is  required for msgcat.
* Update documentation.

### Checklist

- [x] documentation updated
- [ ] tests added

### Todo

- [x] remove old translation files
- [x] add to `mod_translation:generate` function
- [x] include `priv/translations/` in translation lookup files.
- [x] exclude core modules from active module translation generation


modules/mod_translation/support/translation_scan.erl
```diff
@@ -72,14 +72,17 @@ merge_args([{Lang,Text}|Rest], Args) ->
...
 %% @doc Parse the template or Erlang module. Extract all translation tags.
 scan_file(<<".tpl">>, File) ->
-    scan_file(".tpl", File); 
+    scan_file(".tpl", File);
 scan_file(".tpl", File) ->
     case template_compiler:translations(File) of
-        {ok, Translations} -> normalize_line_info(Translations);
+        {ok, Translations} ->
+            %% Trim translation strings because template compiler includes spaces
+            Trimmed = [{z_string:trim(Label), Trans, FileInfo} || {Label, Trans, FileInfo} <- Translations],
```
ddeboer: @mworrell This was needed because the message labels coming in from the template compiler have outside spaces, which resulted in duplicate messages. Should we update the compiler to do the trimming itself?

modules/mod_translation/mod_translation.erl
```diff
@@ -590,10 +591,27 @@ language_list_item(Code, IsEnabled) ->
...
+    translation_po:generate(translation_scan:scan(ActiveModules, Context), Context);
 generate(Host) when is_atom(Host) ->
-    translation_po:generate(translation_scan:scan(z_context:new(Host))).
-
+    generate(z_context:new(Host)).
+
+%% @doc Generate consolidated translation file for all core modules
+%%      (both active and inactive)
+-spec generate_core(#context{}) -> ok.
+generate_core(#context{} = Context) ->
```
ddeboer: Created a seperate `generate_core` function.

src/i18n/z_gettext_compile.erl
```diff
@@ -64,110 +64,62 @@ generate(Filename, Labels) ->
...
-split_string(Str, End) ->
-    split_string(Str, End, 1, []).
-
-split_string(Str, End, End, Acc) ->
-    {lists:reverse(Acc), Str};
-split_string([H|T], End, N, Acc) when N < End ->
-    split_string(T, End, N+1, [H|Acc]);
-split_string([], _End, _N, Acc) ->
-    {lists:reverse(Acc), []}.
-
```
ddeboer: Replaced all this with a binary version.

ddeboer:
The `lang.country.po` files are still in mod_l10n/translations. I suggest to keep them there.

ddeboer:
Do we want tests for this?

modules/mod_translation/support/translation_scan.erl
mworrell: I think we should trim at the compiler, will add it.

ddeboer: Updated template compiler dep and removed trimming from here.

mworrell:
Great work!

PR #1466, opened 2016-10-05
Added function to wait for site startup.

mworrell:
### Description

Fixes a problem with tests and site installation where the site was not started properly before using it.

Adds z_sites_manager:await_startup/1 and/or z_module_manager:await_upgrade/1

### Checklist

- [x] tests added
- [x] no BC breaks


ddeboer:
Awesome!

Issue #1469, opened 2016-10-06
Optimise database connection pool


ddeboer:
On production systems, we’re sometimes running into the maximum number of database connections limit. It seems Zotonic doesn’t properly handle this situation, even after the number of connections has again dropped below the limit. Only restarting the site(s) fixes this. In the logs, we’re getting ‘remaining slots are reserved’:

```
{function_clause,[{m_hierarchy,tree_to_menu,[{error,{{error,{error,fatal,<<"53300">>,<<"remaining connection slots are reserved for non-replication superuser connections">>,[]}},[{z_db,'-q/4-fun-2-',5,[{file,"src/db/z_db.erl"},{line,289}]},{timer,tc,2,[{file,"timer.erl"},{line,181}]},{z_db,with_connection,3,[{file,"src/db/z_db.erl"},{line,206}]},{m_hierarchy,'-tree/2-fun-0-',2,[{file,"src/models/m_hierarchy.erl"},{line,100}]},{depcache,memo,5,[{file,"src/depcache.erl"},{line,100}]},{m_hierarchy,menu,2,[{file,"src/models/m_hierarchy.erl"},{line,187}]},{acl_user_groups_rules,expand,2,[{file,"modules/mod_acl_user_groups/support/acl_user_groups_rules.erl"},{line,57}]},{acl_user_group_rebuilder,rebuild,3,[{file,"modules/mod_acl_user_groups/support/acl_user_group_rebuilder.erl"},{line,49}]}]}},[]],[{file,"src/models/m_hierarchy.erl"},{line,189}]},{acl_user_groups_rules,expand,2,[{file,"modules/mod_acl_user_groups/support/acl_user_groups_rules.erl"},{line,57}]},{acl_user_group_rebuilder,rebuild,3,[{file,"modules/mod_acl_user_groups/support/acl_user_group_rebuilder.erl"},{line,49}]}]}
```

https://github.com/zotonic/zotonic/pull/1455#issuecomment-251786407

@mworrell Please summarise here what you have in mind. :wink:

mworrell:
Currently we have a database pool per site. This pool can grow from 1 to 10 connections (and more during peaks). Unused connections are closed, though one connection is always left open.

These pools function independently from each other. Especially at system start all sites need a couple of database connections, this maxes out the amount of database connections allowed by the PostgreSQL database.

Proposal is to make a single pool, sharing database connections between sites.

The selection of a connection is done by:

 * Same user/database/schema
[...]

PR #1470, opened 2016-10-06
Close open db connections on a 'too may connections' error.

mworrell:
### Description

Issue #1469 

This is a workaround for the problem where a database connection can not be made due to many other sites having open database connections.

If a _too many connections_ error is detected, then we close (almost all) open database connections held by db workers that are not in use.
### Checklist
- [ ] tests added
- [x] no BC breaks


ddeboer:
:+1:

mworrell:
And we need to test it against a real database....

ddeboer:
We have one on Travis. Do you want a test for it?

mworrell:
I was first thinking of a local test, by reconfiguring a local PostgreSQL server with a low connection limit.

mworrell:
I did some extra testing with a couple of sites and only 15 allowed db connections.
It now works well (after restructuring the site startup and fixing some smaller issues with the db connection process)

mworrell:
@ddeboer @arjan can you test this with your own dev env?

mworrell:
@ddeboer ok to merge this (and cherry-pick into 0.x)?

mworrell:
Merged this into the 0.x https://github.com/zotonic/zotonic/commit/8a747af1c4af352d713ade6721cb933e86188578

PR #1477, opened 2016-10-12
Fix #1476 show add connection buttons if linkable

### LINKED ISSUE #1476: Making a resource only linkable hides link buttons
Configuring the ACL module in such a way that a resource is linkable but not editable, hides the actual buttons to link in the admin screen. See screenshot:

screen shot 2016-10-12 at 11 34 06

It seems the linking rights are handed out correctly since disconnecting the resources works as expected. 

I'll look into fixing the relevant templates.

Dirklectisch:
Fix for #1476 

Show 'add connection' button in admin if linkable.

please cherry pick to 0.x after merge

### DIFF
--- a/modules/mod_admin/templates/_admin_edit_content_page_connections_list.tpl
+++ b/modules/mod_admin/templates/_admin_edit_content_page_connections_list.tpl
@@ -35,7 +35,7 @@ Params:
 %}
 {% endwith %}
 
-{% if is_editable %}
+{% if m.acl.is_allowed.link[id] %}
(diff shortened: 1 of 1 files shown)


ddeboer:
So linking from a resource does not require edit permissions on that resource.

@mworrell Does that correspond to way we currently understand the ACLs?

Dirklectisch:
@ddeboer As I understand having link rights on a resources implies that you can create and delete edges that have that resources as a subject.

mworrell:
Yes, link rights is with our ACL module about the subject.

PR #1478, opened 2016-10-13
core: Remove deprecated sendfile dependency

ddeboer:
### Description

tuncer/sendfile is [deprecated](https://github.com/tuncer/sendfile) in favour of Erlang’s native `file:sendfile/2` so remove the dependency.

### Checklist

- [x] no BC breaks


ddeboer:
> when was this deprecated?

Not sure, but it’s deprecated now by the authors of the library themselves: see the link above.

> erlang's sendfile implementation doesn't handle EINPROGRESS and EALREADY errno's correctly

Is that still the case, even for R18++ (our supported versions for 1.0)?

If I’m not mistaken, sendfile is used only by Cowmachine, which itself (still) has a dependency un tuncer/sendfile. So it’s rather safe to remove it here.

mmzeeman:
Hmm, I still think we want to have the possibility to disable the usage of sendfile. I've had problems with it in production on Linux, and it worked incorrectly (served cached or 1/2 content) on virtualbox.

ddeboer:
> disable the usage of sendfile

But where is it *enabled* or used in Zotonic itself? I guess it used to be relied on by Webmachine.

mmzeeman:
So it must be a cowmachine setting indeed.

mmzeeman:
Aha, yes this is now a cowmachine setting I see.

ddeboer:
So we can get rid of it here and (later) decide what we want to do about the deprecated sendfile in Cowmachine: https://github.com/zotonic/cowmachine/issues/1.

mmzeeman:
We should probably add cowmachine setting to our erlang.config skeleton and disable sendfile. After looking at erlang's file:sendfile/2 yesterday I think it is still broken.

mmzeeman:
Hmm, the setting in cowmachine doesn't do anything. So everything can be removed indeed.

I think I have some homework to do to check if erlangs sendfile works reliably. In the past I ran into problems with this. The thing is that the problem surfaces when the number of connections on your server gets around 10k open connections. I think I will still see the crashes I used to see in the past with erlangs sendfile implementation.

mmzeeman:
For reference: https://github.com/klacke/yaws/issues/187

ddeboer:
@mmzeeman Do you agree with merging this and move the discussion to zotonic/cowmachine#1?

mmzeeman:
Yup... I have to check in practice if erlangs sendfile works. I will check this on one of our live servers. The setting can be changed at runtime.

ddeboer:
Nothing better than testing on live systems. 😝

mmzeeman:
Some things are very hard to find out on a test machine. In this case the sendfile issues started taking place when the server is under a load of about 10k concurrent connections or so. Something I can easily test at night. At daytime we usually have around 30k open connections.

PR #1493, opened 2016-10-28
mod_filestore: Add delete interval

### LINKED ISSUE #1488: mod_filestore - delete failures are repeated forever.
At Driebit we use the same files on dev/test/prod so the filestore is configured to not allow deletion. Deletions are marked in the database but continuously queued for actual deletion which fails. The log is getting very full with the same error messages.
Is it possible to add a config setting that prevents the delete queuing?

ddeboer:
### Description

Fix #1488.
- Helps in recovering accidentally deleted files.
- Make deletion optional: useful for read-only setups such as
  non-production environments. Fixes `[error] <0.3542.31>@mod_filestore:delete_ready:302 Could not delete remote file. Path <<"archive/2016/10/16/XXXXXXX.XXX">> (XXX) error {error,forbidden}`

screen shot 2016-10-28 at 09 16 36

This covers our use cases:
- We have one S3 bucket per site, which is shared between (dev, acc, prod) environments. This makes for easy database imports (for testing and debugging) without the need to shuffle files around.
- Non-production environments (dev, acc) should only read S3 files, never delete them, because they may still be referenced in production. 
- On the (AWS) S3-side, we can re-enable the delete permission on our buckets as long as non-production is configured not to _delete_ files.
- The delay interval helps in recovering accidentally deleted files.

### Checklist
- [x] documentation updated
- [x] no BC breaks

### DIFF
--- a/modules/mod_filestore/templates/admin_filestore.tpl
+++ b/modules/mod_filestore/templates/admin_filestore.tpl
@@ -56,8 +56,23 @@
                                 <div class="checkbox col-md-12">
                                     <label for="is_upload_enabled">
                                         <input type="checkbox" id="is_upload_enabled" name="is_upload_enabled" {% if m.config.mod_filestore.is_upload_enabled.value == "true" %}checked{% endif %} />
-                                        {_ Upload new media files to the cloud by default _}
+                                        {_ Upload new media files to the cloud file store _}
                                     </label>
+                                </div>
+                            </div>
+
+                            <div class="form-group">
+                                <div class="form-inline">
+                                    <label for="delete_interval" class="control-label">{_ Delete files from the cloud file store _}&nbsp;</label>
+                                    {% with m.config.mod_filestore.delete_interval.value|default:"0" as value %}
+                                         <select class="form-control input-sm" id="delete_interval" name="delete_interval">
+                                             <option value="0"{% if value == "0" %} selected{% endif %}>{_ Immediately _}</option>
+                                             <option value="1 week"{% if value == "1 week" %} selected{% endif %}>{_ After 1 week _}</option>
+                                             <option value="1 month"{% if value == "1 month" %} selected{% endif %}>{_ After 1 month _}</option>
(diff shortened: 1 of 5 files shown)


ddeboer:
@CyBeRoni @fredpook Have a look.

modules/mod_filestore/templates/admin_filestore.tpl
```diff
@@ -56,11 +56,26 @@
...
                                     </label>
                                 </div>
                             </div>
 
+                            <div class="form-group row">
+                                <div class="col-md-12">
+                                    <label for="delete_interval" class="control-label">{_ Delete files from the cloud file store _}</label>
+                                    {% with m.config.mod_filestore.delete_interval.value|default:"0" as value %}
+                                        <select class="form-control-inline" id="delete_interval" name="delete_interval">
+                                            <option value="false"{% if value == false %} selected{% endif %}>{_ Never _}</option>
```
ArthurClemens: I would put Never as last option, so you get a gradual increase from now to never.

mworrell:
What is the use case for *after ... months* ?  Recovery of SQL backups?

mworrell:
Note that we also delete files on the file system immediately.  If the *after ... months* is for backups then it might be better to have that as a setting for all file deletions, not just the once on remote file stores.

ddeboer:
Yeah: 

> The delay interval helps in recovering accidentally deleted files.

CyBeRoni:
Recovery of mistakenly deleted files.

CyBeRoni:
> If the after ... months is for backups then it might be better to have that as a setting for all file deletions, not just the once on remote file stores.

The reasoning for doing it on the S3 store is that it's harder to take backups of such a storage method, though I don't oppose doing the same for local storage.

mworrell:
The local storage is also a filestore, so maybe we should refactor a bit and use the same logic for all stores.

CyBeRoni:
It makes sense for them to be set differently though.

mworrell:
True, we can keep it for the filestore for now. Though a single approach to file handling would be nice,  which can be done by making the local files a filestore method (nice for later).

PR #1497, opened 2016-10-31
Zmedia upload dialog

fredpook:
In some use cases you may want the initial tab to be always "upload" in the connect media dialog. When the dialog is opened from TinyMCE this is not possible, the tab will be empty. 

I don't see why the initial tab for zmedia should be restricted so this PR will allow any tab to be set as initial.


ddeboer:
Conflicts with the current master, so please update.

ddeboer:
@mworrell Do you agree with the intention of this PR? If so, I’ll clean this up so we can merge it.

mworrell:
Yes, I agree.  I received a clean diff from @fredpook, so we can just merge that.

ddeboer:
Rebased and cleaned up.

Issue #1399, opened 2016-08-29
Update documentation for 1.0


ddeboer:
- [x] check chapters like directory structure that are not updated with the rebar3 changes.
- [x] update Upgrading section ([example with before and after headers](https://github.com/symfony/symfony/blob/master/UPGRADE-3.0.md))
- [x] Document model changes (in upgrading section)
- [x] document datamodel list type that was removed
- [x] document Let’s Encrypt (#1502)
- [x] document relation between `lib-src` and `lib`
- [x] document `Makefile` usage for `lib-src`
- [x] Consistent terminology for SSL/TLS/HTTPS
- [x] Cherry-pick [Custom controller cookbook](http://docs.zotonic.com/en/0.x/cookbook/custom-controller.html) from 0.x

mworrell:
We might want to add an example controller in the documentation, showing all possible callbacks.

ddeboer:
Good idea, I’m definitely in favour of examples. :wink:

Issue #1329, opened 2016-06-28
There should be a way to set HSTS headers


CyBeRoni:
It would be incredibly nice for TLS-enabled Zotonic websites to be able to set an HSTS (HTTP Strict-Transport-Security) headers.

The header should have a configurable max age and allow configuring other directives (such as includeSubDomains and preload) as required. If enabled, it should be included on every response served over HTTPS (taking into account possible TLS-offloading proxies.)

See also https://tools.ietf.org/html/rfc6797.

ddeboer:
Skimming over the RFC, I see only two directives: `includeSubDomains` and `max-age`: is the latter the one you mean by preload?

mworrell:
I am redoing `mod_ssl` a bit. My plan is to make it aware of any SSL terminator proxies in front of the site. In that case we can add these headers to mod_ssl, just like the *secure* flags for the cookies.

mworrell:
BTW I will need a way to see if the request originated via SSL or not. Is there any information added in haproxy (or other SSL terminators) to derive this?

CyBeRoni:
RFC7239 defines such a thing.

ddeboer:
Exactly: what used to be `X-Forwarded-Proto` has now become `Forwarded: proto=https`. So then the logic becomes simple: when mod_ssl is enabled, set the `Strict-Transport-Security` header either
- if the current requests is oer HTTPS
- or if `Forwarded: proto=https` is set.

mworrell:
Merge the changes in #2297 into master

PR #1502, opened 2016-11-03
SSL refactoring - self-signed, ca, letsencrypt

### LINKED ISSUE #705: mod_ssl can overwrite previously created private key
During an installation of a new server I accidentally overwritten the private key. Big oops..

It goes wrong if you do this.
1. Run `openssl req -out sitename.csr -new -newkey rsa:2048 -nodes -keyout sitename.key`
2. Send csr to your ca
3. Start mod_ssl  **this step silently overwrites the `sitename.key`**
4. Install certificate
5. Pull out hair, if you have any :-p

### LINKED ISSUE #1218: Become the first to implement letsencrypt by default
We're at that point now, where nginx is already working on a default https config with letsencrypt. Zotonic could be among the first web-serving systems offering TLS for each domain by default.

### LINKED ISSUE #1517: Move SSL to core
SSL will be the default in Zotonic 1.0. 

1. Remove current mod_ssl. 
2. Move SSL handling to the core.
    - For SSL termination in front of Zotonic (e.g., HAProxy), add a ‘proxy whitelist’ configuration parameter that contains a list of trusted proxies’ IP addresses. 
    - If a request comes in and its request IP matches the whitelist, process its `Forwarded` header to overwrite the request’s IP (to be the client’s IP) and protocol (`https`); see https://github.com/zotonic/zotonic/issues/1329#issuecomment-259076893.
    - Add HSTS support (see #1329).
2. Create two new modules that manage SSL *certificates* (ordered by priority):
    * mod_ssl_ca (‘certificate authority’, i.e. paid certs)
    * mod_ssl_letsencrypt (see #1218)
    * mod_ssl_self_signed already exists, but needs to be fixed (see #1502) 

As discussed with @mworrell today. @CyBeRoni Makes sense?

Is there any compelling reason to support the outdated `X-Forwarded` header?

mworrell:
### Description

Fix #705
Fix #1218
Fix #1517

This removes the mod_ssl_self_signed and moves it into the core.
http2 is disabled due to cipher issues with Firefox and Chrome.

Self signed certificates are stored in their own subdirectories per site and hostname.

We also generate  single DH key for the Zotonic instance, as the generation of a DH key is expensive.
It is stored in `priv/ssl/dh2048.pem`

Self signed certs are used if the notification `#ssl_options` doesn't return a set of SSL options (with certs).

The zotonic_status site has now mod_ssl enabled.
It should force https, if possible.

### Checklist

- [x] Check if a single self signed cert per host is good enough (esp Firefox)
- [x] Check DH key generation during start, might cause timeouts
- [x] Add sha2 signature to the self signed keys
- [x] Decide cert handling for disabled/crashed sites when zotonic_status is served -- for now just the self-signed certificate, this needs extra work for the common situation
- [x] Add logic for https handling of zotonic_status if ssl is terminated by (something like) *haproxy*
- [x] Add proxy_whitelist handling, move ip_whitelist handling to z_stdlib
- [x] Cache certs in sni fun lookups
- [x] Fix `SSL: hello: tls_handshake.erl:167:Fatal error: insufficient security` in Firefox/Chrome **update:** seems to be a problem with some OTP 18 version, no problems in OTP 19.1
- [x] Should we disable TLS 1.0 and 1.1 support? **update:** for now leave 1.0, 1.1 and 1.2 support
- [x] documentation updated
- [x] add extra docs regarding lets encrypt.
- [x] maybe add overview page and functions to test which SSL certs are available and used
- [x] add upgrade notes
- [ ] tests added -- might need test for reading info from certificate
- no BC breaks -- definitely BC breaks


mworrell:
For now the http2 support is not advertised.

TODO:

reorder cipher list? See: https://sparanoid.com/note/http2-and-ecdsa-cipher-suites/
ECDHE-RSA-AES128-GCM-SHA256 and ECDHE-ECDSA-AES128-GCM-SHA256 should be at the top.

Otherwise Chrome will give ERR_SPDY_INADEQUATE_TRANSPORT_SECURITY

There is a problem with Firefox, which *needs* a cipher suite not implemented by Erlang

https://github.com/tatsuhiro-t/lucid/blob/ce8654a75108c15cc786424b3faf1a8e945bfd53/README.rst#current-status

mworrell:
Discussion about `mod_ssl`.  Currently we have the configurations:

 *  `mod_ssl.is_ssl` defines if the default https or http for dispatch rules
 *  `mod_ssl.is_secure` defines if the default secure flag for the session and *remember me* cookies
 * `mod_ssl.is_permanent` defines if the redirect to http should be 302 or 301
 * `site.protocol` defines the default protocol is http or https (esp. in the absence of mod_ssl)

This is a bit confusing and feels like a bad split.  Especially for a world where ssl will be the default.

Maybe it is better to let the site configuration decide about the ssl defaults (is_ssl, is_secure, and is_permanent).  Then we can remove `site.protocol`, fold the ssl dispatch flags into the dispatcher, and the secure cookie settings into the cookie routines.

mod_ssl will then only be used for user-supplied ssl certificates (if any).

/cc @arjan @ddeboer @mmzeeman

mworrell:
See also https://github.com/zotonic/zotonic/issues/1329#issuecomment-259076893

mworrell:
We have now the global configuration:

 * `proxy_whitelist` for the valid ip range of the proxy (as seen from the server)
 * `ssl_only` force ssl on all connections

And the site configuration:

 * `site.ssl_only` to force ssl on all connections, forced to true if the global `ssl_only` flag is set
 * `site.secure_session_cookie` to force the session and auto-login cookies to be secure, forced to `true` if `ssl_only` is set

If the global configuration `ssl_port` is set to `none` then `ssl_only` and `secure_session_cookie` are disabled.

mworrell:
Current status (with SSL Labs)
screen shot 2016-12-02 at 00 46 33

ddeboer:
Awesome work, Marc!

This branch conflicts with the current master. Can you rebase it?

> Should we disable TLS 1.0 and 1.1 support?

Are there any downsides with just going for 1.2? If not, I prefer to do that.

ddeboer:
And maybe change the title of this PR to reflect all that is actually going on here. :wink:

mworrell:
I just checked with @mmzeeman regarding the TLS versions, at the moment we can probably leave it as-is (support 1.0, 1.1 and 1.2).

mworrell:
For the docs, I propose to:

 * Add a chapter on *Port configuration, SSL and certificates* in the Reference guide
 * Add references to that chapter from Installation docs and *global configuration* page
 * Add functional descriptions of the mod_ssl_ca and mod_ssl_letsencrypt, referenced from the various sections.

It is a too much information (and too important) for paragraph on the configuration or installation pages.

mworrell:
@ddeboer could you have a look at that error in the docs, somehow sphinx is not picking up a label.

ddeboer:
Fixed the doc build: the file wasn’t yet included in the TOC. By the way, please mind the hierarchy of headers as described [in our docs](http://docs.zotonic.com/en/latest/developer-guide/contributing.html#heading-styles).

mworrell:
Might need some extra docs for the `mod_ssl_letsencrypt`. For the rest this is ready for merge.

ddeboer:
Cool! I’ll have a look at the docs and do some light testing locally before merging this.

mworrell:
Added some extra notes on top

include/zotonic.hrl
```diff
@@ -137,6 +134,16 @@
...
 
+%% Used for fetching the site dispatch rules (see also )
+-record(site_dispatch_list, {
+            site,
+            hostname,
+            smtphost,
+            hostalias,
+            redirect,
+            dispatch_list
+        }).
```
ddeboer: Please add types.

ddeboer:
It seems this PR has grown to fix other tickets as well. Can you please add them to the description?

mworrell:
I propose to make one extra change: add under the **admin -> system** menu an overview page with all SSL modules and their certificates, including the fallback self-signed ticket.  Then an admin can easily see which certs are used and what their validity is.

Then we can merge the Letsencrypt certificate request form to that page. 

We can also add a re-generate button for the self-signed ticket, for when it is not valid anymore.

ddeboer:
Do you want to make these changes to this PR?

mworrell:
Yes, I want to make those here. As it connects to the different parts and might need some small changes.

include/zotonic.hrl
mworrell: I added the types (and some more).

mworrell: See https://github.com/zotonic/zotonic/pull/1502/files#diff-7dce909e3540def7c27133cd9bbb361cR140

mworrell:
Added the overview page. You can see it at *admin -> system -> SSL Certificates*

This needs to be documented in mod_ssl_letsencrypt (and mod_ssl_ca?).

doc/developer-guide/upgrading.rst
```diff
@@ -370,11 +370,11 @@ The following include files have been renamed:
 HTTPS support
 ^^^^^^^^^^^^^
 
-HTTPS support was moved from the core into a new module, :ref:`mod_ssl`.
+HTTPS support was moved from the core into a new module, *mod_ssl*.
```
ddeboer: Is this still accurate? I don’t see any mod_ssl. We moved this to the *core*, right?

doc/developer-guide/upgrading.rst
```diff
@@ -370,11 +370,11 @@ The following include files have been renamed:
...
 
 The global ``priv/config`` options ``ssl``, ``ssl_certfile``,
 ``ssl_keyfile`` and ``ssl_password`` do no longer have an effect. See
-:ref:`mod_ssl` on how to configure HTTPS support for Zotonic from 0.9
+*mod_ssl* on how to configure HTTPS support for Zotonic from 0.9
```
ddeboer: Also mentions mod_ssl

src/support/z_ssl_certs.erl
```diff
@@ -0,0 +1,329 @@
...
+    case filelib:is_file(Filename) of
+        true ->
+            ok;
+        false ->
+            ok = filelib:ensure_dir(Filename),
+            % Maybe use the -dsaparam, maybe we shouldn't https://www.openssl.org/news/secadv/20160128.txt
+            % Though without this the generation will take a long time indeed...
+            lager:info("Generating ~s bits DH key, this will take a long time (saving to ~p)",
+                       [?DHBITS, Filename]),
+            Command = "openssl dhparam -out " ++ z_utils:os_filename(Filename) ++ " " ++ ?DHBITS,
```
ddeboer: This is a new dependency on openssl, right? We need to add it to the docs and Docker images.

mworrell: We always had that dependency for mod_ssl, so maybe undocumented but not new.

doc/developer-guide/upgrading.rst
mworrell: And mod_ssl has now been removed again.  Maybe we should delete the older upgrade notes, this gets a bit confusing.

ddeboer: Ah, sorry, now I see: you just removed the doc link. Just leave this then. A changelog is supposed to be immutable. :wink: We can always create a separate `upgrading-1.0.rst`.

mworrell:
Documentation is now quit complete, this can be squashed and merged.

ddeboer:
Yes, testing this as we speak on my dev env. :wink:

modules/mod_ssl_letsencrypt/templates/_admin_config_ssl.mod_ssl_letsencrypt.tpl
```diff
@@ -0,0 +1,44 @@
...
+                <p><a href="https://letsencrypt.org" target="_blank">{_ Read more at the Let’s Encrypt website _} &raquo;</a></p>
+
+                {% live template="_ssl_letsencrypt_status.tpl" topic="~site/letsencrypt" %}
+
+
+                {% if m.acl.use.mod_admin_config %}
+                    {% if m.sysconfig.port /= 80 %}
+                        <p class="alert alert-danger">
+                            {_ The port for HTTP requests must be 80, it is now _} {{ m.sysconfig.port }}<br/>
+                            {_ You can change this by setting <tt>port</tt> in the zotonic configuration file located in <tt>~/.zotonic/</tt>. _}
```
ddeboer: The file can be located somewhere else (e.g. `/etc/zotonic`). Better: 

> in the `zotonic.config` configuration file.

modules/mod_ssl_letsencrypt/templates/_admin_config_ssl.mod_ssl_letsencrypt.tpl
```diff
@@ -0,0 +1,44 @@
...
+                        </p>
+                    {% endif %}
+
+                    {% if m.sysconfig.ssl_port /= 443 %}
+                        <p class="alert alert-danger">
+                            {_ The port for HTTPS requests must be 443, it is now _} {{ m.sysconfig.ssl_port }}<br/>
+                            {_ You can change this by setting <tt>ssl_port</tt> in the zotonic configuration file located in <tt>~/.zotonic/</tt>. _}
```
ddeboer: Idem.

doc/developer-guide/deployment/https.rst
```diff
@@ -3,9 +3,10 @@
 HTTPS support
 =============
 
-Zotonic has built-in support for https and certificate handling.
+Zotonic has built-in support for HTTPS and TLS (previously SSL) certificate
+handling.
 
-All sites share a single ssl port using Server Name Indication for 
+All sites share a single TLS port using Server Name Indication for
```
mworrell: I called it *ssl* because we named the options in zotonic.config also `ssl_...`

ddeboer: Good point. There’s still time to change this before 1.0 in #1399.

mworrell:
🎉

ddeboer:
Great job here! Awesome stuff! 👏 👌

doc/developer-guide/deployment/https.rst
jult: Big provider Squarespace (also) recently decided on maintaining the SSL nomenclature (instead of TLS): https://support.squarespace.com/hc/en-us/articles/205815898-Squarespace-and-SSL

mworrell: I agree that there is quite some name confusion between SSL and TLS. And I assume that it is confusing people, even developers.

Maybe we should use *https* where possible? At least that is something most people recognize.

ddeboer: Zotonic is an HTTP framework, so in terms of SSL/TLS it’s only about handling HTTP traffic through it. Therefore, let’s go with ‘HTTPS’ in the docs (e.g., ‘HTTPS port’ is fine with me). We do need to reference the term SSL (TLS) as well, as the modules now have `ssl` in the name.

Let’s go with ‘HTTPS’ in the docs, 
‘HTTPS’ makes sense, as Zotonic is only about HTTP over SSL/TLS. Our modules now have `ssl` in the name.

/cc @CyBeRoni Any input on

mworrell: I agree with using HTTPS where possible.

We can leave some *SSL* and *TLS* words in the text, so that people can search for them and find the correct pages.

CyBeRoni: SSL is deprecated in its entirety and should be replaced by TLS everywhere, really.

mworrell: I agree that technically we should use TLS.

The problem @jult mentions is that everybody, everywhere, is using SSL.  As in "Buy here your SSL Certificates".  See for example https://www.digicert.com/

jult: Wasn't really seeing it as a problem. Just remembered I heard a squarespace advert on some netcast how they offered LetsEncrypt SSL for everyone now. Anyway, you can write "TLS (formerly known as SSL)" or something of course.

PR #1505, opened 2016-11-09
Filestore schema

ddeboer:
### Description

Following #1493, make sure the deleted column is actually created.

### Checklist

- [x] no BC breaks


ddeboer:
Sorry, wrong branch.

PR #1514, opened 2016-11-13
admin: Fix access checks

### LINKED ISSUE #1001: Check access control checks for predicate, user-groups, and categories management
Hi!

When i add a use group members and allow them the use of mod_admin they can change or remove predicates. You get the menu item structure but if you go to categories you get the login screen again. Think this is a bug.

~ Dorien

ddeboer:
### Description

Fix #1001.

* In dispatch rules and admin menu items, only check module use permissions.
* On category and predicate admin screens, check for insert, update and delete permissions to decide which buttons to show.
* Add ACL checks where there weren’t any.

### Checklist

- [x] documentation updated


mworrell:
Merged into master at d80695d

PR #1515, opened 2016-11-14
Fix: change qlobber/pubzub order

alvaropag:
### Description
Change the include order of qlobber.js and pubzub.js to fix a "Reference error: Qlobber is not defined"

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

Call first the qlobber.js file and then pubzub.js to fix dependency exception.


ddeboer:
Cheers, Álvaro! 👍

PR #1519, opened 2016-11-15
Collection page block

mworrell:
### Description

In the admin, when embedding a collection as a page block then allow to edit the haspart edges in-situ.

Also:

 * Add a *catinclude* argument to the live scomp
 * Pass the undo_message_id along the connection edit templates.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Oops, wrong base branch

PR #1522, opened 2016-11-15
core: Remove deprecated log functions

### LINKED ISSUE #1516: Clean up logging
As discussed with @mworrell today:

- [x] Remove `?zWarning` etc. macros.
- [x] Remove `z:log` and move it and the `#zlog{}` notification to a separate module: `z_db_log` (or something similar).
- [x] Document that all logging from Zotonic sites should happen directly through lager (`lager:warning` and friends).
- [x] Document that logging to database is possible but needs a separate `z_db_log:log` call.
- [ ] Refactor e-mail logging to use `z_db_log`.

This builds on top of #1510, but where #1510 keeps BC (on 0.x) this change will drop BC to clean up things.

ddeboer:
### Description

Fix #1516.

### Checklist

- [x] documentation updated

### DIFF
--- a/include/zotonic_log.hrl
+++ b/include/zotonic_log.hrl
@@ -7,9 +7,9 @@
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
 %% You may obtain a copy of the License at
-%% 
+%%
 %%     http://www.apache.org/licenses/LICENSE-2.0
-%% 
+%%
 %% Unless required by applicable law or agreed to in writing, software
 %% distributed under the License is distributed on an "AS IS" BASIS,
 %% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


include/zotonic_log.hrl
```diff
@@ -57,25 +57,7 @@
...
-
 %% Easy to use macros for debugging/development
 -define(PRINT(Var), lager:info("DEBUG: ~p:~p - ~p: ~p~n", [?MODULE, ?LINE, ??Var, Var])).
 -define(STACKTRACE, erlang:display(try throw(a) of _ -> a catch _:_ -> erlang:get_stacktrace() end)).
--define(DEBUG(Msg), z:debug_msg(?MODULE, ?LINE, Msg)).
-
-%% Deprecated shortcuts
--define(LOG(Msg, Args), lager:info(Msg, Args)).
--define(ERROR(Msg, Args), lager:error("~p:~p "++Msg, [?MODULE, ?LINE|Args])).
+-define(DEBUG(Msg), lager:debug("DEBUG: ~p:~p: ~p~n", [?MODULE, ?LINE, Msg])).
```
ddeboer: Not sure if we need to keep `?DEBUG`, as `lager:debug("meh")` is almost just as easy as `?DEBUG("meh")`. @mworrell What do you think?

mmzeeman: Almost the same... The nice thing about the ?DEBUG macro is that it also returns "meh". This makes debugging easy as you can just slap a ?DEBUG around something.

mmzeeman: For example if you have something like this:

```erlang
case proplists:get_value(<<"foo">>, Attrs) of
    ...
```

You can slap a ?DEBUG around the entire construct easily.

```erlang
case ?DEBUG(proplists:get_value(<<"foo">>, Attrs)) of
    ...
```

ddeboer: Good point! I’ll check if I haven't inadvertently changed that behaviour here.

mworrell:
@ddeboer ready for merge?

ddeboer:
Yes, can now be merged. Fixed the `?DEBUG` thingie.

PR #1534, opened 2016-11-23
core: Fix template render error logging

ddeboer:
### Description

- Fix `lager:error` calls throwing a `FORMAT ERROR` due to invalid number of parameters.
- Remove site name from log messages (see also #1510).

### Checklist

- [x] no BC breaks

### DIFF
--- a/src/support/z_template.erl
+++ b/src/support/z_template.erl
@@ -104,31 +104,38 @@ render_block(OptBlock, Template, Vars, Context) when is_map(Vars) ->
     case Result of
         {ok, Output} ->
             Output;
-        {error, {{ErrFile,Line,Col}, _YeccModule, Error}} ->
+        {error, {{ErrFile, Line, Col}, _YeccModule, Error}} ->
             try
                 Error1 = iolist_to_binary(Error),
-                lager:error("[~p] Error rendering template: ~s:~p (~s)~n",
-                           [z_context:site(Context), ErrFile, Line, Col, Error1])
+                lager:error(
+                    "Error rendering template ~s:~p:~p due to ~s~n",
+                    [ErrFile, Line, Col, Error1]
+                )
             catch
                 _:_ ->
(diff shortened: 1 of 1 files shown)


src/support/z_template.erl
```diff
@@ -104,31 +104,17 @@ render_block(OptBlock, Template, Vars, Context) when is_map(Vars) ->
...
+                "Error rendering template ~s:~p:~p due to ~s ~n",
+                [ErrFile, Line, Col, Error]
+            ),
             <<>>;
         {error, Reason} when is_list(Reason); is_binary(Reason) ->
-            try
-                Reason1 = iolist_to_binary(Reason),
-                lager:error("[~p] Error rendering template: ~s (~s)~n",
-                           [z_context:site(Context), Template, Reason1])
-            catch
```
ddeboer: @mworrell I removed the `try … catch`. Why would it be necessary?

mworrell: It is needed because the Reason might not be an `iolist()`, in that case we show it as a `~p` in the catch

ddeboer: What if we replace `iolist_to_binary` with `z_convert:to_binary` (which works on other types as well)?

mworrell: The problem is that the template-compiler can throw reasons with stuff that isn't an iolist, think of atoms and tuples.

ddeboer: Right, and while `~s` will produce nicer formatting (for iolists), `~p` is necessary for non-iolists?

mworrell: I first had that guard, it was not enough ... so we need the `try/catch`.

ddeboer: Ok, will put it back!

ddeboer: Done.

mworrell:
Looks good to me, I also bumped into this issue when working on the SSL things.

Will merge.

PR #1538, opened 2016-11-28
Search weight config

fredpook:
Allow configuration of rank weight.
The default '{0.05, 0.25, 0.5, 1.0}' is a bit too fussy for some clients.
Searches with a rank configured like '{0.05, 0.10, 0.25, 1.0}' will have results with more emphasis on the title text.


ddeboer:
Thanks!

PR #1539, opened 2016-11-28
mod_editor_tinymce - translation for middle size changed to medium size

fredpook:
Due to the the use of {_ Middle _} in person's middle name field we often get that translation in the image size property. Anyway this small fix will be more inline with common usages of Small/Medium/Large for sizes.


ddeboer:
Good one, cheers!

PR #1543, opened 2016-11-29
mod_base: fix livevalidation message_after error when invalid id

### LINKED ISSUE #1542: livevalidation.js error when setting property message_after with invalid id
When using the message_after property of the validate tag with an invalid id the livevalidation.js logic tries to access a property of a null value. Because of this exception the form submit is redirected to $URL$#pb-installed and a blank page is showed.

alvaropag:
### Description
Add a verification to guarantee that this.insertAfterWhatNode is not null when an invalid id is used through the message_after parameter of the validate tag.

Fix #1542 

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Thanks, the element itself is indeed a good default. 👍

mworrell:
Also cherry-picked into the 0.x

PR #1548, opened 2016-11-30
doc: Document Logstash configuration

### LINKED ISSUE #992: logging: normalize log notifications and add support for Kibana
I need to add support for log (of various events/sources) storage in Elastic Search for analysis with Kibana.

For this I think that it is useful to normalize some log events:
1. A standard log event: `#zlog{type, properties)` 
2. Change internal logging to this format
3. Add some extra log events (like access log)

The events are handled by one or more log processors (aka sinks), depending on the enabled modules.

For the Elastic Search logging, the events will be buffered and written in batches to an index like `zotonic-log-<sitename>`   Every log event's _type_ will be mapped to a Elastic Search _type_.

We might want to use an ES index per period, this will be handled by some configuration.
The name of the ES index will be postfixed: `zotonic-log-<sitename>-201507` 
I propose a default period of a year.

Another configuration will decide which log _types_ are written to Elastic Search.

/cc @arjan @mmzeeman @fredpook

ddeboer:
### Description

Fix #992.

### Checklist

- [x] documentation updated
- [x] no BC breaks

### DIFF
--- a/doc/cookbook/index.rst
+++ b/doc/cookbook/index.rst
@@ -17,6 +17,7 @@ Cookbooks
    overriding
    task-queue
    template-icons
+   logstash
    error-pages
 
 Frontend cookbook
--- a/doc/cookbook/logstash.rst
+++ b/doc/cookbook/logstash.rst
@@ -1,0 +1,92 @@
+.. _cookbook-logstash:
+
+Logging to Logstash
+===================
+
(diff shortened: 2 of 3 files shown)


ddeboer:
@mworrell Do you agree that this closes #992?

ddeboer:
Please review and merge.

mworrell:
Looks perfect, thanks for writing this up!

PR #1550, opened 2016-12-02
Fix coding style

### LINKED ISSUE #1275: Improve coding standards
Our [current ones](http://zotonic.com/docs/latest/developer-guide/contributing.html#coding-standards) are very limited.

Questions:
1. order of `-spec` and `%% @doc` function header
2. indent for function bodies
3. explain how to enable [Editconfig plugins](http://editorconfig.org/#download) in editors ([Emacs](https://github.com/editorconfig/editorconfig-emacs), [Vim](https://github.com/editorconfig/editorconfig-vim)), as I see al lot of (e.g.) trailing whitespace in the code, which causes diffs to be unnecessarily large

Does anyone know of linting tools that can catch CS violations?

ddeboer:
### Description

Fix #1275.

This is going to be a huge, fat, ugly pull request, but what the heck – let’s do it anyway. :wink:

### Checklist

- [x] no BC breaks
- [x] Re-enable Elvis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,12 @@ matrix:
       env: BUILD_DOCS=true
     - otp_release: 18.3
     - otp_release: 18.0
-    # - otp_release: 19.0
+    - otp_release: 19.0
+      env: ELVIS=true
     #  env: DIALYZER=true
-  allow_failures:
-    # - otp_release: 19.0
-    #  env: DIALYZER=true
+#  allow_failures:
+#     - otp_release: 19.0
+#       env: DIALYZER=true
   fast_finish: true
 
 addons:
@@ -28,6 +29,7 @@ script:
   - if [[ "$BUILD_DOCS" = true ]]; then make -C doc/ stubs; fi
   - if [[ "$BUILD_DOCS" = true ]]; then make -C doc/ SPHINXOPTS='-nW' html; fi
   - ./rebar3 xref
+  - if [[ "$ELVIS" = true ]]; then wget https://github.com/inaka/elvis/releases/download/0.2.12/elvis && chmod +x elvis && ./elvis rock; fi
   - if [[ "$DIALYZER" = true ]]; then ./rebar3 dialyzer; fi
 sudo: false
 branches:
--- a/doc/developer-guide/contributing.rst
+++ b/doc/developer-guide/contributing.rst
@@ -38,14 +38,18 @@ how you can help improve Zotonic.
 Coding standards
 ----------------
 
-.. note::
-   As these conventions were established only after a large
-   part of the code base has been written, the code style described here
-   is not yet in effect in all parts of the Zotonic code base. We're
-   trying to gradually adapt to it, however.
+The Zotonic code follows `Inaka’s Erlang Coding Guidelines`_. This is enforced
+using the `Elvis`_ code checking tool. You can check the code locally with:
 
-When writing code, do not introduce trailing whitespace and try to keep lines
-of code shorter than 80 characters.
+.. code-block:: bash
+
+    $ elvis rock
+
+When contributing, simply try to follow the coding style as you find it in the
+existing code.
+
+Emacs
+^^^^^
 
 Provided with the Zotonic distribution is a Zotonic template mode,
 ``zotonic-tpl-mode``, which supports the Zotonic flavor of ErlyDtl.
@@ -59,15 +63,6 @@ file:
    (require 'zotonic-tpl-mode)
    ;; optional, for associating .tpl files with zotonic-tpl-mode
    (add-to-list 'auto-mode-alist '("\\.tpl$" . zotonic-tpl-mode))
-
-Indents
-^^^^^^^
-
-We use the "Emacs-style" indenting (using ``erlang-mode`` provided with the
-Erlang distribution). This indenting style seems to be a convention in much
-of the Erlang world.
-
-The indent is 4 spaces (no tabs).
 
 Writing commit messages
 -----------------------
@@ -350,3 +345,5 @@ hotfix, incrementing the patch number (for instance, ``0.16.1``).
 
 .. _Dutch test siren: http://www.invadingholland.com/guides-to-holland/emergency-alarm
 .. _Semantic versioning: http://semver.org/
+.. _Inaka’s Erlang Coding Guidelines: https://github.com/inaka/erlang_guidelines
+.. _Elvis: https://github.com/inaka/elvis
--- a/doc/developer-guide/docker.rst
+++ b/doc/developer-guide/docker.rst
@@ -47,7 +47,7 @@ on Zotonic. Start the container from your local Zotonic clone::
 
 To start the container, use Docker Compose::
 
-    $ docker-compose run --service-ports zotonic bash
+    $ docker-compose run --service-ports zotonic sh
 
 This opens a shell prompt in the Zotonic container. A PostgreSQL container will
 be started automatically as well. In the Zotonic container, you can enter any
--- a/elvis.config
+++ b/elvis.config
@@ -2,10 +2,29 @@
     {elvis, [
         {config, [
             #{
-                dirs => ["src", "include", "modules"],
-                filter => "*.erl",
+                %% dirs => ["src", "include", "modules"],
+                %% filter => ["*/*.erl"],
+                dirs => ["src/models", "src"],
+                filter => ["*.erl"],
                 rules => [
-                    {elvis_style, line_length, #{limit => 100}}
+                    {elvis_style, line_length, #{limit => 110}},
+                    {elvis_style, nesting_level, #{level => 6}},
+                    {elvis_style, used_ignored_variable, disable},
+                    {elvis_style, no_spec_with_records, disable},
+                    {elvis_style, dont_repeat_yourself, #{min_complexity => 50}},
+                    {elvis_style, function_naming_convention, #{regex => "^([a-z][a-z0-9]*_?[0-9]*)*$"}},
+                    %% for eunit tests: https://github.com/inaka/elvis/issues/403
+                    {elvis_style, variable_naming_convention, #{regex => "^(_[A-Z])|([A-Z][0-9a-zA-Z]*)$"}},
+                    {elvis_style, no_debug_call, #{ignore => [zotonic]}},
+                    {elvis_style, invalid_dynamic_call, #{ignore => [
+                        mod_survey,
+                        z_code_reloader,
+                        z_db,
+                        z_db_pool,
+                        z_filewatcher_sup,
+                        z_validation]
+                    }},
+                    {elvis_style, god_modules, #{limit => 50}}
                 ],
                 ruleset => erl_files
             }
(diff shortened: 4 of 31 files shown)


modules/mod_signal/mod_signal.erl
```diff
@@ -182,11 +195,11 @@ handle_call({'connect', Signal, Slot}, _From, State) ->
...
-                  true ->
-                      {slot, Key, Slot};
-                  false ->
-                      [{slot, Key, Slot}, {tags, SignalType, SignalTags}]
-              end,
+        true ->
+            {slot, Key, Slot};
+        false ->
+            [{slot, Key, Slot}, {tags, SignalType, SignalTags}]
+    end,
```
mworrell: Actually I find these *left-aligned* case statements uglier and much less readable then the normal *aligned below the case* versions.

ddeboer: I can imagine. The problem with aligned below is that (depending on the preceding statement) they can jump all over the place and start very far to the right end leaving only a little line length.

mworrell:
This will give quite some merge conflicts with my SSL branch.
Can you wait for that one?

Something else: maybe 132 columns?  The line printers we used in the 80s had already 132 columns... and it is still quite readable.

At least it is not 80 column punch cards.

modules/mod_signal/mod_signal.erl
mworrell: If it is starting to jump too far then we that case statement should be lifted out and made into its own function :)

ddeboer:
> Can you wait for that one?

Sure. Any idea as to what still needs to be done for the SSL PR? Just the docs?

> maybe 132 columns? The line printers we used in the 80s had already 132 columns... and it is still quite readable.

Let’s just follow common practice here as well. But Inaka specifies 80, which is (very) narrow, so I’m going with 110 for now.

mworrell:
Did Elvis do this automatically, there are quite some changes that are unexpected (to say it nicely :) )

mmzeeman:
BTW.

Erlang has automatic tools to tidy code automatically. This is an escript to use `erl_tidy` from the commandline.

```erlang
#!/usr/bin/env escript

main([Name]) ->
    erl_tidy:file(Name, config());
main(_) ->
    usage().

config() ->
    [{stdout, true},
     {keep_unused, true},
     %% erl_prettypr settings
     {paper, 132},
     {ribbon, 90}].

usage() ->
    io:put_chars("usage: erlfmt.escript <file>\n"),
    halt(1).
```

ddeboer:
Thanks for the tip, @mmzeeman. I’ve been playing around with automatic tools, but some have some unwanted consequences, as noted by @mworrell.

mmzeeman:
For example:

This `support/z_xml.erl`

```erlang
-module(z_xml).
-author("Marc Worrell <marc@worrell.nl").

%% interface functions
-export([
    escape/1
]).


%% @doc Escape a html text to valid a xml text so that it can be transported in XML.  Translates control characters to
%% spaces, except for TAB, CR and LF.
%% @spec escape(iolist()) -> iolist()
escape(undefined) ->
    undefined;
escape(<<>>) ->
    <<>>;
escape([]) ->
    [];
escape(L) when is_list(L) ->
    escape(list_to_binary(L));
escape(B) when is_binary(B) ->
    escape(B, <<>>).

    escape(<<>>, Acc) ->
        Acc;
    escape(<<$&, T/binary>>, Acc) ->
        escape(T, <<Acc/binary, "&#38;">>);
    escape(<<$<, T/binary>>, Acc) ->
        escape(T, <<Acc/binary, "&#60;">>);
    escape(<<$>, T/binary>>, Acc) ->
        escape(T, <<Acc/binary, "&#62;">>);
    escape(<<$", T/binary>>, Acc) ->
        escape(T, <<Acc/binary, "&#34;">>);
    escape(<<$', T/binary>>, Acc) ->
        escape(T, <<Acc/binary, "&#39;">>);
    escape(<<C, T/binary>>, Acc) when C == 9; C == 10; C == 13 ->
        escape(T, <<Acc/binary, C>>);
    escape(<<C, T/binary>>, Acc) when C < 32 ->
        escape(T, <<Acc/binary, 32>>);
    escape(<<C, T/binary>>, Acc) ->
        escape(T, <<Acc/binary, C>>).
```

Is turned into this:

```erlang
-module(z_xml).

-author("Marc Worrell <marc@worrell.nl").

%% interface functions
-export([escape/1]).

%% @doc Escape a html text to valid a xml text so that it can be transported in XML.  Translates control characters to
%% spaces, except for TAB, CR and LF.
%% @spec escape(iolist()) -> iolist()
escape(undefined) -> undefined;
escape(<<>>) -> <<>>;
escape([]) -> [];
escape(L) when is_list(L) -> escape(list_to_binary(L));
escape(B) when is_binary(B) -> escape(B, <<>>).

escape(<<>>, Acc) -> Acc;
escape(<<$&, T/binary>>, Acc) -> escape(T, <<Acc/binary, "&#38;">>);
escape(<<$<, T/binary>>, Acc) -> escape(T, <<Acc/binary, "&#60;">>);
escape(<<$>, T/binary>>, Acc) -> escape(T, <<Acc/binary, "&#62;">>);
escape(<<$", T/binary>>, Acc) -> escape(T, <<Acc/binary, "&#34;">>);
escape(<<$', T/binary>>, Acc) -> escape(T, <<Acc/binary, "&#39;">>);
escape(<<C, T/binary>>, Acc) when C == 9; C == 10; C == 13 ->
    escape(T, <<Acc/binary, C>>);
escape(<<C, T/binary>>, Acc) when C < 32 -> escape(T, <<Acc/binary, 32>>);
escape(<<C, T/binary>>, Acc) -> escape(T, <<Acc/binary, C>>).
```

ddeboer:
Even that is a bit too aggressive: what’s wrong with functions on two lines?

mworrell:
Functions on one line are quite common, then it is easier to see the pattern matching.

ddeboer:
Unfortunately, that’s where things become too subjective, in my opinion. That’s why I’m relying on some [external standards](https://github.com/zotonic/zotonic/issues/1275#issuecomment-220820881).

mworrell:
I think expressing intent and readability are *always* more important than some standard indentation.

mworrell:
The example `z_xml` module @mmzeeman showed above is very clear after the tidy. We should check that one in :-)

ddeboer:
> I think expressing intent and readability are always more important than some standard indentation.

True, but that’s no reason to not follow a standard. Guidelines such as Inaka’s are meant to improve readability by improving consistency. After all, readability in Zotonic’s code is (currently) hurt the most by the extremely inconsistent nature of the code (which I do understand, as the code has grown over 7 years). Inconsistency is a problem particularly for newcomers to the framework. 

I consider coming up with and agreeing on our own standards a waste of time: person A considers X more readable, person B prefers Y for clear intent etc.. Moreover, to gain momentum in the Erlang community, it makes sense to join some existing standards instead of coming up with our own.

> The example z_xml module @mmzeeman showed above is very clear after the tidy. We should check that one in :-)

If you like it, please do. But then also come up with a general rule (in this case, when to choose one-line over multi-line functions) and an automated way to test this to prevent us preferring our own subjective prejudices. Only then will we be able to have consistent code. 

For reference, this is what `erl_tidy` does to `m_predicate`:

```
-export([all/1, flush/1, for_subject/2, get/2, id_to_name/2, insert/2, is_predicate/2,
	 is_used/2, m_find_value/3, m_to_list/2, m_value/2, name_to_id/2, name_to_id_check/2,
	 object_category/2, objects/2, subject_category/2, subjects/2, update_noflush/4]).
```

Not so readable – in my humble opinion, course.

So,

1. Which rules in https://github.com/inaka/erlang_guidelines don’t we agree with?
2. Which rules should we add (e.g. one-liners)?
3. Which disagreements about the changes in this PR do we have? I think most of it is about indents (after `case` and `fun`).

mworrell:
The one-line rule is:

 * A function has multiple (similar) guards.
 * The function body is small

Most of the "rules" are actually common sense.

But I like to make clear structure.
So:

```erlang
       X = #somerecordname{id=Foo,
                           attr=Bar,
                           ...},
```
(had to play with the spaces here, GitHub doesn't to wysiwyg for the indentation)

and not:

```erlang
       X = #somerecordname{id=Foo,
                 attr=Bar,
                 ...},
```
(you can't easily see which fields are set)

but ok:

```erlang
       X = #somerecordname{
                 id=Foo,
                 attr=Bar,
                 ...},
```

Also:

```erlang
case some_function(
     arg, arg,
     arg)
of
     guard -> ...
     ...
end
```

and not:

```erlang
case some_function(
     arg, arg,
     arg) of
     guard -> ...
     ...
end
```
(you can't easily scan the guards)

With a `fun()` I also prefer to have the `end` at the same indentation as the `fun`.
In that way it is immediately clear what the fun is.

Same for `case` (or `if`) and `end`.

Those are conceptual blocks, and it helps to be able to see them as a single block, without searching in the first non-indented line to see what kind of block it is.

mworrell:
Having said all this, as long I can still read the code and it won't throw me off because of some funky structure (the case guards example above) then I am completely ok with the cleanup.

mmzeeman:
@ddeboer IMHO the code we have right now is consistent enough. If you want to change things, please go ahead. But if what we really want to do is attract new people there might be more effective ways to do that.

1) Have a clear slogan/mission of what we are trying to do with Zotonic. Zotonic is very different from the average run of the mill web framework/cms. It amazes me every time when other developers tell me what they have to do to get a site with 200k unique visitors running reliably. This is a no-brainer with zotonic. My guess is that if we explain what we are trying to do and how we do it, this will attract new people.

2) Make it easy for outsiders to join. Never turn away anyone who wants to contribute.

ddeboer:
> Most of the "rules" are actually common sense.

Yup, but still the Zotonic code doesn’t follow many of those, for instance the maximum nesting level – which definitely helps readability. 

I find the one-liner ‘rule’ a little too vague (but it doesn’t really matter, as this PR doesn’t change those), but I definitely agree with both records and cases. Do you see any examples in this PR where something else happened? That must’ve been unintentional. 

> But if what we really want to do is attract new people there might be more effective ways to do that.

Of course: this PR is not primarily geared towards that. Just to be clear here: the main goals of this PR are to:

1. Follow some simple conventions to improve code consistency and readability
2. Automate those checks (using Elvis) for new code changes so we don’t have to worry about it.

We had too many alerts when I previously enabled Elvis, so we have to make the code consistent first so we can re-enable the checks.

mworrell:
There were some places where the record fields and functions ended up un-aligned.
If most work is done then I can just walk through this and correct these by hand (hopefully not triggering Elvis errors :) )

mmzeeman:
I don't like the the Inaka rules which enforce style like the max level nesting and friends. This has nothing to do with consistent formatting.

Also, for the alignment, I use proportional fonts to edit code, so that would be very hard for me. Still waiting for elastic tab stops to become common.

ddeboer:
This fixes some CS issues and will have Elvis check src/*.erl and src/models/*.erl on Travis for any violations. The other directories are yet untouched.

PR #1555, opened 2016-12-05
docker: Install openssl to support TLS

ddeboer:
### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,3 +18,4 @@ services:
             - ./:/opt/zotonic
         ports:
             - 80:8000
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -18,3 +18,6 @@ services:
             - ./:/opt/zotonic
         ports:
             - 80:8000
+            - 8000:8000
+            - 443:8443
+            - 8443:8443
```
ddeboer: Remove some ports here.

PR #1563, opened 2016-12-22
Documentation update. Incorrect information

row-b:
The process_get/2 function expects 2 parameters (ReqData, Context) not only Context. The documentation misses the first parameter a couple of times

### Description

The documentation for API services mentions the process_get/2 function multiple times with a single parameter. 

* process_get(_Context) is replaced by (_ReqData, _Context) 
* process_get(Context) is replaced by (_ReqData, Context) 

### Checklist

- [ X] documentation updated
- [X ] no BC breaks

### DIFF
--- a/doc/developer-guide/services.rst
+++ b/doc/developer-guide/services.rst
@@ -122,7 +122,7 @@ of a services which handles a GET request is listed below::
 
     -include_lib("zotonic.hrl").
 
-    process_get(_Context) ->
+    process_get(_ReqData, _Context) ->
         Stats = [{count, 12310}, {uptime, 399}],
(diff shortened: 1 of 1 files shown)


mworrell:
I think this is because we have updated the docs already for the 1.0.
In the 1.0 we use cowmachine (instead of webmachine), which has a simpler interface for controllers.

Maybe we can make a note that this is a breaking change between the 0.x and 1.0.

ddeboer:
Please consult the [stable version of the docs](http://docs.zotonic.com/en/stable/developer-guide/services.html) for the 0.x versions. #1564 adds a link to the stable docs to make this clearer.

PR #1569, opened 2017-01-03
Minor typo in syntax fixed

row-b:
{{ok, Value, Context}; to {{ok, Value}, Context};

### Description

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


ddeboer:
Good catch, thanks!

PR #1576, opened 2017-01-11
core: Accept push queue names as binaries. Fixes #1575

### LINKED ISSUE #1575: z_transport push_queue encoding/decoding problem
The recent update of ubf encoding and decoding can cause problems.

screenshot from 2017-01-10 16 39 26

This is caused by the encoding and decoding of the push_queue field of acks and messages. They are sent as atom, but sometimes received as binary.

I want to change z_transport so that it accepts atom and binary push_queue parameters correctly. 

Ping @mworrell

mmzeeman:
### Description

Fix #1575

This fix makes zotonic accept push_queue names as binaries. Javascript doesn't really have atoms. Only undefined, true and false are transformed. 

### Checklist

- [ ] documentation updated (Not applicable)
- [ ] tests added  (Not done, we need end-to-end tests for this)
- [X] no BC breaks

### DIFF
--- a/src/support/z_transport.erl
+++ b/src/support/z_transport.erl
@@ -107,7 +107,7 @@ transport(L, Context) when is_list(L) ->
                   end,
                   L);
 transport(#z_msg_v1{} = Msg, Context) ->
-    case Msg#z_msg_v1.push_queue of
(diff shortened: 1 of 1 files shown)


mmzeeman:
This fix can be merged with 0.x too.

PR #1578, opened 2017-01-11
Add ACL checks on some properties

### LINKED ISSUE #1211: Private details of person records
We have ACL checks on property access.
The ACL property checks are not implemented by the current standard shipping modules.

With the enhancements to mod_export it is now easy to export email addresses of publicly viewable people.This is unwanted behavior.

With @ddeboer we came to the following idea, which needs some feedback and details filled in.
We make a new ACL module that adds some property checks for person records.
- Not expose `email`, `address_...`, `phone`, geo-locations and `date_start/end`.
- The address fields in `mailing_...` are exposed.
- An extra email fields `mailing_email` should be added (and also phone?)

Besides that we will need some settings in the record itself to toggle the privacy level of this information.

We need to think of indexing, normally we index everything. Not indexing the private properties will hide these from search results, this is both wanted and unwanted.

/cc @arjan @michielklonhammer @mmzeeman

mworrell:
### Description

Fix #1211 

Add ACL checks on some privacy sensitive properties.
The restrictions are regulated using the `privacy` property.

The `privacy` property can have the following values:

 - 0 public
 - 10 members
 - 20 member of same user group (except default members group)
 - 30 collaboration group members
 - 40 collaboration group managers
 - 50 private

The default for *person* is 30, for other categories 0.
We use increments of 10 so that more refined options can be added later.

The protected properties are:

- email
- phone
- phone_mobile
- phone_alt
- address_street_1
- address_street_2
- address_postcode
- address_city
- date_start
- date_end
- location_lat
- location_lng
- pivot_location_lat
- pivot_location_lng
- pivot_geocode
- pivot_geocode_qhash

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
+++ b/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
@@ -32,6 +32,7 @@
         has_collab_groups/1,
 
         acl_is_allowed/2,
+        acl_is_allowed_prop/3,
         acl_logon/2,
         acl_logoff/2,
         acl_context_authenticated/1,
@@ -45,9 +46,8 @@
         can_update_category/3,
         can_rsc_insert/3,
         can_move/3,
-        default_content_group/2
-
-    ,can_rsc/3
+        default_content_group/2,
+        can_rsc/3
     ]).
 
 -include_lib("zotonic.hrl").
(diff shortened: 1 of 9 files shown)


modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
```diff
@@ -212,6 +212,110 @@ acl_is_allowed(#acl_is_allowed{action=Action, object=ModuleName}, Context) when
...
+is_private_property(address_city) -> true;
+is_private_property(date_start) -> true;
+is_private_property(date_end) -> true;
+is_private_property(location_lat) -> true;
+is_private_property(location_lng) -> true;
+is_private_property(pivot_location_lat) -> true;
+is_private_property(pivot_location_lng) -> true;
+is_private_property(pivot_geocode) -> true;
+is_private_property(pivot_geocode_qhash) -> true;
+is_private_property(_) -> false.
```
ddeboer: What is the recommended course of action for changing this list, e.g. making a person’s name private as well? Should we have a separate `observe_prop_privacy` notification or just observe the higher-level `observe_acl_is_allowed_prop`?

modules/mod_acl_user_groups/templates/_admin_edit_content_acl.tpl
```diff
@@ -14,19 +14,42 @@
...
+    	    <a href="#" id="{{ #changecg }}" class="btn btn-default">{_ Change category and/or content group... _}</a>
+    		{% wire id=#changecg
+    				action={submit closest}
+    				action={dialog_open title=_"Category &amp; Content group" template="_action_dialog_change_category.tpl" id=id}
+    		%}
+    	{% endif %}
+    </div>
+
+    <div class="form-group">
+        <label class="control-label">{_ Show private details to _}</label>
```
ddeboer: I think we call ‘details’ ‘properties’ in most places.

modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
mworrell: I think for now a higher level observe_acl_is_allowed_prop. This could also map it to another property for checking by the `mod_acl_user_groups`.  I am a bit concerned about too much flexibility here as this is on a critical path in the property access methods.

modules/mod_acl_user_groups/templates/_admin_edit_content_acl.tpl
mworrell: Fair enough, changing it.

modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
mworrell: BTW if a person's name should be private, then I assume it is better to place the whole resource into a different content group.

This privacy setting should be restricted to some details of an otherwise viewable resource.

mworrell:
@ddeboer I propose to make the pivot-handling for private properties a separate issue. In that case we can also merge this into the 0.x (as pivoting in the master and 0.x are different due to the use of the pivoting templates)

mworrell:
Rebased and documentation added.

ddeboer:
I prefer to focus our efforts on a 1.0 release, so I’m okay with not having this in 0.x.

mworrell:
I agree, let's keep this 1.0 only.

PR #1591, opened 2017-01-24
mod_base: let filter without_embedded_media also consider text blocks.

### LINKED ISSUE #1566: mod_base - filter_without_embedded_media does not consider text blocks
Body text is scanned for z-media but text blocks are not. We get double images because they are not filtered.

mworrell:
### Description

Fix #1566

Add filter embedded_media.
Let filter without_embedded_media also consider the text blocks and `body_extra`.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/doc/ref/filters/filter_embedded_media.rst
+++ b/doc/ref/filters/filter_embedded_media.rst
@@ -1,0 +1,30 @@
+
+.. include:: meta-embedded_media.rst
+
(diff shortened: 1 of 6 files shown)


ddeboer:
Also cherry-picked to master.

PR #1598, opened 2017-02-01
core: Make pivot index name unique

ddeboer:
### Checklist

- [x] no BC breaks


mworrell:
Also merged into master.

PR #1606, opened 2017-02-11
core: Add facets property to search_result

### LINKED ISSUE #1447: Extend search_result with facets
Add a `facets` property to the `#search_result{}` record. While probably not used in the Zotonic core (facets are quite hard to do in Postgres) it will be very useful to return facets from `observe_search_query/2` observers that query [Elasticsearch](https://github.com/zotonic/zotonic/milestone/3), Solr and the like.

ddeboer:
### Description

Fix #1447.

### Checklist

- [x] no BC breaks

### DIFF
--- a/include/zotonic.hrl
+++ b/include/zotonic.hrl
@@ -139,8 +139,19 @@
 %% Default page length for search
 -define(SEARCH_PAGELEN, 20).
 
-%% Used for search results
--record(search_result, {result=[], page=1, pagelen, total, all, pages, next, prev}).
+%% @doc A set of search results
+-record(search_result, {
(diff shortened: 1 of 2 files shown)


ddeboer:
They are opaque indeed. The only thing this PR does is add a place for search modules to return the facets with the search results. The search module (or individual site) templates then decide how to process those facets to (e.g.) show a list of filters.

mworrell:
Ok with that for now, later we can establish a standard format.

ddeboer:
Alright, so can be merged?

mworrell:
Also merged into master at c734c6c

PR #1612, opened 2017-02-20
base: Remove empty pager CSS

ddeboer:
### Checklist

- [x] no BC breaks


mworrell:
Good one, css should be somewhere else.

PR #1617, opened 2017-02-27
Some nl translation fixes.

mworrell:
### Description

This fixes NL translations.

Some are taken from the 0.11/0.x and some are new.

We should also check the other languages.


ddeboer:
Still WIP?

mworrell:
Just merge it :-)

More are coming later.

PR #1627, opened 2017-03-02
Allow admin to override default config location also for rebar and such

CyBeRoni:
### Description

This PR expands on #1082 to also use the adminstrator-defined config file in places where rather than find_config_arg, find_config is used. Notably, this is done in rebar. The logic is identical to that earlier occurrence.

### DIFF
--- a/src/scripts/helpers/zotonic_setup
+++ b/src/scripts/helpers/zotonic_setup
@@ -62,9 +62,13 @@ function find_enabled_sites {
 export -f find_enabled_sites
 
 function find_config {
+    # Allow admin to override default behaviour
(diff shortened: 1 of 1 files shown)


CyBeRoni:
(Please cherry-pick to 0.x)

src/scripts/helpers/zotonic_setup
```diff
@@ -62,9 +62,13 @@ function find_enabled_sites {
 export -f find_enabled_sites
 
 function find_config {
+    # Allow admin to override default behaviour
+    [ ! -z "$ZOTONIC_CONFIG_DIR" ] && { echo "${ZOTONIC_CONFIG_DIR}/$1"; return 0; }
```
ddeboer: This check can now be removed from `find_config_arg` as that calls `find_config`.

CyBeRoni: Done.

PR #1643, opened 2017-03-22
core: Fix problems when updating dates. (#1642)

### LINKED ISSUE #1642: Not possible to add Date range
I tried to add an Event with a date range. The event gets added but the From and Till dates are not saved. Used Firefox or Chromium. Erlang 19 (erts 8.2). Zotonic 1.0-dev-62520f4

karlsson:
### Description

Fix #1642 

Fix problems when updating publication and event start and end dates

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/models/m_rsc_update.erl
+++ b/src/models/m_rsc_update.erl
@@ -1124,6 +1124,8 @@ to_int("") ->
     undefined;
 to_int(<<>>) ->
     undefined;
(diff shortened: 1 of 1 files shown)


mworrell:
Thanks for this quick fix!

PR #1651, opened 2017-04-05
mod_seo: Rename shorturl to shortlink

ddeboer:
### Description

mod_seo: Rename shorturl to shortlink  

shorturl is not accepted by https://validator.w3.org/. Probably shortlink (http://microformats.org/wiki/rel-shortlink) was meant instead of shorturl.

### Checklist

- [x] no BC breaks

### DIFF
--- a/modules/mod_seo/templates/_html_head.tpl
+++ b/modules/mod_seo/templates/_html_head.tpl
@@ -1,5 +1,5 @@
 {% if id %}
-	<link rel="shorturl" href="{% url id id=id %}" />
+	<link rel="shortlink" href="{% url id id=id %}" />
 	<link rel="canonical" href="{{ m.rsc[id].page_url }}" />
 {% endif %}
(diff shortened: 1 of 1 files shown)


mmzeeman:
Does this do anything useful? Why are there two links (canonical/short) anyway?

ddeboer:
Yeah, I was wondering about that too. It *can* be used by (mobile) clients to show a shorter version of the link, a bit like URL shortening services.

mworrell:
It used to be `shorturl`. That turned out to be a trademark by shorturl.com, so the w3c settled on `shortlink`

ddeboer:
As far as I can tell, `shortlink` is no W3C recommendation. Do we still agree on keeping it?

mworrell:
Ah, indeed, not w3c.  But keep it, as it is a valid microformat.

PR #1665, opened 2017-04-20
mod_acl_user_groups: Fix non-admins denied permission to insert file

ddeboer:
### Description

Fix https://github.com/driebit/ginger/issues/283.

#1640 introduced a bug: non-admins no longer have permission to insert files. This is because `Props` can contain `content_group_id` (with value `undefined`) when coming from `action_admin_dialog_media_upload`, so `proplists:lookup` is [not none](https://github.com/zotonic/zotonic/pull/1640/files#diff-2b13477af6ebbdd6a5dba710dbdbc463R203).

### Checklist

- [x] no BC breaks

### DIFF
--- a/modules/mod_admin/actions/action_admin_dialog_media_upload.erl
+++ b/modules/mod_admin/actions/action_admin_dialog_media_upload.erl
@@ -109,17 +109,21 @@ event(#submit{message={media_url, EventProps}}, Context) ->
 
 
 add_content_group(EventProps, Props, Context) ->
-    case proplists:get_value(subject_id, EventProps) of
+    case content_group_id(
+        proplists:get_value(content_group_id, EventProps),
+        proplists:get_value(subject_id, EventProps),
+        Context
+    ) of
         undefined ->
-            [ {content_group_id, proplists:get_value(content_group_id, EventProps)} | Props ];
-        SubjectId when is_integer(SubjectId) ->
-            ContentGroupdId = case proplists:get_value(content_group_id, EventProps) of
-                                    undefined -> m_rsc:p_no_acl(SubjectId, content_group_id, Context);
-                                    CGId -> CGId
-                              end,
-            [ {content_group_id, ContentGroupdId} | Props ]
+            Props;
+        ContentGroupId ->
+            [{content_group_id, ContentGroupId} | Props]
     end.
 
+content_group_id(undefined, SubjectId, Context) when is_integer(SubjectId) ->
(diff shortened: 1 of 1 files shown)


modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
```diff
@@ -637,7 +639,7 @@ is_owner(insert_rsc, _Context) ->
     true;
 is_owner(Id, #context{user_id=UserId} = Context) ->
     case z_notifier:first(#acl_is_owner{
-            id=Id, 
+            id=Id,
```
mworrell: Can you please change your editor to not modify whitespace? :)

mworrell:
Shouldn't the `undefined` content-group be mapped to the default content group?  Or is there some content group selection made later?

mmzeeman:
Until somebody makes a content group named undefined ;-)

mmzeeman:
Shouldn't https://github.com/zotonic/zotonic/blob/master/modules/mod_admin/actions/action_admin_dialog_media_upload.erl#L111 be adapted in such a way that it actually adds a content_group_id, instead of inserting undefined?

mworrell:
Agree, we should be able to upload a file to a specified content group (aka with ACL specific settings)

mworrell:
@mmzeeman the 'undefined' content group is during reads mapped to the default content group. This is used for situations where content was created without the `mod_content_groups` enabled.

mmzeeman:
Check, good to know. 

We recently started using mod_content_groups, without mod_acl_user_groups, only recently.

ddeboer:
So, to summarise, should `acl_user_groups_checks` or should it not support `{content_group_id, undefined}`? If it should not, I can change [action_admin_dialog_media_upload](https://github.com/zotonic/zotonic/blob/master/modules/mod_admin/actions/action_admin_dialog_media_upload.erl#L111 ) to remove the `content_group_id` key in case no (or the default) content group applies.

mworrell:
I think content group id `undefined` should map to the default content group, as that is also the behavior when reading resources (it maps `undefined` to default content group).

So for the fix it would be better to just not pass the content group, or use an explicit content group.

ddeboer:
Updated this PR: I replaced the changes in `acl_user_groups_checks` with a check around content_group_id in the media upload code.

mworrell:
Good, then we can later (in the master) check how to set the content group in the dialogs for the media uploads.  And what the handling of the 'undefined' value should be.

PR #1672, opened 2017-04-26
core: Read erlang.config in sitetest command

ddeboer:
### Description

The `zotonic sitetest` command didn’t read `erlang.config`. This PR adds `erlang.config` to the sitetest arguments. This solves configuration differences between live and test, which caused some of our tests to fail. 

@arjan Did you ignore `erlang.config` on purpose? It seems that we can add it without problems.

### Checklist

- [x] no BC breaks


ddeboer:
Ping @arjan.

PR #1674, opened 2017-04-30
core: fix a problem where compilation of templates generates random atoms

### LINKED ISSUE #1673: ErlyDTL compiler creates too many atoms
One of our servers just crashed with `no more index entries in atom_tab (max=1048576)`. As it turns out, the number of atom entries is very high on other servers as well. @mworrell reports nearly 1M entries on one of his servers.

The culprit seems to be [this line](https://github.com/zotonic/zotonic/blob/0.x/src/erlydtl/erlydtl_compiler.erl#L158) in erlydtl_compiler.

mworrell:
### Description

Fix #1673

This replaces dynamic atom generation with predictable enumerated atoms, such limiting the number of different atoms to be generated.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/erlydtl/erlydtl_compiler.erl
+++ b/src/erlydtl/erlydtl_compiler.erl
@@ -155,7 +155,7 @@ init_dtl_context(File, BaseFile, Module, Options, ZContext) when is_list(Module)
 init_dtl_context(File, BaseFile, Module, Options, ZContext) ->
     Ctx = #dtl_context{},
     #dtl_context{
-        local_scopes = [ [{'$autoid', erl_syntax:variable("AutoId_"++z_ids:identifier())}] ],
+        local_scopes = [ [{'$autoid', erl_syntax:variable("AutoId")}] ],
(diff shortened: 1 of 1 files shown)


mworrell:
@ddeboer please check this against your projects. It works well with the projects (and admin) I tested against.

mmzeeman:
Indeed, this is better, but this can only be a problem when templates are (re)compiled a lot. Is this because mod_development was running in production?

mworrell:
The 0.x is a bit conservative when it recompiles, like when modules are switched on and off, or when translations are loaded. 

The original assumption (in 2009) was that there were not a lot of templates, so that this leak would be small enough to disregard.

ddeboer:
We’re currently testing this branch on our test server. Will report here if we find any problems.

mmzeeman:
The branch worked for us... Didn't look at the atoms used before and after though.

Issue #295, opened 2012-02-20
Erlang and file descriptors (or: why I only got blank pages)


kaos:
The resolution to this issue lies in documentation or wiki or the like, I guess.

I post it here for now to bring it to attention.
If the core dev team already was very well aware of this, I feel it's time it was communicated to the rest of the zotonic user base.

I noticed I got a lot of completely empty responses from a zotonic instance (regardless of which site I was accessing).
Looking at the logs, I found pages upon pages of errors like this:

```
=ERROR REPORT==== 20-Feb-2012::20:59:26 ===
{mochiweb_socket_server,317,{acceptor_error,{error,accept_failed}}}

=ERROR REPORT==== 20-Feb-2012::20:59:26 ===
    application: mochiweb
    "Accept failed error"
    "{error,emfile}"

=ERROR REPORT==== 20-Feb-2012::20:59:27 ===
{mochiweb_socket_server,317,{acceptor_error,{error,accept_failed}}}

=INFO REPORT==== 20-Feb-2012::20:59:34 ===
Could not find template: article.tpl (enoent)
=ERROR REPORT==== 20-Feb-2012::20:59:34 ===
webmachine error: path="/en/article/684/chuck-moore-on-the-lost-art-of-keeping-it-simple"
{error,{throw,{error,{template_not_found,"article.tpl",enoent}},
              [{z_template,render,3},
               {resource_page,html,1},
               {resource_page,to_html,2},
               {webmachine_resource,resource_call,3},
               {webmachine_resource,do,3},
               {webmachine_decision_core,decision,3},
               {webmachine_decision_core,handle_request,2},
               {webmachine_mochiweb,loop,2}]}}
```

I know that `article.tpl` exists, so a little google to assist, turning up this nice blog post: http://metajack.im/2008/09/23/file-descriptors-are-yummy-or-common-pitfalls-of-ejabberd/

Looking at my ulimit, it says it is set to 1024, and checking how much beam has eaten up gives me slightly more...?!

```
[root@ganesha sasl]# lsof | grep beam | wc -l
1093
```

which is a considerable amount out of the total 2457 lines that lsof spits out.
I'm wondering whether there can be a fd leak...? Better keep an eye on it.

A quick reset of all beam processes trunc's it down to 114:

```
[root@ganesha sasl]# kill `pidof beam`
[root@ganesha sasl]# lsof | grep beam | wc -l
114
```

And my pages serve just fine again.

PR #1680, opened 2017-05-03
doc: Add server configuration chapter

ddeboer:
### Description

Ref #295.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/doc/developer-guide/deployment/server-configuration.rst
+++ b/doc/developer-guide/deployment/server-configuration.rst
@@ -1,0 +1,92 @@
+.. highlight:: bash
+
+.. _guide-deployment-server-configuration:
+
+Server configuration
+====================
+
+This chapter describes how to configure your Linux server for running Zotonic.
+
+.. _file-descriptors:
+
+File descriptors
+----------------
+
+By default, Linux limits the number of file descriptors (i.e. the maximum number
+of open files) at 1024. Running Zotonic at scale will require many more than
+that, particularly because Zotonic uses
+:ref:`WebSockets <controller-websocket>` extensively; remember that every open
+port is an open file.
+
+The limit applies on several levels:
+
+1. the process, in our case the Erlang VM (BEAM) that runs Zotonic
+2. the Zotonic user
+3. system-wide.
+
+To check the current process usage, find BEAM’s PID::
+
+    # pidof beam.smp
+    10006 (for instance)
+
+Then count its open files::
+
+    # lsof -a -p 10006 | wc -l
+
+And compare it to the process limit::
+
+    # cat /proc/10006/limits | grep 'Max open files'
+
+You can raise the process limit by adding ``ulimit -n`` to your Zotonic init
+script::
+
+    ulimit -n 50000
+
+Or change the limit in your system definition. For instance, when using
+systemd::
+
+    [Service]
+    LimitNOFILE=50000
+
+Finally, make sure to check `your system-wide limits`_, too.
+
+Network connections
+-------------------
+
+If you have stateful connection tracking enabled, high-traffic Zotonic sites
+may `overflow your conntrack table`_.
+
+Compare the current number of connections with the limit::
+
+    # sysctl net.netfilter.nf_conntrack_max
+    # sysctl net.netfilter.nf_conntrack_count
+
+If you need to raise the limit, edit ``/etc/sysctl.conf``:
+
+.. code-block:: none
+    :caption: /etc/sysctl.conf
+
+    net.netfilter.nf_conntrack_max = some_number
+
+Moreover, to reduce the number of open connections, you can decrease their
+time-out values so they get closed sooner. By default, inactive connections can
+stay in the conntrack table for 5(!) days. To change this:
+
+.. code-block:: none
+    :caption: /etc/sysctl.conf
+
+    net.netfilter.nf_conntrack_tcp_timeout_established = 600
+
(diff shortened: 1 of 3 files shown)


ddeboer:
@CyBeRoni Please review!

mmzeeman:
I used this as source of information http://antmeetspenguin.blogspot.nl/2011/01/high-performance-linux-router.html

CyBeRoni:
That look accurate to me. I think it's useful to add @mmzeeman's netfilter suggestion too, as that limit (while default higher than the 50000 you use in the doc) can be easily reached by high-volume websites on servers using stateful connection tracking (such as when using NAT or when using a stateful firewall). The way to set it on Linux would be to put

`net.nf_conntrack_max = some_number`

in `/etc/sysctl.conf`. Of course if a separate firewall is put in front of Zotonic (whether Linux-based or not) then the limit should be set there.

mmzeeman:
I also increased the hash size according to the rule of the thumb suggestions given in that article.

mmzeeman:
And lowered `net.netfilter.nf_conntrack_tcp_timeout_established` to 10 minutes. I didn't touch the other values.

mmzeeman:
FYI, that is on a server which does about 12M hits per day. The biggest bottleneck now is writing the access log to disk.

ddeboer:
Added paragraph on conntrack.

doc/developer-guide/deployment/server-configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+
+If you need to raise the limit, edit ``/etc/sysctl.conf``:
+
+.. code-block:: none
+    :caption: /etc/sysctl.conf
+
+    net.netfilter.nf_conntrack_max = some_number
+
+Moreover, to reduce the number of open connections, you can decrease their
+time-out values so get closed sooner. By default connection stay in the
```
CyBeRoni: connection*s*

doc/developer-guide/deployment/server-configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+conntrack table for 5(!) days. To change this:
+
+.. code-block:: none
+    :caption: /etc/sysctl.conf
+
+    net.netfilter.nf_conntrack_tcp_timeout_established = 600
+
+If you have a proxy in front of Zotonic (e.g. HAProxy or Varnish), you need to
+change the limits on the proxy instead of the Zotonic server.
```
CyBeRoni: You might need to do both, since a large part of the connections are comet/websocket which would be passed through to zotonic directly.

doc/developer-guide/deployment/server-configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+If you need to raise the limit, edit ``/etc/sysctl.conf``:
+
+.. code-block:: none
+    :caption: /etc/sysctl.conf
+
+    net.netfilter.nf_conntrack_max = some_number
+
+Moreover, to reduce the number of open connections, you can decrease their
+time-out values so get closed sooner. By default connection stay in the
+conntrack table for 5(!) days. To change this:
```
CyBeRoni: Note this only applies to open connections that see no activity. So for example a websocket connection to an idle client would be closed if it stayed entirely idle (including keepalives sent by the OS) for this amount of time. So a better line would be "can stay in the connection tabled for up to 5 days".

ddeboer:
Updated according to review.

doc/developer-guide/deployment/server-configuration.rst
mmzeeman: Thanks for the clarification. We sent ping/pong messages over the websocket to check if it is alive...

mmzeeman:
I'll add something about increasing the hash later today too. More connections also means that you have to increase the hash size I think.

@CyBeRoni is this the most handy method of connecting a high volume site to the internet. You mention something about stateful connection tracking. Is there a "raw" option which does not track connection and sends the raw packets to the port 8000 & 8443 only?

CyBeRoni:
I presume you mean getting it on port 80/443 without starting Zotonic as root. There are a number of ways to do that:

 - Use something like [authbind](https://en.wikipedia.org/wiki/Authbind) which uses a setuid root helper to bind the port for you
 - Use another program that binds to ports 80/443 and forwards traffic to Zotonic (like Varnish or HAProxy)
 - Use IPTables REDIRECT 
 - Use a new kernel with the ability to set ambient capabilities, add cap_net_bind_service and let zotonic bind to port 80/443 itself (all the time running as an unprivileged user)

Unfortunately most of these options (except the second) are Linux-only, though other platforms may have similar options available. (Example: [Solaris](https://blogs.oracle.com/tls/entry/solaris_10_non_root_user)).

> Is there a "raw" option which does not track connection and sends the raw packets to port 8000 & 8443 only?

There is no way that I know of to do NPT in Linux without using connection tracking. If you get Zotonic to listen on port 80/443, however, it is possible to avoid it by either not using any stateful firewall features (i.e., not using *any* iptables command involving -m state or -t nat, including -L) or by explicitly telling iptables to not track those packets like so:

```
iptables -t raw -A PREROUTING -p tcp --dport 80 -j NOTRACK
iptables -t raw -A OUTPUT -p tcp --sport 80 -j NOTRACK
```

I use the HAProxy option myself, as this also allows me to route traffic to various destinations by url.

PR #1683, opened 2017-05-04
Compatibility with older ACL modules.

### LINKED ISSUE #1633: mod_acl_simple_roles and mod_acl_adminonly no longer work in master
The old acl modules no longer work in master because macro's are missing and some observers are changed.

Wasn't it the idea to move these modules out of the core but still have them functional for backwards compatibility?

mworrell:
### Description

Fix #1633

Core changes to support older ACL modules again.

Changes to those modules are also needed, and being worked on.

The `mod_acl_adminonly` module is now compatible with this branch.

The `mod_acl_simple_roles` is a bit more complicated as it is also used for 0.x sites.

Maybe it is an idea (going forward) to make a function that maps the simple-roles model to a user-groups model?  Depending on the visible_for flag we can move content to different content groups.

### Checklist

- [x] documentation updated (see the mod_acl_adminonly module)
- [x] no BC breaks

### DIFF
--- a/include/zotonic.hrl
+++ b/include/zotonic.hrl
@@ -213,10 +213,12 @@
 %% ACL fields for an acl check. Fields are initialized for the visible resource.
 %% This is used for fetching the acl fields from a resource record.
 -record(acl_props, {
-    is_published=true,
-    is_authoritative=true,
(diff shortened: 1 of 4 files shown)


mworrell:
@ddeboer 😆

PR #1689, opened 2017-05-09
mod_base: Support maps result in moreresults filter

ddeboer:
### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


ddeboer:
Not sure what’s the use case of the `id` template var, thus why `to_id` is necessary at all, but this is the smallest non-breaking change to support maps.

mworrell:
You need the id for the optional catinclude.

mworrell:
Also merging into master (with small changes, as master is OTP 18 or newer).

PR #1693, opened 2017-05-16
core: Delay notifications during database transactions

### LINKED ISSUE #497: core: deadlock in module initialization
When the mod_survey is initialized from the admin then there is a timeout error.

This timeout error occurs because the mod_search is listening to the #rsc_update_done{} notification. This is done in its own process, so handled asynchronously.

Problem is that the initialization, and from that the #rsc_update_done{} notification, is done inside a transaction. All changes to the database are not yet committed.

mod_search is requesting the path of the non-committed "survey" category. That starts a memo entry in the depcache. The resulting category query is then waiting for the transaction that is inserting category.

When mod_survey then inserts the next category ("poll"), which is a sub-category of the previous "survey" category, it also needs the path of the the "survey" category. Which it can't get because the depcache then queues this request for the result of the memo call from the mod_search.

After the depcache gave a timeout this all ends.

How shall we solve this?

I am thinking of:
- Adding a mechanism to handle all #rsc_update_done{} notifications after the transaction is committed
- Change mod_search to do some work inside the calling process (using a pid_observe_rsc_update_done/2 function) before dispatching it to the mod_search process.

What do you think?

cc @arjan @kaos @mmzeeman

ddeboer:
### Description

Fix #497. 

Among other issues, this fixes the following error during manage_schema (and because of this, random z_sitetest failures):

```{'EXIT',{{case_clause,{rollback,{{badmatch,notfound},[{m_category,move_below,3,[{file,"/opt/zotonic/src/models/m_category.erl"},{line,704}]},{z_datamodel,'-manage/4-lc$^0/1-0-',4,[{file,"/opt/zotonic/src/support/z_datamodel.erl"},{line,65}]},{z_datamodel,manage,4,[{file,"/opt/zotonic/src/support/z_datamodel.erl"},{line,65}]},…,{z_db,'-transaction1/2-fun-0-',3,[{file,"/opt/zotonic/src/db/z_db.erl"},{line,135}]},{timer,tc,2,[{file,"timer.erl"},{line,181}]},{z_db,with_connection,3,[{file,"/opt/zotonic/src/db/z_db.erl"},{line,200}]},{z_db,transaction,3,[{file,"/opt/zotonic/src/db/z_db.erl"},{line,95}]}]}}},[{z_module_manager,manage_schema,4,[{file,"src/support/z_module_manager.erl"},{line,871}]},{z_module_manager,'-start_child/6-fun-2-',6,[{file,"src/support/z_module_manager.erl"},{line,704}]}]}}```

Thanks @mworrell for the help!

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/db/z_db.erl
+++ b/src/db/z_db.erl
@@ -123,7 +123,7 @@ transaction(Function, Options, Context) ->
 transaction1(Function, #context{dbc=undefined} = Context) ->
     case has_connection(Context) of
         true ->
-            with_connection(
+            case with_connection(
               fun(C) ->
                     Context1 = Context#context{dbc=C},
                     DbDriver = z_context:db_driver(Context),
@@ -139,7 +139,9 @@ transaction1(Function, #context{dbc=undefined} = Context) ->
                             R ->
                                 case DbDriver:squery(C, "COMMIT", ?TIMEOUT) of
                                     {ok, [], []} -> ok;
-                                    {error, _} = ErrorCommit -> throw(ErrorCommit)
+                                    {error, _} = ErrorCommit ->
+                                        z_notifier:notify_queue_flush(Context),
+                                        throw(ErrorCommit)
                                 end,
                                R
                         end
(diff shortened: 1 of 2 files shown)


mworrell:
Also merged into master.

PR #1697, opened 2017-05-17
mod_editor_tinymce: Don't override language if already set

ddeboer:
### Description

Ref. https://github.com/driebit/ginger/pull/294.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Also merged into master

PR #1699, opened 2017-05-18
core: Untangle name_to_id and rid

### LINKED ISSUE #440: name_to_id does not work for resources which have names with all numbers
m_rsc:name_to_id simply returns the passed number as the resource id. 

All uses of name_to_id should be checked, because some of those uses could expect this behaviour. If this is the case, m_rsc:rid should be used instead.

ddeboer:
### Description

- Make m_rsc:name_to_id/2 stricter by removing implicit conversion from
  textual name to resource id integer.
- Remove digital name conversion from m_rsc:name/2:
- Remove name_to_id_check/2 from m_category, m_predicate and m_rsc to
  reduce confusion.

Fix #440.

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/src/models/m_edge.erl
+++ b/src/models/m_edge.erl
@@ -194,13 +194,13 @@ insert(SubjectId, PredId, ObjectId, Opts, Context)
     end;
 insert(SubjectId, Pred, ObjectId, Opts, Context)
     when is_integer(SubjectId), is_integer(ObjectId) ->
-    PredId = m_predicate:name_to_id_check(Pred, Context),
+    {ok, PredId} = m_predicate:name_to_id(Pred, Context),
     insert1(SubjectId, PredId, ObjectId, Opts, Context);
 insert(SubjectId, Pred, Object, Opts, Context) when is_integer(SubjectId) ->
-    ObjectId = m_rsc:name_to_id_check(Object, Context),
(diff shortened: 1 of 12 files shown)


mworrell:
Looks good to me.

Less is more!

src/models/m_edge.erl
```diff
@@ -809,13 +816,13 @@ update_sequence(Id, Pred, ObjectIds, Context) ->
...
         true ->
-            PredId = m_predicate:name_to_id_check(Pred, Context),
+            {ok, PredId} = m_predicate:name_to_id(Pred, Context),
             F = fun(Ctx) ->
                 All = z_db:q("
                             select object_id, id
                             from edge
                             where predicate_id = $1
-                              and subject_id = $2", [PredId, m_rsc:name_to_id_check(Id, Context)], Ctx),
+                              and subject_id = $2", [PredId, m_rsc:rid(Id, Context)], Ctx),
```
mworrell: This will change the crash from Erlang to a database constraint error.  Which is ok with me.

ddeboer: I think the Erlang crash is clearer than some database constraint error, so fixed this.

PR #1700, opened 2017-05-18
core: Return error tuple instead of throwing exception

### LINKED ISSUE #1641: Remove exceptions
In the resource routines some exceptions are thrown.

There are a couple of problems with that:

 1. This is against the principle of non-local returns in Erlang
 2. Code becomes harder to reason with
 3. The *throw* looks different after a `z_db:transaction/2` call, as it is rethrown but with the original stack trace tagged on

I propose to replace most throws with simple `{error, …}` return values.

ddeboer:
### Description

Fix #1641.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


ddeboer:
Now green. Please check whether I forgot any throws.

mworrell:
Looks good to me!

The remaining throws in `z_db:q` (and friends) we can remove in a separate issue.

ddeboer:
See #1703 for `z_db`.

ddeboer:
Added a section in the Upgrading doc chapter.

PR #1702, opened 2017-05-19
mod_base: Support ISO 8601 dates in filter_date

ddeboer:
### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Should we also accept BC dates (like `"-9000-01-01"`)?

mworrell:
Also merged into master

PR #1707, opened 2017-05-22
core: Override ebin output directory (fix #1706)

### LINKED ISSUE #1706: Compiled user files in two places
When `user_ebin_dir` is configured (on master), and some file in the user dir is compiled, e.g. `zotonic/user/sites/yoursite.erl`, the compiled version ends up in two places, with one of course hiding the other:

- `zotonic/user/ebin/yoursite.beam`
- `zotonic/_build/default/lib/zotonic/ebin/yoursite.beam` (the Rebar dir).

ddeboer:
### Description

Fix #1706.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/elvis.config
+++ b/elvis.config
@@ -22,7 +22,8 @@
                         z_db,
                         z_db_pool,
                         z_filewatcher_sup,
-                        z_validation]
+                        z_validation,
+                        zotonic_compile]
                     }},
                     {elvis_style, god_modules, #{limit => 50}}
                 ],
--- a/src/filewatcher/z_filewatcher_handler.erl
+++ b/src/filewatcher/z_filewatcher_handler.erl
@@ -126,9 +126,8 @@ handle_file(_Verb, _Basename, ".hrl", F) ->
           end),
     "Rebuilding due to change header file: " ++ filename:basename(F);
 
-handle_file(_Verb, "erlydtl_parser.yrl", ".yrl", F) ->
-    TargetDir = filename:join(os:getenv("ZOTONIC"), "src"),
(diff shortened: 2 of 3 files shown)


ddeboer:
While this works well when calling `make` on the CLI, this doesn’t work when recompiling a running Zotonic: that still creates *both* .beam files. @mworrell Will your change fix that, too?

mworrell:
@ddeboer what do you mean with *both* beam files?

The fix I pushed will try to recompile with the same options and the same out-dir as the original compilation.

mworrell:
Am a bit unsure what to think of this Elvis "error"

```
# src/zotonic_compile.erl [FAIL]
  - invalid_dynamic_call
    - Remove the dynamic function call on line 150. Only modules that define callbacks should make dynamic calls.
```

The offending function:

```Erlang
previous_compile_options(Filename) ->
    Module = z_convert:to_atom(filename:rootname(filename:basename(Filename))),
    try
        MInfo = Module:module_info(compile),
        proplists:get_value(options, MInfo)
    catch
        error:undef -> undefined
    end.
```

https://travis-ci.org/zotonic/zotonic/jobs/234775310#L1460

mworrell:
I added `zotonic_compile` to the Elvis `invalid_dynamic_call` exception list.

ddeboer:
> @ddeboer what do you mean with both beam files?

The two beam files in [both locations](https://github.com/zotonic/zotonic/issues/1706#issue-230301336).

mmzeeman:
Works for me... except that the user ebin dir is not created automatically when it does not exist. But that is probably another issue.

PR #1711, opened 2017-05-28
doc: Move all doc build dependencies to one place 

ddeboer:
Only use Python packages from requirements.txt, not the Debian (outdated) ones.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,6 @@ matrix:
       env: DIALYZER=true
   fast_finish: true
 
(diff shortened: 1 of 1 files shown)


ddeboer:
While not strictly necessary (for one or another reason the build is green again on master), I’m still merging this as it puts our doc dependencies in one place (Python’s `requirements.txt`).

PR #1712, opened 2017-05-28
mod_seo: Remove meta title and keywords

ddeboer:
### Description

* Remove seo_title, which wasn't used at all.
* Remove seo_keywords, which was used for meta keywords. Google simply ignores those (https://support.google.com/webmasters/answer/79812 and https://webmasters.googleblog.com/2009/09/google-does-not-use-keywords-meta-tag.html), so it makes no sense to bother editors with them.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Only for master?

PR #1713, opened 2017-05-28
slice filter: convert binary

onnet:
Hi!

while trying to slice binary variable (call["caller_id_number"]|slice:[-1,]), got an error: Stack: [{lists,nthtail,[-1,[<<"user_ec34e4">>]],[{file,"lists.erl"},{line,180}]} 

It appears that template_compiler_runtime:to_list/2 just wraps binary var into brackets instead of converting it.
Provided patch solved the problem, do you think there could be any drawbacks?

Regards,
Kirill

### DIFF
--- a/modules/mod_base/filters/filter_slice.erl
+++ b/modules/mod_base/filters/filter_slice.erl
@@ -50,6 +50,8 @@ slice(undefined, _, _Context) ->
 
 slice(List, Slice, _Context) when is_list(List) ->
     slice1(List, Slice);
+slice(Binary, Slice, _Context) when is_binary(Binary) ->
+    slice1(z_convert:to_list(Binary), Slice);
 slice(MaybeList, Slice, Context) ->
     slice1(z_template_compiler_runtime:to_list(MaybeList, Context), Slice).
 


mworrell:
I am wondering if we should take into account unicode code points in this case.  I assume you want something like the "last character" of this string?

onnet:
@mworrell Hi Mark.
Yes, I was dealing with "last digit", therefore wasn't thinking about unicode.
It certainly should be taken into account.

mworrell:
Ok, then I merge it and tune your code a bit, ok?

(As your change definitely makes sense!)

mworrell:
@ddeboer I think we need to move to unicode-codepoints strings and utf-8 binaries.  Now many lists are assumed to be utf-8, which is not the normal way strings are represented in Erlang.  What do you think?

onnet:
@mworrell Thanks a lot!

mworrell:
@onnet thanks!

ddeboer:
> I think we need to move to unicode-codepoints strings and utf-8 binaries.

Agreed. Of course with a preference for using binaries where possible (instead of lists), because they are less ambiguous (and slightly faster).

mworrell:
Definitely, we should keep UTF-8 texts as binaries whenever possible

mworrell:
Such a change would make the system more Erlang-strings friendly.  Which can only be good.

PR #1714, opened 2017-05-29
mod_base: extended support for binary input to the slice filter.

mworrell:
### Description

Add *native* binary support to the `slice` filter.

This also changes the behavior of a negative start argument.

Previously the following was documented:

```
{{ [1,2,3,4,5,6,7,8,9,0]|slice:[-7,] }} -> [3,4,5,6,7,8,9,0]
```

But I think it should have been:

```
{{ [1,2,3,4,5,6,7,8,9,0]|slice:[-7,] }} -> [4,5,6,7,8,9,0]
```

As `"1234567890"|slice:[-1,]` should return a string with the last character (`"0"`)

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/modules/mod_base/filters/filter_slice.erl
+++ b/modules/mod_base/filters/filter_slice.erl
@@ -22,12 +22,12 @@
 % Get all elements from element M to element N:
 % {{ list|slice:[3,7] }} -> [3,4,5,6,7]
 % {{ list|slice:[3,-3] }} -> [3,4,5,6,7]
-% {{ list|slice:[-7,-3] }} -> [3,4,5,6,7]
-% {{ list|slice:[-7,7] }} -> [3,4,5,6,7]
+% {{ list|slice:[-7,-3] }} -> [4,5,6,7]
+% {{ list|slice:[-7,7] }} -> [4,5,6,7]
 %
 % Get all elements except the first N:
 % {{ list|slice:[3,] }} -> [3,4,5,6,7,8,9,0]
-% {{ list|slice:[-7,] }} -> [3,4,5,6,7,8,9,0]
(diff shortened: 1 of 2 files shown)


modules/mod_base/filters/filter_slice.erl
```diff
@@ -38,58 +38,54 @@
...
-    slice1(List, [undefined, z_convert:to_integer(M)]).
+substring(<<>>, _, _) -> <<>>;
+substring(B, 1, N) -> substring_1(B, N, <<>>);
+substring(<<_/utf8, B/binary>>, M, N) -> substring(B, M - 1, N);
+substring(<<_, B/binary>>, M, N) -> substring(B, M - 1, N).
 
+substring_1(_, 0, Acc) -> Acc;
+substring_1(<<>>, _, Acc) -> Acc;
+substring_1(<<C/utf8, B/binary>>, N, Acc) -> substring_1(B, N - 1, <<Acc/binary, C/utf8>>);
+substring_1(<<C, B/binary>>, N, Acc) -> substring_1(B, N - 1, <<Acc/binary, C/utf8>>).
```
ddeboer: While this works, using the `binary:part` BIF looks easier to me.

mworrell: Problem is that we want to use utf8 characters, and not bytes.

ddeboer: Good point, I thought `binary:part` was able to handle `/utf8`, but that’s definitely not the case.

ddeboer:
This should also go to 0.x, right?

mworrell:
Could do, but not really needed, as the string behavior on the 0.x branch is different and most strings will be represented as lists of UTF-8 bytes.

PR #1728, opened 2017-06-18
[WIP] core: Add generic task queue

ddeboer:
### Description

* Client modules can add items to the queue.
* This queue can replace z_pivot_rsc for both its task queue and its pivot queue.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### To do

- [ ] Make `z_queue:dequeue` go through a worker pool (sidejob) to speed up processing but do so safely through capacity limiting. Create a pool per module/function queue callback to separate different types of load/backpressure, e.g. (1) CPU-intensive (resource pivoting, database calls), (2) memory-intensive or (3) dependent on external systems (such as an API or document store).

### DIFF
--- a/src/support/z_queue.erl
+++ b/src/support/z_queue.erl
@@ -1,0 +1,197 @@
+%% @doc A generic task queue with prioritized, scheduled and batched execution
+%%      and failure handling through exponential backoff.
+-module(z_queue).
+
+-export([
+    queue/2,
+    start_link/1,
+    init/1,
+    handle_call/3,
+    handle_cast/2,
+    handle_info/2,
+    terminate/2,
+    code_change/3
+]).
+
+-behaviour(gen_server).
+
+-record(state, {
+    host :: atom()
+}).
+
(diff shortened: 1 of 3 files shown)


src/support/z_queue.erl
```diff
@@ -0,0 +1,197 @@
...
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+install(Context) ->
+    case z_db:table_exists(?TABLE, Context) of
+        false ->
+            ok = z_db:create_table(
+            ?TABLE,
+            [
+                #column_def{name = id, type = "serial", is_nullable = false, primary_key = true},
```
ddeboer: Make that a bigserial

mworrell:
Saw that a long time ago I started a review, but apparently never clicked on "submit" ...

mworrell:
Closed.

We made some adjustments to the 1.x task queue system, especially regarding retries and back-off.
Idea to have optionally running multiple tasks in parallel is something that we could investigate.

PR #1733, opened 2017-06-22
build: Test against Erlang/OTP 20

ddeboer:
### Description

* Make Zotonic compatible with Erlang 20. 
* Remove Erlang 18 support.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


ddeboer:
Apparently Travis doesn’t support OTP 20 yet, so we’ll have to wait a bit.

ddeboer:
Now green on Erlang/OTP 20.

@mworrell Please review the changes in z_file_entry, which I changed from a deprecated gen_fsm to the new gen_statem.

src/support/z_file_entry.erl
```diff
@@ -198,89 +201,76 @@ locate(timeout, State) ->
...
     end;
-handle_info(Info, StateName, State) ->
-    lager:info("Unexpected info ~p in state ~p", [Info, StateName]),
+handle_event(EventType, EventContent, StateName, State) ->
+    lager:error("Unexpected event ~p ~p in state ~p", [EventType, EventContent, StateName]),
     {next_state, StateName, State, timeout(StateName, State#state.is_found)}.
 
 code_change(_OldVsn, StateName, State, _Extra) ->
-    {next_state, StateName, State, timeout(StateName, State#state.is_found)}.
+    {ok, StateName, State}.
```
mworrell: Does `statem` keeps its timeout after a code_change?

ddeboer: I’m not sure; couldn’t find this in the docs: http://erlang.org/doc/man/gen_statem.html. 

This callback is optional, and only required for release upgrade (which we currently don’t use, right?). So perhaps it’s best to just remove the function?

PR #1735, opened 2017-06-27
Re-create custom pivot table when definition changes

Dirklectisch:
The schema of a pivot table often evolves over time. Changing a pivot table currently means you have manually delete the table from all environments which is quite a hassle. I propose that we just delete and re-init the table when the definition changes.


mworrell:
I am not sure about the equality check for the columns. As one comes from the spec, and one comes from the database. There could be a (small) textual change between them, and still have the same spec. Such a textual difference would then result in the new pivot table being rebuild on every module init.

Dirklectisch:
You will lose the table if you start renaming columns in it, that's correct. But in normal scenarios the column names in the db and the spec should match.

mworrell:
I was thinking of the *Type* of the columns. Matching on the names is a good idea and stable.

Dirklectisch:
@mworrell Changed the comparison so that it only compares column names.

ddeboer:
LGTM to now. 

@Dirklectisch Can you add a note to [the docs](http://docs.zotonic.com/en/0.x/cookbook/custom-pivot.html) explaining that pivot changes will now re-create the table, losing all (albeit transient) data currently in it?

Dirklectisch:
@ddeboer Docs updated

ddeboer:
Thanks!

PR #1750, opened 2017-07-10
core: Compile the user/ dir for user sites and modules

### LINKED ISSUE #1743: OTP-ification: ensure configured user/modules and user/sites dirs are found by rebar3
This can be done after #1717 has been merged

The directories in `user/modules` and `user/sites` are found via the rebar `project_app_dirs` setting.

Todo:

 - [ ] Add `rebar.config.script` to add the configured user directories

ddeboer:
### Description

Don't check out user sites in `_checkouts`, but add `user/` as an app dir. (This only works for sites and modules that are OTP apps.)

Issue #1743.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -9,3 +9,5 @@
 ]}.
 
 {extra_src_dirs, ["test"]}.
+
+{project_app_dirs, ["apps/*", "lib/*", ".", "user/*", "user/*/*"]}.


mworrell:
I was thinking of using a script and dynamically adding the configured user/modules and user/sites.

ddeboer:
Sure, but this actually works pretty well, too. And it’s standard rebar3. :wink:

mworrell:
Was just wondering, because then you need soft link the *user/...* dirs to the places that you are using. Which we now use that configuration for.

Also doesn't the `user/*/*` not clash with any erlang projects sitting at `user/.` ?

An `rebar.config.script` can actually pick up the configuration and then add the correct paths to the project_app_dirs

The idea to use these is very good, btw.

ddeboer:
Fair enough: we should make this more flexible by wrapping it in a `rebar.config.script`. 

Still, is this change good enough for now? At least it allows us to run custom sites in `user/sites`.

mworrell:
Merge for now, but don't close the issue. I will add the script fingy.

PR #1756, opened 2017-07-13
core: Pivot subject/object when edge is changed

### LINKED ISSUE #1722: On edge insert/delete, pivot both subject and object
Now only the _subject_ is touched, which actually isn’t necessary. Only re-pivoting (instead of touching) the subject is more efficient. If we do this, we can also re-pivot the object, to make sure we have [consistent data](https://github.com/zotonic/zotonic/issues/1698#issuecomment-307908358).

As discussed with @mworrell.

ddeboer:
### Description

core: Pivot subject/object when edge is changed

Fix #1722.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_edge.erl
+++ b/apps/zotonic_core/src/models/m_edge.erl
@@ -216,7 +216,7 @@ insert1(SubjectId, PredId, ObjectId, Opts, Context) ->
             F = fun(Ctx) ->
                 case z_convert:to_bool(proplists:get_value(no_touch, Opts, false)) of
                     true -> skip;
-                    false -> m_rsc:touch(SubjectId, Ctx)
+                    false -> pivot_resources([SubjectId, ObjectId], Ctx)
                 end,
                 SeqOpt = case proplists:get_value(seq, Opts) of
                              S when is_integer(S) -> [{seq, S}];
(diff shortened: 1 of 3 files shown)


fredpook:
Can this be included in the next release, please?

mworrell:
This also changes the behavior that the modification time and modifier of a resource is set if an edge is changed.  Do we want to include this now or first have some further discussion about this?

ddeboer:
> This also changes the behavior that the modification time and modifier of a resource is set if an edge is changed. 

This is the current behaviour. However, I tried to prevent it in this PR by only _pivoting_ (using `z_pivot_rsc:pivot`), not updating, the subject and object.

apps/zotonic_core/src/models/m_edge.erl
```diff
@@ -216,7 +216,7 @@ insert1(SubjectId, PredId, ObjectId, Opts, Context) ->
             F = fun(Ctx) ->
                 case z_convert:to_bool(proplists:get_value(no_touch, Opts, false)) of
                     true -> skip;
-                    false -> m_rsc:touch(SubjectId, Ctx)
+                    false -> pivot_resources([SubjectId, ObjectId], Ctx)
```
ddeboer: @mworrell As you can see, `no_touch` now prevents the subject and object from being pivoted. Should we rename this option to `no_pivot`? I Or simply remove it?

mworrell:
@fredpook you requested to have this in the 0.x, what part do you need there?

fredpook:
This is actually just needed for keeping the elasticsearch index up-to-date.

The ES index has the subject with the outgoing edges and the object with the corresponding incoming edges.
Both will have to be pivotted in order for ES to be consistent.

Issue #1737, opened 2017-06-30
OTP-ification: add-site fixes


mworrell:
This can be done after #1717 has been merged

rebar3 can use templates from plugins.

This is to move all skel files to a *zotonic* plugin.
This can then also add commands to make a new module or site according to one of the templates.

 - [ ] move `priv/skel` files to `apps/zotonic_site_status/priv/`
 - [x] reorganize *skel* projects for OTP compatibilty
 - [x] fix *new site* buttons in zotonic_status (create in *user/sites*)
 - [ ] gracious handling of newly added site or module (rebar compile and code paths)
 - [ ] implement the `addsite` erlscript (should interface to the zotonic_site_status functionality) See also #2184 

Optionally:

 - [ ] Replace skel files with rebar3 templates

PR #1760, opened 2017-07-16
Reorganize priv/skel projects for OTP compatibility (part of #1737)

karlsson:
### Description

Fix part of #1737

Restructure skeleton sites to OTP structure and use rebar3 when deploying site

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
@karlsson thanks!

PR #1763, opened 2017-07-22
mod_seo: Add Schema.org structured metadata

ddeboer:
### Description

* Add structured metadata for most content types.
* Add documentation for the new features.
* Add documentation for some existing features, such as the `noindex` flag.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_category.erl
+++ b/apps/zotonic_core/src/models/m_category.erl
@@ -65,6 +65,7 @@
     fold/4,
 
     move_below/3,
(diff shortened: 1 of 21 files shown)


mworrell:
Wondering if we should always supply the JSON-LD in the page or as a `<link />` element to a JSON-LD file?

Also shouldn't we add a dispatch for the `controller_id`, or is that a separate thing we need to merge from the ginger modules?

PR #1772, opened 2017-07-27
Signup email translation missing punctuation

rl-king:
### Description
Adds missing punctuation to translated sentence "in het invoerveld"


mworrell:
Thanks!

PR #1783, opened 2017-08-16
doc: Document anondo include parameter

### LINKED ISSUE #1780: Include a template as anonymous user
It can be useful to do ACL checks anonymously, e.g. to make sure any user can access some template. `{% include %}` already has a [sudo](http://docs.zotonic.com/en/latest/ref/tags/tag_include.html) argument. What about adding an `anon` or `anonymous` argument too? Any `m.acl.is_allowed…` checks in the included template are then executed against the anonymous user. The extra argument can call `z_acl:anondo/1`.

ddeboer:
### Description

Fix #1780.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Thanks!

PR #1789, opened 2017-08-19
core: Check access rights by default

### LINKED ISSUE #1226: Don't expose unchecked properties in public API
We currently export functions that expose resource properties that have not been checked for ACL permissions. For instance: `m_rsc:get` and `m_rsc:get_raw`.

Proposal:
1. Rename `m_rsc:get` to `m_rsc:get_cached`. Don’t export this function (it’s for internal use only).
2. Rename `m_rsc:get_visible` to `m_rsc:get`.
3. Don’t export `m_rsc:get_raw`.
4. Document this and note that developers can always get the full resource by passing a `z_acl:sudo(Context)`.

Are there are other modules where we expose unchecked data?

ddeboer:
### Description

Fix #1226.

* Add notes to Upgrading doc.
* Add m_rsc:get/3 with Options.
* Remove m_rsc:get_raw/2 and m_rsc:get_raw_lock/2.
* Remove m_rsc:get_visible/2.

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_category.erl
+++ b/apps/zotonic_core/src/models/m_category.erl
@@ -379,7 +379,7 @@ prune(N, CS) ->
 
 
 %% @doc Get the basic properties of a category
--spec get(m_rsc:resource(), #context{}) -> list() | undefined.
+-spec get(m_rsc:resource(), z:context()) -> list() | undefined.
 get(undefined, _Context) ->
     undefined;
 get(Id, Context) when is_integer(Id) ->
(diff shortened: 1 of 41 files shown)


ddeboer:
We still need `m_rsc:p_no_acl/3` as it is called from within the ACL routines. Calling `m_rsc:p/3` instead causes an infinite recursion (as it goes again through the ACL routines).

mworrell:
Good one, the ACL routines are indeed a prime user of the `p_no_acl/3` function. And for obvious reasons, as you state.  So let's keep that one in and just add some very clear documentation *why* it is there.

mworrell:
I am thinking of redoing this, keeping the `p_no_acl/3` routine, but changing the `m_rsc:get` routines so that admin rights are needed if the raw version is requested.

ddeboer:
So that would mean keep what we have here but export `m_rsc:p_no_acl/3` (for the ACL routines), right?

mworrell:
Yes, keep everything (especially the `get` changes!) and export `p_no_acl` for the low level routines like ACLs.

Most important is that the raw get and `m_get` routines are checked with the ACL.

mworrell:
Property checks on `m_rsc:get/2` and removal of `m_rsc:get_visible/2` has been added to master in #1961

mworrell:
The functions `m_rsc:get/2`, `m_rsc:get_raw/2` and `m_rsc:get_raw_lock/2` now all filter their results on visibility.

Closing this issue.

mworrell:
@ddeboer I left the branch intact, so we can still decide if we want to add _get with options_.

PR #1794, opened 2017-08-23
Authenticate using GitHub

mworrell:
### Description

Preparation for mod_github: authenticate using GitHub.

Add GitHub icons.
Also add 'target=_blank' for auth links, in case the popup-window js is not loaded.

We need to decide if `mod_github` should be in the core or outside.

For now I keep it outside. We could decide to move it into the core if we want tighter integration with GitHub?  @ddeboer what do you think?

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
See also https://github.com/zotonic/mod_github

ddeboer:
I agree with having it separate from Zotonic core (but under the Zotonic namespace, so all of us can edit it).

ddeboer:
@mworrell Please merge into master.

mworrell:
Merged.

PR #1806, opened 2017-08-31
tests: Raise timeout to fix build failures

ddeboer:
### Description

Not pretty, but seems to be the only way to get the tests running reliable again.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
All in support for this, as we did fix the master, and spending time on this issue on 0.x is wasteful.

PR #1807, opened 2017-08-31
core: Use ceil/1 BIF

ddeboer:
### Description

Fix "Warning: ambiguous call of overridden auto-imported BIF ceil/1"

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


ddeboer:
This (of course) only works on Erlang/OTP 20, and we currently have 19 as the minimum version.

mworrell:
How far is cowboy with demanding at least OTP 20?

ddeboer:
Cowboy currently only requires 19. Not sure whether they have a reason to switch to 20.

mworrell:
I propose to fix this with a bigger update removing all warnings of this kind.

PR #1808, opened 2017-09-01
Add livereload client to mod_development

rl-king:
### Description

This PR adds the [livereload](https://github.com/livereload/livereload-js) script to the document head if mod_development is enabled. The script attempts to connect to a livereload server over a websocket which will tell the browser when and what to reload or replace. 

If there is no livereload server running the the script will re-attempt to connect on a set (but configurable) interval. This will result in a browser console error for each attempt.

### Checklist

- [] documentation updated
- [] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_development/lib/js/mod_development-livereload.js
+++ b/modules/mod_development/lib/js/mod_development-livereload.js
@@ -1,0 +1,10 @@
+(function () {
+
+    var url = 'http://' + (location.host || 'localhost').split(':')[0] + ':35729/livereload.js?snipver=1';
+    var script = document.createElement('script');
+
+    script.src = url;
+
+    document.head.appendChild(script);
+
+}());
--- a/modules/mod_development/templates/_html_head.tpl
+++ b/modules/mod_development/templates/_html_head.tpl
@@ -1,0 +1,5 @@
+{#
+    The following script adds the livereload script-tag to the document <head>
+#}
+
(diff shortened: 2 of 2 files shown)


mworrell:
As we track all changes to files, we do now when what changes.
These changes could be pushed to the client, which can then decide what to do.

The push could just be done by listening to a topic, like `~site/development/filechange` where we can publish the file that is changed, and how it is changed. We should also publish the *lib* url for the changed file (if it is a lib file).

The page subscribes to this topic and then passes the received file to the

Is there a local API for the *livereload.js* that can be used?

rl-king:
I believe @ddeboer had given this some thought.

mworrell:
After a quick glance into the live reload sources it looks like that on CSS changes it wants to match the correct filenames. This won't work well with our lib tags, which combine several files and also attach a (version-)hash.

I think on CSS changes we can ping the client and let it fetch new combined css urls to replace the existing link elements (if changed).

That would work better with how we load our combined css files.

Of course, we already have automatic less/scss/coffee compilation. Though we might want to make that a bit more flexible.

Didn't check how live reload handles javascript changes.

ddeboer:
Discussed this with @mworrell @rl-king. We decided not to adopt Livereload but create a simpler solution instead (on master) that works well with Zotonic’s concatenated files.

@rl-king and others: if you want to use Livereload, you can still do so by installing the browser plugin.

PR #1815, opened 2017-09-26
[WIP] Fix postgres unavailability

### LINKED ISSUE #1776: More gracefully handle postgresql unavailability
If postgres is unavailable for a moment, zotonic will stop random sites (depending on which have activity during that time problably) and never restart them. 

Database connections should be retried indefinitely instead of simply crashing permanently; the unavailability could just be temporary for an update or something.

ddeboer:
### Description

This doesn’t fully work yet, as introducing error handling causes changes downstream, such as the status site not working properly. Submitting as a PR as a reference for future work on the issue (probably on master).

Fix #1776.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/src/db/z_db_pgsql.erl
+++ b/src/db/z_db_pgsql.erl
@@ -45,7 +45,7 @@
 
 -define(IDLE_TIMEOUT, 60000).
 
--define(CONNECT_RETRIES, 50).
+-define(CONNECT_RETRIES, 500).
 -define(CONNECT_RETRY_SLEEP, 10000).
 -define(CONNECT_RETRY_SHORT, 10).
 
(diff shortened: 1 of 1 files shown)


crownedgrouse:
could this keep pages cache longer in case of database access failure ? this way at least web site would continue to show things ? (would not be possible for admin/edition access).

mworrell:
For consistency it is actually quite hard to assume anything if the db comes back after a prolonged outage. So after it came back we *should* flush caches.

All other error handling is for later - for now I am working on master to get smoother start/restart behaviour in the presence of db connectivity issues.

crownedgrouse:
Yes I agree when db comes back, caches have to be flushed, but when db is down, is there any way to let caches valid until it comes back ? I do not have clear ideas on how this work on zotonic.

mworrell:
We can now try to move some of these changes to the master, as the stability and restart behavior of the sites manager is greatly improved.

NB: This won't be merged into 0.x due to the huge code differences.

mworrell:
Closed, please continue at #1826 which is the same work, but then on master.

PR #1822, opened 2017-10-11
Refactor sites manager

mworrell:
### Description

This helps with issue #1815 

Major changes to the sites- and modules managers.

 1. Use normal supervisors, remove z_supervisor and z_supervisor_named
 2. A backoff handling when modules or sites repeatedly fail
 3. Always try to restart a failed site or module (after the backoff)
 4. API changes (especially for z_sites_manager).

A further cleanup and refactor would be nice, and can be done after this.
Especially:

 * Split the module scanning and site scanning code from this into a separate (zotonic-wide) module.
 * Move all SQL from z_module_manager to a separate module model.
 * Easier/better API to inspect the status of running sites and modules (should be similar for both).
 * More maps, less proplists...


mworrell:
@ddeboer am working on the tests

mworrell:
@ddeboer Tests are passing, guess we can merge 💥

PR #1825, opened 2017-10-12
Delay requests till site is running.

mworrell:
### Description

Delay HTTP and SMTP requests till sites are running.

This checks on a request if the site is running.
If not then it might delay the request and start the site.
If a site crashed then this will delay the request to wait for a restart.

Now the sites dispatcher also knows of the hostnames of failed and not-yet started sites.

TODO:

 - [x] Decide on 400 and 503 error page content
 - [x] Check 'todo' in z_sites_dispatcher for some redirects

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@ddeboer @ArthurClemens @mmzeeman @arjan do you guys have any idea for a generic 400 and 503 error?

The 400 is shown if there is no host available to route the request (and there is no generic fallback site like `zotonic_site_status`).

The 503 is shown if a site takes too long to start or is currently in `failed` state during a longer backoff.  (We wait up to 30 seconds to see if the site becomes available, after that a 503 is returned.)

mworrell:
@ddeboer @mmzeeman Ready for merge, admin etc works. I think save to merge with master (as it does fix some issues) so that we can continue with the next issues.

mworrell:
I just made very bare bones html error pages.
They are in `zotonic_core/priv/htmlerrors`

mworrell:
As this fixes quite some unrelated things, I just go ahead and merge it.

PR #1826, opened 2017-10-13
Add more psql connect error handling and better connect retries.

### LINKED ISSUE #1776: More gracefully handle postgresql unavailability

mworrell:
Copied from #1815, now on master instead of 0.x

This doesn’t fully work yet, as introducing error handling causes changes downstream, such as the status site not working properly. Submitting as a PR as a reference for future work on the issue.

Fix #1776.

- [x] Start Zotonic without psql, does it start?
- [ ] Stop psql with running Zotonic, do sites stop?
- [x] Start psql with running Zotonic, do sites restart?
- [ ] Quick(er) stop/start of psql, are sites still running?

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/db/z_db_pgsql.erl
+++ b/apps/zotonic_core/src/db/z_db_pgsql.erl
@@ -43,6 +43,7 @@
 
 -define(TERM_MAGIC_NUMBER, 16#01326A3A:1/big-unsigned-unit:32).
 
+-define(CONNECT_TIMEOUT, 5000).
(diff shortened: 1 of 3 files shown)


mworrell:
Merging - it is still a bit noisy with crash messages, but is functional.

Issue #1793, opened 2017-08-23
Livereload for templates, CSS and JS


ddeboer:
I propose an out-of-the-box solution for reloading the page when templates, CSS and/or JS files have changed. 

### Solution

We can either run some minimal JS in the browser when files change (`location.reload(true);`) or rely on a more fully-fledged [LiveReload](https://github.com/livereload/livereload-js#what-is-livereloadjs) solution. LiveReload has a [simple WebSocket protocol](http://livereload.com/api/protocol/).

Zotonic [already watches](https://github.com/zotonic/zotonic/blob/f4fef7f95233f00fdb39a913742a947c3b076430/apps/zotonic_launcher/src/filewatcher/z_filewatcher_handler.erl) templates, CSS and JS files, so this should not be hard to add. Add this functionality to mod_development, so it can easily be enabled/disabled depending on the environment (development vs. production).

### Separation of concerns

- Have Zotonic only watch _compiled_ assets.
- Other tools (such as Webpack or Gulp) take care of compiling SCSS and JS to CSS and (minified) JS. This way, developers keep full control over their frontend stack.

### Alternatives

- Document how IDEs can be configured to enable Livereload. 

### To do

- [ ] update [docs](http://docs.zotonic.com/en/latest/ref/modules/mod_development.html)

ddeboer:
As discussed in #1808:

> We decided not to adopt Livereload but create a simpler solution instead (on master) that works well with Zotonic’s concatenated files.

mworrell:
Currently the file watcher is part of `zotonic_launcher`.  This module also handles the recompilation of changed files.

It seems more logical to me to move the recompilation to `zotonic_mod_development`, at least if it concerns handling of *lib* files.  The recompilation of Erlang files could stay in the `zotonic_launcher`.

mworrell:
Moving the file watcher to its own app.

I saw in the inotify code a `killall inotifywait`. I suspect this is because on a close of the beam.smp the started process does not terminate.

I propose to remove this and start using https://github.com/saleyn/erlexec instead.
It is also on hex.pm: https://hex.pm/packages/erlexec

mworrell:
The refactoring of the filewatcher takes place here https://github.com/zotonic/zotonic/issues/1812

mworrell:
I have added a simple live reload to th mod_development (0.x only for now).
See merge request #1924 

/cc @rl-king

mworrell:
[...]

PR #1828, opened 2017-10-14
Refactor the filewatcher into its own application.

### LINKED ISSUE #1812: Move source (less/sass/etc) files out of public lib directory
Currently the `.less`, `.sass` and build-tool configs for generated css/js files are placed in the *lib* directory.

These source files should not be loadable by the client, so should be moved to another directory.

I propose to add a `lib-src` directory on the same level as the `lib` directory.

```
priv/
       lib/
       lib-src/
...
```

The *changed file handler* could then watch changes in these directories to know that a build tool must be called.

mworrell:
### Description

Fix #1812 
Also parts of #1793 

Refactor file watcher into its own application.

This uses `erlexec` to run external commands.
`erlexec` doesn't support Windows, so that might be a problem.

To do:

 - [x] refactor z_notifier into its own app, enabling system-wide notifications
 - [x] add `zotonic_mod_cron` for the `tick_...`notifications
 - [x] add (system-level) notification on changed files
 - [x] lib / lib-src support
 - [x] make changed file handler a process to prevent rapid similar updates
 - [x] decouple handling of translations, dispatch, and module indexer
 - [x] testing

Move to #1737:
  - gracious handling of newly added site or module

Not really needed, so not done:
 - move recompilation, sass handling etc, to `zotonic_mod_development` application

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
From #1793:

Currently the file watcher is part of zotonic_launcher. This module also handles the recompilation of changed files.

It seems more logical to me to move the recompilation to zotonic_mod_development, at least if it concerns handling of lib files.

Actually, looking at the code, I propose to:

Move the file watcher to its own app zotonic_filewatcher
zotonic_filewatcher will handle Erlang and beam files).
Handling of priv files to separate modules (via a notification).
The zotonic_mod_development will then handle the all less, sass, etc. files (unless another module picks up the notification).

mworrell:
For the less/sass (etc) handling:

 * Move all less/sass files to `priv/lib-src/`
 * Added `build.sh` scripts which are run to build the css if there are changes
 * Added `build-targets` files which contain (one per line) the files generated by `build.sh`

`build.sh` should take the input files from `priv/lib-src/` and generate output files in `priv/lib/`

The build script is found by checking the current directory and then walking up till the lib-src dir.

mworrell:
Replace the `build.sh` scripts with a`Makefile` which is executed in the directory where the file is placed.

mworrell:
@ddeboer I still have some calls to `z_utils:os_escape/1` in the filehandler  Would be good to move these routines to `zotonic_stdlib`, am just wondering to what module.

mworrell:
@ddeboer @ArthurClemens ready for merge.  Are you ok with the 

We can do the move of the *z_utils* references in another issue (good to clean that one up anyway).
At the moment this just means that `zotonic_filewatcher` and `zotonic_filehandler` are not completely decoupled from `zotonic_core`.

PR #1846, opened 2017-11-09
Simplify models, remove #m{...}.

### LINKED ISSUE #1837: Re-model the models - simplification
The models now use a recursive callback for resolving nested lookups in a `#m{}`

The new `template_compiler` already supports a direct lookup of a nested property (example: `m.foo.bar.plop.pliep`)

Proposal is to remove the recursive callback and to replace it with the nested lookup.

This enables opening the models to the API and the MQTT topic tree.

mworrell:
### Description

Fix #1837

This replaces all model callbacks with a single m_get/2 function.

The `m_get/2` is now directly called, with a list corresponding to the dotted expression and the current context.

Example:

```
     {{ m.foo.bar.pief[2].poef }}
```

Calls:

```Erlang
   m_foo:m_get([ bar, pief, 2, poef ], Context)
```

This function returns a tuple with the result and the part of the dotted expression that was not handled:

```Erlang
  {42, [ poef ]}
```

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] ~no BC breaks~ **Definitiely BC breaks**

### DIFF
--- a/apps/zotonic_core/src/models/m_acl.erl
+++ b/apps/zotonic_core/src/models/m_acl.erl
@@ -1,10 +1,8 @@
 %% @author Marc Worrell <marc@worrell.nl>
-%% @copyright 2009 Marc Worrell
-%% Date: 2009-04-27
-%%
+%% @copyright 2009-2017 Marc Worrell
 %% @doc Template access for access control functions and state
 
-%% Copyright 2009 Marc Worrell
+%% Copyright 2009-2017 Marc Worrell
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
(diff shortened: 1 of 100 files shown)


apps/zotonic_core/src/models/m_acl.erl
```diff
@@ -25,32 +23,63 @@
...
+    Context1 = case z_notifier:first(#acl_context_authenticated{}, Context) of
+                    undefined -> Context;
+                    Ctx -> Ctx
+               end,
+    {z_acl:is_allowed(Action, Object, Context1), Rest};
+
+% Error, unknown lookup.
+m_get(Vs, _Context) ->
+    lager:info("Unknown ~p lookup: ~p", [?MODULE, Vs]),
+    {undefined, []}.
```
ddeboer: Why not just crash here? Is the error message too unclear?

mworrell: It is called from templates, best to give some feedback and continue.

I also used it to fix any problems in the transition (as you shouldn't get there).

mworrell: But indeed, it could be removed.

ddeboer: I’m okay with keeping it there, but then we should change it to `lager:error` for proper logging.

ddeboer:
@arjan Do you agree with this change? If so, @mworrell can you push some updated documentation?

arjan:
What's the idea behind also returning the part of the dotted expression that was not handled?

mworrell:
That is because the template compiler doesn't know what part of the list will be handled.  For example:

```
{{ m.search[{latest cat=`news`}].total }}
```

The m_get would be called as:

``
m_get([ {latest, [{cat, news}]}, total ]
``
Here we would return:

```
{#m_search_result{...}, [ total ]}
```

So that the template proceeds with lookup up `total` in the search result.

mworrell:
BTW the long term plan with this change is to also open the models directly for HTTP and MQTT API requests.

apps/zotonic_core/src/models/m_acl.erl
mworrell: Done, now an error

mworrell:
@arjan @ddeboer This is ready for merging (if no objections).

PR #1847, opened 2017-11-09
[mod_export] Observe content disposition

fredpook:
Rss feeds are currently handled by browsers as a download while the Rss reader plugins want Content-Disposition inline.
This PR allows Content-Disposition to be inline instead of attachment.
It also fixes a problem with unpublished resources to show up in feeds.

### DIFF
--- a/modules/mod_export/controllers/controller_export_resource.erl
+++ b/modules/mod_export/controllers/controller_export_resource.erl
@@ -112,24 +112,26 @@ set_filename(Id, ContentType, Dispatch, Context) ->
                     Exts -> binary_to_list(hd(Exts))
                 end,
     {ok, Disposition} = z_notifier:first(#export_resource_content_disposition{id=Id, dispatch=Dispatch, content_type=ContentType}, Context),
-    case z_notifier:first(#export_resource_filename{id=Id, dispatch=Dispatch, content_type=ContentType}, Context) of
+
+    Filename = case z_notifier:first(#export_resource_filename{id=Id, dispatch=Dispatch, content_type=ContentType}, Context) of
         undefined ->
             Cat = m_rsc:p_no_acl(Id, category, Context),
-            Filename2 = "export-"
-                        ++z_convert:to_list(proplists:get_value(name, Cat))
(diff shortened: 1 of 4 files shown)


modules/mod_export/support/export_encoder_atom.erl
```diff
@@ -54,11 +54,16 @@ header(Header, #state{id=Id} = State, Context) ->
     {ok, iolist_to_binary(Bin), State}.
 
 row(Row, #state{} = State, Context) when is_integer(Row) ->
-    Vars = [
-        {id, Row}
-    ],
-    {Bin, _Context} = z_template:render_to_iolist({cat, "atom/entry.tpl"}, Vars, Context),
-    {ok, iolist_to_binary(Bin), State};
+    case m_rsc:p(Row, is_published, Context) of
```
ddeboer: Should we add an ACL check here? If so, it should be `is_visible`.

modules/mod_export/controllers/controller_export_resource.erl
```diff
@@ -111,29 +111,25 @@ set_filename(Id, ContentType, Dispatch, Context) ->
...
-                                id=Id,
-                                dispatch=Dispatch,
-                                content_type=ContentType}, Context)
-    of
+    {ok, Disposition} = z_notifier:first(#export_resource_content_disposition{id=Id, dispatch=Dispatch, content_type=ContentType}, Context),
+    case z_notifier:first(#export_resource_filename{id=Id, dispatch=Dispatch, content_type=ContentType}, Context) of
         undefined ->
             Cat = m_rsc:p_no_acl(Id, category, Context),
-            Filename = "export-"
+            Filename2 = "export-"
```
ddeboer: The `Filename` variable should be moved before the `case`:

```erlang
Filename = case z_notifier:first(#export_resource_filename{id=Id, dispatch=Dispatch, content_type=ContentType}, Context) of
    undefined ->
        ...
```

ddeboer:
LGTM now. @mworrell Please have a look and squash & merge if okay.

PR #1848, opened 2017-11-10
Simplify module data management by adding manage_data/2 callback. Fix…

### LINKED ISSUE #1834: Access control rules: default rules are missing
From Gitter:

> There should at least be “good defaults” in place so that the site works as expected and that you can learn from the existing settings.

> There is actually some "default rules install", but it doesn't seem to be installed.

mworrell:
### Description

Fix #1834

Simplify data model installation by splitting it into two steps and removing the notification.

Two steps are now:

 * manage_schema/2
 * manage_data/2

Note that `manage_data/2` is called if and only if the `manage_schema/2` is called.

The previous observer was never called as the module's observers get only installed after the datamodel installation.

### TODO

- [x] OK on ACL rules?  ( esp @ddeboer :) )  See: https://github.com/zotonic/zotonic/blob/module-manage-data/apps/zotonic_mod_acl_user_groups/src/support/acl_default_rules.erl

Add to acl_rules:

 - [x] use mod_seo
 - [x] use mod_translation
 - [x] use mod_rest

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks (some, documented)


mworrell:
@ArthurClemens @ddeboer @arjan Please check the default access control rules in https://github.com/zotonic/zotonic/blob/7e7de67f9d7d6b93153062821a0377b9a66a4e71/apps/zotonic_mod_acl_user_groups/src/support/acl_default_rules.erl

mworrell:
Toe te voegen:

 - [ ] mod_seo
 - [ ] mod_translation
 - [ ] mod_rest

doc/developer-guide/modules.rst
```diff
@@ -367,6 +367,19 @@ installation and upgrade of data. You can define:
...
+``manage_data/2`` function is called. The function ``manage_data/2`` is
+called if and only if the ``manage_schema/2`` is called. If you only want
+a ``manage_data/2`` function, then add a dummy ``manage_schema/2`` function
+that returns `ok` and does nothing else.
+
+.. note::
+
+    The function ``manage_schema/2`` is called inside a transaction, so that any
+    installation errors are rolled back. ``manage_data/2`` Is called outside
+    a transaction, and after all resources, predicated etc. are installed, but
```
ddeboer: predicates

ddeboer:
What exactly do we gain by this change:

1. compared to `#datamodel{}.data`?
2. compared to just adding the extra logic in `manage_schema/2` (by calling `z_datamodel:manage/3` manually and then doing extra things like adding ACL rules, updating resources etc.)?

mod_acl_user_groups docs should also be updated to reflect these changes.

mworrell:
> Interesting! But what exactly do we gain by this change:
> 
> 1. compared to #datamodel{}.data?

You can use logic to manipulate data and lookup ids.
Which is quite essential, otherwise we needed to add a function as #datamodel.data

> 2. compared to just adding the extra logic in manage_schema/2 (by calling z_datamodel:manage/3 manually and then doing extra things like adding ACL rules, updating resources etc.)?

`manage_schema/2` is called within a transaction, so that any database changes are rolled back and we don't end up with a half changed database.

`manage_data/2` is called after the `manage_schema/2` transaction and after all `#datamodel` changes.

> mod_acl_user_groups docs should also be updated to reflect these changes.

I assume you mean the default rules?
Yes, after we finalized those default rules.

ddeboer:
> I assume you mean the default rules?

I meant the [Managed rules](http://docs.zotonic.com/en/latest/ref/modules/mod_acl_user_groups.html#managed-rules) section.

mworrell:
Updated the ACL rules.

This is ready to be merged.

doc/developer-guide/modules.rst
mworrell: Fixed

mworrell:
@ddeboer can you check the default ACL rules?  If you are ok, then this can be merged.

PR #1853, opened 2017-11-15
Add security for template access to m.config, m.site and other models.

### LINKED ISSUE #1849: DISCUSS: ACL checks in models
In #1226 we are working on making the `m_rsc` model safer to use.

I think we might want to tighten other models as well. 
Just think what `m_identity` and `m_config` can do...

Problem with m_config is that it contains a mix of public, harmless, and very sensitive settings.

What could be a good way of separating the settings, so that templates (and APIs) can only access configurations if and only if they have the appropriate access rights?

We could apply one or more of the following:

 1. Heuristics (anything with 'secret', 'password' or 'account' needs admin rights)
 2. List of known *sensitive* configs
 3. Special 'is_secret' flag per config

What do you think?

In the same manner we can add extra security to the other models.

I am especially thinking of this as I want to:

 1. Open the models via HTTP and MQTT APIs (per model opt-in).
 2. Make rendering any template safe (to render random templates via APIs)

/cc @ddeboer @mmzeeman @arjan

mworrell:
### Description

Fixes #1849 

This adds basic security to the configuration-related models.

The`m.config` model is now only accessible for users with admin rights.
Same for most of the properties of the `m.sysconfig` and `m.site` models.

Some models are receiving better ACL checks.

### Todo

 - [x] update upgrade notes
 - [x] template (admin) testing

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] ~~no BC breaks~~ This introduces BC breaks for templates that access the config model.


mworrell:
@ddeboer This can be merged, if you are ok with the changes.

/cc @arjan @mmzeeman @ArthurClemens

doc/ref/models/model_config.rst
```diff
@@ -8,6 +8,12 @@ Zotonic has two places where a site’s configuration is kept:
 - the site’s ``config`` database table. Entries in the config table overrule
   any module settings from the config file.
 
+.. note::
+
+    Configuration keys are only accessible from templates using ``{{ m.config }}``
+    for users with administrator rights. To acces other configs, use or add
```
ddeboer: ‘access’

apps/zotonic_core/src/models/m_sysconfig.erl
```diff
@@ -31,8 +31,11 @@
 
 %% @doc Fetch the value for the key from a model source
 -spec m_get( list(), z:context() ) -> {term(), list()}.
-m_get([ Key | Rest ], _Context) ->
-    {z_config:get(Key), Rest};
+m_get([ Key | Rest ], Context) ->
+    case z_acl:is_admin(Context) of
+        true -> {z_config:get(Key), Rest};
+        false -> {undefined, Rest}
+    end;
```
ddeboer: Wow, this module already existed? :wink:

doc/developer-guide/upgrading.rst
```diff
@@ -198,6 +198,24 @@ Templates
...
+configurations.
+
+Examples:
+
+ * ``m.config.site.title.value`` is now ``m.site.title``
+ * ``m.config.mod_editor_tinymce.version.value`` is now ``m.editor_tinymce.version``
+
+Check the various models of the modules for the new lookups.
+
+
```
ddeboer: This can go under the ‘Templates’ header above. I like the before/after paradigm as I introduced it in other places in the Upgrading chapter.

apps/zotonic_core/src/models/m_sysconfig.erl
mworrell: Yes, surprise, surprise :-)

mworrell:
@ddeboer done, and also added some extra security to the mailing list module.

PR #1862, opened 2017-12-14
mod_oauth: support OAuth signature check on POST to controller_api.

mworrell:
### Description

Add POST support of OAuth signature check on API calls.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Tested in production, merging.

PR #1863, opened 2018-01-02
[WIP]Convert bash scripts to erlang modules/escripts 

ffringel:
### Description

Convert Zotonic start scripts from bash to erlang modules and escripts


mworrell:
Hi @ffringel, I am curious how far you got with this?  Would love to use your escripts instead of the shell scripts we have now.

mworrell:
Superseded by #2184

PR #1878, opened 2018-02-12
mod_base: Fix Firefox error when logging in (#1877)

### LINKED ISSUE #1877: Firefox shows 'Fail: error' when logging in
Firefox (Quantum 58.0.2) doesn’t seem to be happy about the WebSocket connection cleanup that was done in 1a2132654047fe50125bcab0d9f5822325357ad8 (introduced in Zotonic 0.35.0). While logging in continues to work, this is shown in the console:

screen shot 2018-02-08 at 15 59 22

Is this caused by a new session being started when the user logs in?

ddeboer:
### Description

Fix #1877.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_base/lib/js/apps/zotonic-1.0.js
+++ b/modules/mod_base/lib/js/apps/zotonic-1.0.js
@@ -897,8 +897,13 @@ function z_ajax(options, data)
         error: function(xmlHttpRequest, textStatus, errorThrown)
         {
             z_stop_spinner();
-            $.misc.error("FAIL: " + textStatus);
-            z_unmask_error(options.trigger_id);
+
+            if (!z_page_unloading) {
+                $.misc.error("FAIL: " + textStatus);
+                z_unmask_error(options.trigger_id);
+            } else {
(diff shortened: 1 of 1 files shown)


ddeboer:
@mmzeeman Unfortunately, your change doesn’t fix the issue.

mmzeeman:
Sorry to hear that @ddeboer. I still can't reproduce the error myself. Does it also happen when you login to zotonic's status site (localhost)?

The fix actually fixes a problem during login.

The grey bullet means that there is no response (yet). Is this request cancelled?

What is the error you are getting now? What is also strange is that there are two websocket connections after the logon. Is there a frame on the page? (i'm trying to reproduce it)

mmzeeman:
I've seen requests getting lost (no response on client side) by zotonic before. In that case it was caused by a bug which was caused by an unexpected message which arrived in mochiweb which caused a crash which was silently ignored. I just want to make sure that this connection drop is not caused by a hidden crash somewhere before applying the proposed ostrich algorithm.

PR #1885, opened 2018-03-08
core: Don't default start date to now

### LINKED ISSUE #1536: From-date behaviour when only entering Till-date on resource
When I leave the 'From' date blank for a resource and I enter a 'Till' date, today's date is saved as the From date. This is not visible untill the admin page is refreshed. I might not know a person's date of birth or an event's start date, so I would expect the field to stay either blank or get an error message.

ddeboer:
### Description

Fix #1536.

Anything I’m missing here? Why actually would we want to default start date to now if an end date is given? Why not just leave the start date as entered (i.e., empty)? Or does this also involve searches/pivots?

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/src/models/m_rsc_update.erl
+++ b/src/models/m_rsc_update.erl
@@ -929,8 +929,17 @@ recombine_dates(Id, Props, Context) ->
     {Dates1, DateGroups} = group_dates(Dates),
     {DateGroups1, DatesNull} = collect_empty_date_groups(DateGroups, [], []),
     {Dates2, DatesNull1} = collect_empty_dates(Dates1, [], DatesNull),
-    Dates3 = [ {Name, date_from_default(LocalNow, D)} || {Name, D} <- Dates2 ],
-    DateGroups2 = [ {Name, dategroup_fill_parts(date_from_default(LocalNow, S), E)} || {Name, {S,E}} <- DateGroups1 ],
+    Dates3 = [
+        {Name, date_from_default(default_date(Name, LocalNow), D)}
+        || {Name, D} <- Dates2
+    ],
+    DateGroups2 = [
+        {Name, dategroup_fill_parts(
+                    Name,
+                    date_from_default(default_date(Name, LocalNow), S),
+                    E)}
+        || {Name, {S,E}} <- DateGroups1
+    ],
     Dates4 = lists:foldl(
                     fun({Name, {S, E}}, Acc) ->
                         [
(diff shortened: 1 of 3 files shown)


mworrell:
Have to review this - there is some tricky logic behind most of the date calculations.

crownedgrouse:
But smart things to do with COALESCE in postgresql for empty column...

ddeboer:
Ping @mworrell: I’d like to get this into the next release.

mworrell:
We could do the *coalesce* during pivot (as we need to provide a date anyway).
Same for the `date_end` where we put *St Juttemis* (year 9999) in the end date.

We need to check Erlang code for some range checks, very likely just some filters like `in_future` and `in_past`.

While doing so we could also remove the *St Juttemis* value of `date_end` and only use that for `pivot_date_end`.

The page https://www.postgresql.org/docs/9.1/static/datatype-datetime.html mentions that the minimum date is 4713BC, so we could put that date in when the date_start is empty.

mworrell:
@ddeboer can you check this with your project?

ddeboer:
```erlang
m_rsc:update(402, [{"dt:y:1:date_end", "-1000"}], C).
** exception error: no function clause matching calendar:last_day_of_the_month(-1000,12) (calendar.erl, line 240)
     in function  m_rsc_update:dategroup_fill_parts/3 (/opt/zotonic/src/models/m_rsc_update.erl, line 1120)
     in call from m_rsc_update:'-recombine_dates/3-lc$^1/1-1-'/2 (/opt/zotonic/src/models/m_rsc_update.erl, line 937)
     in call from m_rsc_update:recombine_dates/3 (/opt/zotonic/src/models/m_rsc_update.erl, line 936)
     in call from m_rsc_update:normalize_props/4 (/opt/zotonic/src/models/m_rsc_update.erl, line 512)
     in call from m_rsc_update:update_normalize_props/3 (/opt/zotonic/src/models/m_rsc_update.erl, line 302)
```

src/models/m_rsc_update.erl
```diff
@@ -1086,39 +1095,61 @@ group_dates(Dates) ->
...
+dategroup_fill_parts( Name, {{Ys,Ms,Ds},{Hs,Is,Ss}}, {{undefined,Me,De},{He,Ie,Se}} ) when is_integer(Ys) ->
+    dategroup_fill_parts( Name, {{Ys,Ms,Ds},{Hs,Is,Ss}}, {{Ys,Me,De},{He,Ie,Se}} );
+dategroup_fill_parts( Name, {{Ys,Ms,Ds},{Hs,Is,Ss}}, {{Ys,undefined,De},{He,Ie,Se}} ) when is_integer(Ms) ->
+    dategroup_fill_parts( Name, {{Ys,Ms,Ds},{Hs,Is,Ss}} ,{{Ys,Ms,De},{He,Ie,Se}} );
+dategroup_fill_parts( Name, {{Ys,Ms,Ds},{Hs,Is,Ss}}, {{Ys,Ms,undefined},{He,Ie,Se}} ) when is_integer(Ds) ->
+    dategroup_fill_parts( Name, {{Ys,Ms,Ds},{Hs,Is,Ss}}, {{Ys,Ms,Ds},{He,Ie,Se}} );
+dategroup_fill_parts( Name, S, {{undefined,Me,De},{He,Ie,Se}} ) ->
+    dategroup_fill_parts( Name, S, {{9999,Me,De},{He,Ie,Se}} );
+dategroup_fill_parts( Name, S, {{Ye,undefined,De},{He,Ie,Se}} ) ->
+    dategroup_fill_parts( Name, S ,{{Ye,12,De},{He,Ie,Se}} );
```
ddeboer: Why assume `12` here? This causes the error posted below.

mworrell:
I don’t see your comment in the pull request, somehow.

The ’12’ is added because it is an end date, and the ’12’ is the last month of the year.

This to year ranges work.

mworrell:
The `calendar:last_day_of_the_month/2` doesn't work with negative years, doesn't it?
We might want to add a wrapper that can handle BCE years.

mworrell:
Added `z_datetime:last_day_of_the_month/2`, which uses the Julian rules for dates before Oct 1582.

```erlang
(zotonic001@Lamma)5> m_rsc:update(1, [{"dt:y:1:date_end", "-1000"}], C).          
{ok,1}
(zotonic001@Lamma)6> m_rsc:p(1, date_end, C).
{{-1000,12,31},{23,59,59}}
```

We now use the same leap year calculations as PostrgreSQL.
PostgreSQL uses internally the Julian day numbers (which starts at -4713 BCE) with the Gregorian leap year definitions.

ddeboer:
@mworrell Can you cherry-pick this one to master?

mworrell:
Manually merged it into master.

PR #1890, opened 2018-03-19
Set session cookie to secure if site protocol is https. Fixes #1889

### LINKED ISSUE #1889: Add 'secure' to session cookie if terminated by proxy
Right now `mod_ssl` adds the `secure` attribute to the session cookie if the connection is a TLS connection.

The `secure` flag should also be added if the `site.protocol` configuration is set to `https` and the TLS connection is terminated by a proxy.

mworrell:
### Description

Fix #1889

Set `secure` option for session cookie if the `site.protocol` is `https`

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


ddeboer:
Do we need this in master too?

PR #1891, opened 2018-03-20
decouple mod_translation

onnet:
In conjunction with https://github.com/zotonic/zotonic/pull/1888 mod_translation could be used while running Zotonic without mod_admin+Postgres

### DIFF
--- a/apps/zotonic_mod_translation/include/mod_translation.hrl
+++ b/apps/zotonic_mod_translation/include/mod_translation.hrl
@@ -1,0 +1,8 @@
+
+-record(menu_item,
+        {id,
+         parent,
+         label,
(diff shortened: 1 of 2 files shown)


mworrell:
Is it an option to include the admin menu from the path, but not start `mod_admin`

Am a bit afraid the menu definition might change in the future, which will give problems.

onnet:
@mworrell I believe include admin menu from mod_admin path is also acceptable, at least for now. Addressed your comment.

mworrell:
Thanks, looks good to me.

In the future we might want to split the module between an admin and a non-admin part.  Or move the admin part to mod_admin.

PR #1892, opened 2018-03-20
Zotonic returns 200 instead of 404 on "Not Found"

onnet:
Conversation here: https://groups.google.com/forum/#!topic/zotonic-developers/AlCy9NB2aq4


mworrell:
Thanks!

PR #1900, opened 2018-04-10
mod_search: Fix hassubject/hasobject argument parsing

ddeboer:
### Description

mod_search: Fix hassubject/hasobject argument parsing

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


robvandenbogaard:
I think this fix is all right, because before it `[Id, Predicate]` could only be matched by `parse_edges` that does not expect this list form, it requires wrapping in [] indeed.

PR #1902, opened 2018-04-12
Cotonic

mworrell:
### Description

Integrate Cotonic

This replaces the transport, MQTT and other UI parts of Zotonic with Cotonic.

This is a *major* overhaul

 - [x] Add Cotonic checkout to mod_base
 - [x] Remove old MQTT modules & interfaces
 - [x] Integrate `mqtt_sessions` app
 - [x] Move `z_mqtt_acl` into new mqtt_sessions runtime module
 - [x] Add `render_state` field to `#context{}` replacing the z_script buffers
 - [x] Remove session and page session (will be replace with MQTT sessions)
 - [x] Remove transport code (Javascript and Erlang)
 - [x] Remove postback, web socket and comet controller
 - [x] Move all wire, action and other JS-template code to new module `zotonic_mod_wires`
 - [x] Add MQTT web socket controller
 - [x] Remove / replace `#z_msg_v1{}` handling
 - [x] Integrate models into the topic tree
 - [x] Integrate models into API (replace API controller)
 - [x] Fixes and changes for legacy code (Pubzub, live, etc.)
 - [x] Check old MQTT topics and live topics (templates and erlang code)
 - [x] Handle form POST to topics, pass routing-id (for simple file uploads)
 - [x] Check the access control on model topics and routing topics
 - [x] Events for rsc model (`model/rsc/.../event/...`)
 - [x] Access control check for `m_email_status`
 - [x] Access control check for `m_facebook` graph calls
 - [x] Access control check for `m_survey`
 - [x] Timezone handling (client side timezone, user preferences)
 - [x] Language setting (client side language, user preferences)
 - [x] Propagate timezone / language changes to MQTT connection
 - [x] Add controller to post to topic tree (publish)
 - [x] Need to be able to set persistent id, for anonymous surveys
 - [x] OAuth tests and functionality (mod_oauth)
 - [x] Services: admin
 - [x] Services: zotonic_status - add model entry for status api calls
 - [x] Replace sites update in `zotonic_status` with model subscription
 - [x] Services: check if there are other calls in the js / templates
 - [x] Replace mod_backup download with resource `{% url api zotonic_http_accept="bert" star=["rsc", id] %}` (replacing with `{% url id zotonic_http_accept="application/x-bert" id=id %}`)
 - [x] Drop `persistent` table and references to it
 - [x] error messages on log on (need UI component or css)
 - [x] UI for password reminder and reset (prefer to use UI component)
[...]

PR #1904, opened 2018-04-19
mod_base: Fix DELETE calls on API services

ddeboer:
### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_base/controllers/controller_api.erl
+++ b/modules/mod_base/controllers/controller_api.erl
@@ -29,6 +29,7 @@
     process_post/2,
     is_authorized/2,
     content_types_provided/2,
+    delete_resource/2,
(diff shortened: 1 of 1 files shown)


mworrell:
Why is this an issue?  Someone using DELETE?

ddeboer:
Yup, see the reference to https://github.com/driebit/ginger/pull/380 above.

mworrell:
You can add it.

Note that we are replacing the `controller_api` and deprecating all services in https://github.com/zotonic/zotonic/pull/1902

ddeboer:
Why deprecate the services? Is the recommended course on master to just create custom (Cowboy) controllers?

PR #1906, opened 2018-04-20
remove user_id from google analytics params

### LINKED ISSUE #1804: Anonymize Google Analytics userId property
In #1249 we added the Google Analytics `userId` property. [According to Google](https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id#user_id):

> The USER_ID value should be a unique, persistent, and non-personally identifiable string identifier that represents a user or signed-in account across devices.

However, as we use the person’s resource id, that can be related back to a user (by looking up the id in the admin, for instance). Should we hash the id to make it at least a bit harder to find the user behind it?

DorienD:
Fix #1804

Remove user_id param and update the documentation.

### DIFF
--- a/doc/ref/modules/mod_seo.rst
+++ b/doc/ref/modules/mod_seo.rst
@@ -12,9 +12,18 @@ To enable `Google Universal Analytics`_ tracking on your Zotonic website, go to
 ``http://yoursite/admin/seo`` in your browser and enter your Google Analytics
 tracking code.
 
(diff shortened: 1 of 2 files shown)


mworrell:
Even with documentation 😄 
Thanks!

mworrell:
Manually merged into master https://github.com/zotonic/zotonic/commit/1ffeac7bfb64515e0bc721e09b10032f227ad6ff

Issue #1095, opened 2015-11-26
Unicode for name, slug and page_path


mworrell:
Currently we do transliteration of unicode characters to lowercase ascii.

This was a good solution for 2009. Currently browsers can handle unicode characters in the path if it is percent encoded.

I propose to:
- Map to lowercase
- Allow unicode characters that are not unsafe (think of removing `<>'"`)
- Percent encode remaining characters, but display in the form fields by percent-decoding

The names could be just unicode, with a simple filter to map to an extended 'identifier' form. Think of mapping to lower case and removing ascii characters not in "[a-z0-9_]".

We need to extend the template tokenizer to accept these extra characters in identifiers.

mworrell:
BTW additional option is to make the page_path and slug translatable

mworrell:
Unicode is now supported in te page path and slug.

PR #1911, opened 2018-05-28
Make slug and seo fields translatable

mworrell:
### Description

Issue #1095

Makes the slug and seo fields translatable.

No BC breaks, afaik. Unless the fields are exported in JSON, as the values are now `{trans, ...}` records.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
This fixes only the *slug* part of #1095, not the *page_path* and *name* part.

The *page_path* is a tad more complicate (due to 1:n and also language mapping).
The *name* part needs changes in the tokenizer of the templates, otherwise you can't refer to those names.

PR #1921, opened 2018-06-27
doc: Document user missing from edge notifications

ddeboer:
- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Thanks!

PR #1928, opened 2018-08-04
core: Don't crash when files have been deleted

ddeboer:
### Description

When switching to a Git branch that does not contain some .erl file, the filewatcher still tried to compile that file, crashing with:

`gen_server z_filewatcher_inotify terminated with reason: no match of right hand value {error,enoent} in make:recompilep1/5 line 232`

This change prevents the crash by not compiling deleted files.

Question: is ignoring deletes sufficient or should we add a `code:delete/1` and `code:purge/1` instead?

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/src/filewatcher/z_filewatcher_handler.erl
+++ b/src/filewatcher/z_filewatcher_handler.erl
@@ -120,7 +120,7 @@ handle_file(_Verb, "erlydtl_parser.yrl", ".yrl", F) ->
     os:cmd("erlc -o "++z_utils:os_escape(TargetDir)++" "++z_utils:os_escape(F)),
     "Rebuilding yecc file: " ++ filename:basename(F);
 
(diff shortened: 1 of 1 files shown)


ddeboer:
So do you agree with this @mmzeeman? What would be your answer to my question?

mmzeeman:
It is pretty hard to get this right, depending on the changes. The automatic code reload was not made for this. We usually have to do a complete make clean; and make to get it right.

ddeboer:
Yeah, that’s what we do on production, too. My fix, however, concerns development, mainly. It’s quite annoying to have the inotify watcher crash only because you’re switching between branches that do or do not contain some .erl file.

mworrell:
In master we already ignore the delete event on .erl files.

PR #1931, opened 2018-08-30
Support failure_message in email_unique validator

loetie:
### Description

Currently the failure_message property for this validator is ignored. This adds this message instead of the current message which it now defaults to, "please correct" is the default message. If no failure_message defined then it still uses "please correct".


mworrell:
Also manually merged (with small changes) into the master.

Thanks @loetie!

PR #1937, opened 2018-10-04
mod_admin: Pass content group id when inserting a resource

ddeboer:
### Description

The callbacks check for content_group_id, but this was never passed to
them.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Where does the `content_group_id` in the templates come from?

In the handler code I see that, indeed, the `content_group_id` from the event args is taken. And if that is undefined that then the `content_group_id` from the subject (if any) is taken.

If the `content_group_id` in the templates makes sense, then I am all ok with this change.

PR #1946, opened 2018-11-07
refactor: simpler infix append

teresy:
### Description

Small refactor: it's possible to use a more concise infix append in `filter_format_price.erl`

### Checklist

- [ ] ~documentation updated~
- [ ] ~tests added~
- [x] no BC breaks


mworrell:
Thanks, simpler is always better!

PR #1947, opened 2018-11-07
Remove unnecessary call to internal function

rmies:
Please note that I did not "fix" the indentation: this gives a confusing diff.


mworrell:
Merging.

Note that these routines are refactored in master, so I am not merging into master.

PR #1948, opened 2018-11-07
Exclude node_modules from the file watcher

rmies:


mworrell:
Shouldn't this be suppressed in the blacklist?

https://github.com/zotonic/zotonic/blob/0.x/src/filewatcher/z_filewatcher_handler.erl#L36

In master you might actually not want to do this if the `node_modules` are in `lib-src`. Where do you locate them?

rmies:
Cool, moved it to the blacklist. Don't care about master (for now) since ginger is running on 0.x.

mworrell:
Merged (only for 0.x - please revisit this when using master)

rmies:
Thanks, will do.

PR #1956, opened 2018-12-14
Make acceptable mime types configurable per user group

mworrell:
### Description

Make the acceptable mime types of uploaded files configurable per group.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mmzeeman:
The security headers work fine except for one. One of our clients embeds our sevice into their own website. We need to be able to configure the value for the x-frame-options header. Maybe even dynamically. Going to look for a way to make that possible as the security headers is a zotonic wide setting.

mmzeeman:
Would it be an idea to use a `z_notifier:first/2` in `z_context:set_security_header/1`. That way modules and sites can customise the headers at the expense of an extra notify. The most expensive thing here is probably the ets lookup.

This can also be a way to add, more strict, content-security-policy headers. Most of the default security headers here are basically the standard browser settings.

Maybe like this:

```erlang
set_security_headers(Context = #context{wm_reqdata=ReqData}) ->
    Default = [ {"X-Frame-Options", "sameorigin"},
                {"X-Xss-Protection", "1"},
                {"X-Content-Type-Options", "nosniff"},
                {"X-Permitted-Cross-Domain-Policies", "none"},
                {"Referrer-Policy", "origin-when-cross-origin"} ],

    SecurityHeaders = case z_notifier:first(#security_headers{headers=Default}, Context) of
        undefined -> Default;
        Custom -> Custom
    end,

    RD1 = set_resp_headers(SecurityHeaders, ReqData),

    Context#context{ wm_reqdata = RD1 }.
```

mworrell:
That notifier sounds good to me.
Then we can also always set these headers, and remove the configuration switch.

Shall we use lowercase headers? As that is less confusion and in master we have to use lower case (and binaries) for cowboy.

mworrell:
Added the notification, as suggested by @mmzeeman.

We should check when we add these headers (and the no cache headers), they might need to be set earlier in the request. Though I think they are especially critical for dynamic pages, which always pass this point in the code.

mworrell:
Moved setting the security-headers to context initialization, so all controllers get the same set of headers.

mmzeeman:
Couldn't callback and language be escaped when they are taken from the user?

https://github.com/zotonic/zotonic/blob/configurable-mime-types/modules/mod_admin/templates/_action_dialog_connect.tpl#L21

```
callback|default:q.callback|default:"window.zAdminConnectDone",
language|default:q.language|default:z_language,
```

mworrell:
Those values are only used in conjunction with signed callbacks, so to abuse them you need to have full (javascript) access to the page already.  Out of abundance I added a sanity check on the callback name and an escape on the `q.language` value.

mworrell:
@ddeboer Added some documentation.  If it is green we can merge.

PR #1963, opened 2019-01-02
WIP: Add support for dmy formatted dates when updating models

linuss:


ddeboer:
Our use case here is to show dates in the admin in a localised format (7-1-2019 instead of 2019-01-07), and this change seemed necessary.

mworrell:
I propose to restructure this.  Scheduling it preemptively for  the 0.48.

Agreeing with to re-check the date-picker options.

mworrell:
Moving to 0.49, as we have still pending reviews.

mworrell:
@ddeboer any movement on this?

mworrell:
Checking if there is another solution to this.

mworrell:
Superseded by #2098

PR #1969, opened 2019-01-09
Move mnesia files to a directory per node.

mworrell:
### Description

Place all mnesia files in a directory per node.

Instead of `priv/mnesia` we will now use something like `priv/mnesia/zotonic001@localhost`

On first creation any mnesia files in priv/mnesia will be moved to the new directory.

This solves a problem where a Zotonic node can't be started after the node name changed or if mnesia files from another node are placed in the `priv/mnesia` directory.

This is also compatible with the changes in master.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/support/z_sites_manager.erl
+++ b/src/support/z_sites_manager.erl
@@ -47,7 +47,10 @@
     await_startup/1,
 
     get_site_config_overrides/1,
-    put_site_config_overrides/2
+    put_site_config_overrides/2,
+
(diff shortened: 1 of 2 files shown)


mworrell:
@ddeboer can you check what goes wrong in the Docker tests>

ddeboer:
Looks like a temporary hiccup; works now, so merging.

PR #1973, opened 2019-01-14
doc: Document BC break in password reset

ddeboer:
### Description

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
The logon form is also changed, basically we route these postbacks now to `controller_logon` and the event handlers in `mod_authentication` are removed.

So we should also mention: 

 * `_logon_login_form.tpl`
 * `_logon_expired_form.tpl`
 * `_logon_reminder_form.tpl`

mworrell:
Added in 6de4e4a

PR #1977, opened 2019-01-23
Badmatch in expired logon on get_username value

loetie:
m_identity:get_username does not return {ok, Username} but only Username itself or undefined. Currently gives an badmatch error since it cannot be matched on {ok, Username}

### Description

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Will merge this, needs some additional logic though (there must be an username).

PR #1978, opened 2019-01-23
core: Fix {expired, _} tuple not handled correctly

ddeboer:
### Description

controller_logon handles an {expired, UserId} case, but it was not possible to return that from e.g. observe_identity_password_match/2.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/models/m_identity.erl
+++ b/src/models/m_identity.erl
@@ -354,7 +354,7 @@ check_username_pw(Username, Password, Context) ->
                         },
                         Context),
                     {ok, RscId};
-                {error, _} = Error ->
+                Error ->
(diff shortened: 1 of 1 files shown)


ddeboer:
Ping @mworrell

mworrell:
Isn't it better to make this `{error, {expired, ...}}` ?

mworrell:
Better to have some consistent format.

mworrell:
Changed this to merge into the 0.45 branch (which we will merge back into 0.x)

mworrell:
Merging this, will change that tuple afterwards.

PR #1989, opened 2019-02-05
Display unique name

DorienD:
This will display the unique name on the dialog. For quicker access to the unique name. 

I think this is a tiny change with a big impact for people using search queries and need the unique name.


mworrell:
You want to force an unique name on all resources in the `meta` category?

So also for keywords and the like?

mworrell:
@DorienD maybe only have the validation for `predicate` and `category`?  And the show the field as an optional field for the other `meta` other categories.

DorienD:
What do you mean by showing it as an optional field? Like just showing the field without being able to edit it?

mworrell:
"optional" as in: visible and editable for all _meta_, but not required.

mworrell:
See #2097 which makes the validation dependent on the category of the resource being edited.

PR #1991, opened 2019-02-06
mod_bootstrap: Upgrade to 3.4.0

ddeboer:
### Description

Upgrade Bootstrap to 3.4.0, which fixed https://nvd.nist.gov/vuln/detail/CVE-2018-14041.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
I assume this is just the stock Bootstrap, in the same configuration as we used before?

ddeboer:
What do you mean by ‘same configuration’? I just downloaded Bootstrap and replaced the files.

mworrell:
Ah, I though there was some config file being used.

@ArthurClemens do you know if we can just use the "vanilla" bootstrap?

PR #1993, opened 2019-02-06
core: Add cors_headers notification

ddeboer:
### Description

Keep BC, but allow CORS to be disabled in controller_file.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Just wondering.  Maybe it is good to do something for every request?  Pass the controller and dispatch rule, then let the observer decide if it wants to change the header.

In the master we could add this to the middleware that sits between the dispatcher and cowmachine.

We could also make this part of the 'security headers' notification?  (iff we pass the dispatch and controller information).

Another option is to add it to the dispatch rule.

(Just thinking aloud).

ddeboer:
> Pass the controller and dispatch rule, then let the observer decide if it wants to change the header.

I thought about that as well, but decided to have the controller itself (which should know best) provide a default. That’s also the reason I didn’t do this:

> We could also make this part of the 'security headers' notification?

as you may want to override just the security headers but have but still apply the controller-provided default for the CORS ones.

ddeboer:
@mworrell Can we merge this?

mworrell:
@ddeboer Can you check the merge conflicts?

ddeboer:
Resolved.

PR #1998, opened 2019-02-11
mod_authentication: add config to enable reminder email to unregister…

mworrell:
### Description

If a reminder email is requested for an unknown email address,  then do not send the email.

This behavior can be changed by setting the option `mod_authentication.email_reminder_if_nomatch` to `1`.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


ddeboer:
Yes, this should be off by default.

What about renaming ‘reminder’ to ‘password reset’? We don’t actually ‘remind’ users of anything, do we? (Like sending their plain text password in the e-mail, as it was done in the nineties. :p)

PR #2000, opened 2019-02-12
[mod_admin] Don't let flush button in admin interface flush all sites

### LINKED ISSUE #1995: Flush cache admin action is flushing all sites, is that intended and desirable?
Pushing the Flush cache button in the admin interface of a specific site flushes caches of all sites, because the corresponding admin action uses `flush/0` instead of `flush/1` with site parameter. I would say that this is unexpected from the admin user's view. Other sites are unknowingly impacted while the intention is to flush caches only of the site at hand.

https://github.com/zotonic/zotonic/blob/012319f62d9c67f67bd95a97d482778d84ceeaf8/apps/zotonic_mod_admin/src/actions/action_admin_admin_tasks.erl#L24

robvandenbogaard:
### Description

Fix #1995 

Preserve current functionality in admin_task action `flush_all_sites` and edit task `flush` to pass the context to the flush call, causing it to only flush the site at hand.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_admin/actions/action_admin_admin_tasks.erl
+++ b/modules/mod_admin/actions/action_admin_admin_tasks.erl
@@ -21,7 +21,7 @@ render_action(TriggerId, TargetId, Args, Context) ->
 
 %% @doc Flush the caches of all sites.
 event(#postback{message={admin_tasks, [{task, "flush_all_sites"}]}}, Context) ->
-    do(fun z:flush/0, "Caches have been flushed.", Context);
+    do(fun z:flush/0, "All sites have been flushed.", Context);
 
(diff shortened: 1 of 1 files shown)


mworrell:
Shouldn't we add a button "flush all sites"  (only for admin user nr 1)

robvandenbogaard:
I would rather add that to the interface where all sites are listed. So you
see which ones would be impacted.

Op di 12 feb. 2019 16:27 schreef Marc Worrell <notifications@github.com:

mworrell:
Merging, though we might want to add that other button to the `zotonic_status` site,

PR #2007, opened 2019-02-20
doc: Explain anonymous context in task queue callback

ddeboer:
### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Also merged into master.

PR #2008, opened 2019-02-22
core: Default publication_start to null

ddeboer:
### Description

When creating an unpublished resource, publication_start defaults to the
date of creation. This seems wrong: publication_start should be empty
until the resource is published.

Only set publication_start if it is null and is_published is true.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Good idea.

I think that we should do the business logic of setting the pubdate_start (iff it is null and is_published is set to true) in the transaction logic, just before the database update query.

Now it is dependent on the form submit, but this should also work with programmatically setting the is_published flag.

src/models/tests/m_rsc_db_tests.erl
```diff
@@ -28,13 +28,15 @@ modify_rsc_test() ->
     ?assertEqual(true, m_rsc:is_a(Id, text, AdminC)),
 
     ?assertEqual(false, m_rsc:p(Id, is_published, AdminC)),
+    ?assertEqual(undefined, m_rsc:p(Id, publication_start, AdminC)),
```
ddeboer: @mworrell Any idea why this test fails? It’s green locally.

ddeboer:
> I think that we should do the business logic of setting the pubdate_start (iff it is null and is_published is set to true) in the transaction logic, just before the database update query.

Done.

src/models/tests/m_rsc_db_tests.erl
mworrell: I've changed the initial `rsc` definition in `z_install.erl`. Awaiting Travis now.

src/install/z_installer.erl
```diff
@@ -563,11 +576,15 @@ convert_category_hierarchy(C, Database, Schema) ->
...
+    case is_column_nullable(C, "rsc", "publication_start", Database, Schema) of
+        true -> ok;
+        false ->
+            {ok, [], []} = pgsql:squery(
+                C,
+                "ALTER TABLE rsc "
+                    "ALTER COLUMN publication_start DROP NOT NULL, "
+                    "ALTER COLUMN publication_start DROP DEFAULT"
+            ),
+            ok
```
ddeboer: @mworrell Why did you change this; just to perform fewer `ALTER` operations on the database?

src/install/z_install.erl
```diff
@@ -127,7 +127,7 @@ model_pgsql() ->
       is_featured boolean NOT NULL DEFAULT false,
       is_protected boolean NOT NULL DEFAULT false,
       is_dependent boolean NOT NULL DEFAULT false,
-      publication_start timestamp with time zone NOT NULL DEFAULT now(),
+      publication_start timestamp with time zone,
```
ddeboer: So this is what fixed the test, right?

mworrell: Yes

src/install/z_installer.erl
mworrell: yes, otherwise each startup we do those "alter" statements.

PR #2010, opened 2019-02-27
Fix #2009: Add lists to plugins to support lists in editor

### LINKED ISSUE #2009: Tinymce bullist en numlist don't work anymore
Since the release of the new tinymce editor we need a special plugin for the numlist en bullist, lists. This has to be added to the tiny-init.js of the new version.

DorienD:
### Description

Fix #2009

Add tinymce plugin `lists` to support numlist and bullist to new tinymce editor 4.9.3

### DIFF
--- a/modules/mod_editor_tinymce/lib/js/tinymce-4.9.3/tiny-init.js
+++ b/modules/mod_editor_tinymce/lib/js/tinymce-4.9.3/tiny-init.js
@@ -3,7 +3,7 @@ if (typeof tinyInit !== 'object')
     selector: "textarea",
     mode: "none",
     theme: "modern",
-    plugins: "compat3x advlist code paste table link zlink zmedia autosave directionality autoresize",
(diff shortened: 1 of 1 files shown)


DorienD:
@mworrell @ddeboer could we add this to a 0.46.1 release?

mworrell:
Can you (temporarily) switch to an older tinyMCE version?  Then you don't need to wait for a release.

mworrell:
@ddeboer Do we want to make a patch release?  The 0.47 is planned for coming Monday.

ddeboer:
Nope, let’s just stick to 0.47.

mworrell:
ok, closing this issue.

@DorienD thanks for the fix!

DorienD:
When will 0.47 be tagged? Since all our production sites will be affected if we release 0.46.

ddeboer:
https://github.com/zotonic/zotonic/milestones

PR #2011, opened 2019-02-28
core: Edge insert top

mworrell:
### Description

Add an option to insert new edges _before_ existing edges.

Normally edges were inserted after all existing edges.
This option makes it possible to invert that behavior.

Two places:

 - `is_insert_before` property on the predicate
 - `is_insert_before` option to `m_edge:insert/4`

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/models/m_edge.erl
+++ b/src/models/m_edge.erl
@@ -62,11 +62,18 @@
 
 -include_lib("zotonic.hrl").
 
--type insert_opts() :: is_insert_before
-                     | no_touch
-                     | {seq, integer()}
-                     | {creator_id, m_rsc:resource_id()}
-                     | {created, calendar:datetime()}.
+-type insert_options() :: [ insert_option() ].
+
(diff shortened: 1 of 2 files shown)


src/models/m_edge.erl
```diff
@@ -62,6 +62,11 @@
 
 -include_lib("zotonic.hrl").
 
+-type insert_opts() :: is_insert_before
+                     | no_touch
+                     | {seq, integer()}
+                     | {creator_id, m_rsc:resource_id()}
+                     | {created, calendar:datetime()}.
```
ddeboer: This type definition looks incorrect, because actually `insert_opts()` is a list. So did you mean `[insert_option()]` instead?

mworrell: Good catch. Fixed it and also renamed it to `insert_options()`

PR #2012, opened 2019-03-01
Let user re-agree to T&C.  Restructure logon errors.

mworrell:
### Description

Add option to force the user to re-agree with the T&C after authentication.
This is regulated by the config `mod_authentication.tos_update_agree`
The user's property `tos_agreed` (or the user's creation date) is checked against the publication dates of the resources `signup_tos` and `signup_privacy`.
If either publication date is newer than the user's agree date then the user is forced to agree with the T&C after authentication (and before actual logon).

This also restructures the error messages and actions for the logon templates.
There is a separate template for every error message, so that it is easier to change these without resorting to copying the authentication dialogs.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_authentication/templates/_logon_error.tpl
+++ b/modules/mod_authentication/templates/_logon_error.tpl
@@ -1,62 +1,32 @@
+
+{###################################}
+{# Show / hide logon form elements #}
+{###################################}
+
+{% wire action={unmask target=" form"} %}
+
 {% if reason == "pw" %}
-    <p>
-        {_ Either the email or the password you entered is incorrect. Please check your entry and try again. _}
-    </p>
-
-    <p><a href="{% url logon_reminder %}" id="logon_error_link_reminder">{_ Need help signing in? _}</a></p>
-
+    {% wire action={focus target="username"} %}
 {% elseif reason == "need_passcode" %}
-
-    <p>
-        {_ Please enter the two-factor authentication passcode. _}
-    </p>
-
+    {% wire action={focus target="passcode"} %}
 {% elseif reason == "passcode" %}
-
-    <p>
-        {_ The two-factor authentication passcode did not match. Please check your entry and try again. _}
-    </p>
-
-    <p><a href="{% url logon_reminder %}" id="logon_error_link_reminder">{_ Need help signing in? _}</a></p>
-
+    {% wire action={focus target="passcode"} %}
 {% elseif reason == "ratelimit" %}
-
-    <p>
-        {_ Too many retries. _}
-        {_ Please try again in _}
-        {% with m.ratelimit.timeout as seconds %}
-            {% if seconds == 3600 %}{_ an hour _}.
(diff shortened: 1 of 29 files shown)


modules/mod_authentication/templates/_logon_error.tpl
```diff
@@ -1,62 +1,32 @@
...
+    {% wire action={focus target="username"} %}
+{% endif %}
 
-    <p>{_ The two passwords should be equal. Please retype them. _}</p>
-
-    <p>{_ Passwords should have at least six characters. _}<p>
-    <p>{_ Use some non alphabetical characters or digits to make it harder to guess. _}</p>
-
-{% elseif reason %}
+{% catinclude "logon_error/action.tpl" [reason|as_atom] %}
```
ddeboer: Doesn't the `catinclude` need the resource id as an argument?

mworrell: Yes, it actually also accepts a list of atoms.

In the master I want to let it also accept a list of binaries. Not in the 0.x as with erlydtl it is a bit confusing that some strings are lists and some strings are binaries.

mworrell:
Shall we add this to the 0.47? Or schedule it for the 0.48?

modules/mod_authentication/templates/logon_error/message.unequal.tpl
```diff
@@ -0,0 +1,4 @@
+    <p>{_ The two passwords should be equal. Please retype them. _}</p>
+
+    <p>{_ Passwords should have at least six characters. _}<p>
```
ddeboer: Doesn’t this go for `tooshort` only?

modules/mod_authentication/templates/logon_error/message.reminder.tpl
```diff
@@ -0,0 +1,4 @@
+    <p>{_ You've entered an unknown username or email address. Please try again. _}</p>
```
ddeboer: Do we still show this message after #1910?

mworrell: Actually not, the error is now a nondescript `pw`, so let's remove this message.

modules/mod_authentication/templates/logon_error/message.unequal.tpl
mworrell: It is a bit silly that this error is hard-coded this way, especially as the length is a configuration.
What do you propose?  Echo the minimum length here?

mworrell: Maybe just remove it and force the minimum length using a validator.

mworrell: There is already a validator:

```
            type={
                length minimum=min_length
                too_short_message=_"Your password is too short." ++ " " ++ _"Minimum characters: " ++ min_length
            }
```

So the whole message length string is not needed.  I'll remove it.

PR #2014, opened 2019-03-01
core: Prevent too many WebSockets

### LINKED ISSUE #2013: Push connection problems
This week we have run into some strange push connection problems. 

We first got a warning that the server was having > 100k open tcp connections. When I looked I found there where also a lot of sessions and page sessions.

When investigating a little bit deeper I found sessions which had over 3000 pages connected to it.

After enabeling the access log I saw that these where caused by newer safari browsers which tried to open a lot of websocket connections at once. I did a small cleanup of the server side websocket handling code which cleans up unused websockets a little bit better. (Added some code to automatically close sockets when an one ip address had > 300 connections open).

Then later I could reproduce the problem. When I opened the lid of my macbook it started to create websockets like crazy, and it would leave them open. The server created a new session and a new page process for each websocket. The sockets where never closed.

No I have changed zotonic.js a bit so it will not create a new websocket when it already has one, and it looks like the problem has gone away. (With some peculiarities. Now I get a lot of comet requests, but that is not so bad for the server).

mmzeeman:
### Description

Fix #2013

### Checklist

- [ ] documentation updated
- [ ] tests added
- [X] no BC breaks


mmzeeman:
Small problem... the sort filter won't sort.

mmzeeman:
Ok... the sort filter didn't work for proplists, so I worked around it to get the ordering right.

PR #2018, opened 2019-03-08
Add block around modal footer

DorienD:
### Description

Add block around modal footer in dialog edit basics, so there is less need to overwrite the whole template.


mworrell:
Just wondering, what kind of buttons are you using?

PR #2021, opened 2019-03-13
mod_bootstrap: Upgrade to 3.4.1 and add a upgrade description. Fixes …

### LINKED ISSUE #2020: Upgrade of bootstrap caused reappearance of #939
Last month bootstrap was upgraded, this caused the reappearance of #939.

There is a script hidden in the priv directory of `mod_bootstrap`.

We should probably add a note somewhere that we need to run this script when we upgrade bootstrap

mmzeeman:
### Description

Fix #2020

- Upgrades bootstrap to 3.4.1 by using the update script.
- Added a readme so that we not forget to use the update script in the future.

### Checklist

- [X] documentation updated
- [ ] tests added
- [X] no BC breaks


mworrell:
Great, I love the readme, we should use that for more modules.

PR #2034, opened 2019-04-02
Redesign of connect/new tab.

### LINKED ISSUE #2027: New/find dialog change requests
Change requests:

- [ ] Make it more clear that the preview is an overlay
- [ ] Click on the list item in the suggestions will result in an edge in stead of preview
- [ ] Change on category will result in category filter in suggestions on the right
- [ ] Add a helper text above the search result
- [ ] We will make a sketch for a design proposal

mworrell:
### Description

Fix #2027 

Redesign of the new connect/new tab.

 * Opens preview of items over the "new" form
 * Remove filter "valid for <predicate>", category id from form is now leading
 * Helper text above search result

To do:

 - [x] Hide & disable upload if selected category is not a media category
 - [x] Config key (?) to default to 'own content' in the search result
 - [x] Explanation headers on right column and preview panel
 - [x] Rename preview panel template (as it is only part of the dialog)
 - [x] Add translations (admin .pot / .po)

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/modules/mod_acl_user_groups/templates/_admin_catcg.tpl
+++ b/modules/mod_acl_user_groups/templates/_admin_catcg.tpl
@@ -2,7 +2,7 @@
     <label class="control-label col-md-3" for="{{ #category }}">{_ Category _}</label>
     <div class="col-md-9">
         {% if is_nocatselect %}
-            <input class="form-control" type="text" readonly value="{{ cat_id.title }}" />
+            <input class="form-control nosubmit" type="text" readonly value="{{ cat_id.title }}" />
             <input type="hidden" id="{{ #catsel }}" name="category_id" value="{{ cat_id }}"/>
         {% else %}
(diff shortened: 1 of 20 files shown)


mworrell:
Ping @DorienD

DorienD:
Would it be possible to also keep the 'old' new tab? So we can choose per site which one we use? So default this new_connect tab, and second the new page tab.

mworrell:
I am a bit careful with too many variations of the same content-creation forms. You could think of configuration options, but then you will also need to modify the text and logic of the tabs.

DorienD:
Why would the text change in the tabs?

mworrell:
Because then it should only be "Create" and not "Create or find"

DorienD:
What i propose is keeping this new one as the standard and let new still available. The new dialog could keep the old texts, since it is a view on its own?

PR #2041, opened 2019-04-11
mod_admin: Upgrade Bootstrap

### LINKED ISSUE #2033: mod_admin uses a separate Bootstrap version
Bootstrap was recently upgraded (#2021, #1991) but mod_admin has its own, outdated, version in `admin-bootstrap3/js/bootstrap.min.js`. @mworrell Is this on purpose? I suggest we make mod_admin depend on mod_bootstrap and use that version instead.

ddeboer:
### Description

* Remove separate Bootstrap instance in mod_admin and use the one from
  mod_bootstrap for easier maintainability.
* This fixes a security report about mod_admin containing an outdated
  and insecure Bootstrap version.

Fix #2033.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
This give a problem with existing sites that overruled the bootstrap module with their own design (and kept using the javascript from mod_bootstrap).

Would it be possible to have the `mod_bootstrap` css and js available on a separate admin dispatch path?  Either by adding dispatch rules or by having some (soft) links?  (Am not sure about the links because of Windows)

ddeboer:
Isn’t it actually a feature that the admin now resembles the site more closely if CSS for the latter’s Bootstrap has been changed?

Do you mean a dispatch rule that points to the _default_ mod_bootstrap, unmodified by the user’s changes?

mworrell:
Yes, something like that, or at least that the bootstrap is available on two paths, and that you can then overrule the one or the other.

mworrell:
I am not so sure about overruling the admin, as the site might use a subset of bootstrap that is not compatible with the admin.

(Though I agree that having the admin look like the site is a nice feature)

mworrell:
Of course we could decide that _if_ the site includes its own bootstrap _then_ it should either contain everything that the admin needs _or_ should be placed on a different path.

/cc @ArthurClemens @arjan What do you guys think?

ddeboer:
That dependency is actually unnecessary, as we only use the templates from mod_bootstrap, so it doesn’t need to be active.

mworrell:
Don’t we use the css and js from mod_bootstrap?

ddeboer:
Yes, we do. But that still doesn’t require the module to be active, does it? 

Anyway, I’m okay with the dependency as mod_bootstrap itself is an empty module, and it makes it a bit clearer that we use stuff from mod_bootstrap in mod_admin.

mworrell:
Moving this back to 0.49, I agree with the change though.  We just need some documentation.

doc/developer-guide/releasenotes/rel_0.49.0.rst
```diff
@@ -0,0 +1,11 @@
+.. _rel-0.49.0:
+
+Release 0.49.0
+==============
+
+Welcome to Zotonic 0.49.0, released on TODO.
+
+Main changes are:
+
+* Upgraded Bootstrap styling in mod_admin to 3.4.1; mod_admin no longer ships
```
ddeboer: Nog just styling but also JS.

modules/mod_admin/mod_admin.erl
```diff
@@ -24,6 +24,7 @@
 -mod_description("Provides administrative interface for editing pages, media, users etc.").
 -mod_depends([base, authentication, mod_search, mod_mqtt]).
 -mod_provides([admin]).
+-mod_depends([bootstrap]).
```
ddeboer: Does this mean users will have to add `mod_bootstrap` to their `config`?

doc/developer-guide/releasenotes/rel_0.49.0.rst
```diff
@@ -0,0 +1,13 @@
...
+==============
+
+Welcome to Zotonic 0.49.0, released on TODO.
+
+Main changes are:
+
+* **BC break:** Upgraded Bootstrap CSS and JavaScript in mod_admin to 3.4.1;
+  mod_admin no longer ships with a separate Bootstrap instance but uses the one
+  from mod_bootstrap. Any customizations you made to the Bootstrap CSS will
+  therefore be visible in the admin, too.
```
ddeboer: Expanded the release notes. @mworrell Is this clear enough?

modules/mod_admin/mod_admin.erl
mworrell: No, just any module that provides `bootstrap`, which might also be their own module.

mworrell:
Am ok with this change, it removes code and makes it easier to adapt the admin.

mworrell:
Done! :-)

PR #2042, opened 2019-04-11
WIP: Add audit log module

### LINKED ISSUE #1529: Audit module
Continued from https://github.com/zotonic/zotonic/issues/1182#issuecomment-261201253.

> What about the amount of time the actions need to be stored? Two months seems reasonable.

Why not forever? In the [event sourcing](http://martinfowler.com/eaaDev/EventSourcing.html) way of building applications, every change is captured and persisted. The idea is to store the events forever to keep a full history of all changes. Why not do the same?

ddeboer:
### Description

Fix #1529.

### Checklist

- [ ] Add schema for `log_audit` table.
- [ ] Update admin templates to show and filter on all `#log_audit{}` properties, similar to the e-mail log.
- [ ] documentation updated
- [x] no BC breaks

### DIFF
--- a/modules/mod_audit/mod_audit.erl
+++ b/modules/mod_audit/mod_audit.erl
@@ -1,0 +1,136 @@
+%% Copyright 2019 Zotonic
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+-module(mod_audit).
+-author("David de Boer <david@ddeboer.nl>").
+
+-mod_title("Audit log").
+-mod_description("Audit log of content and admin changes").
+-mod_depends([mod_logging]).
+-mod_prio(400).
+
+-export([
+    observe_m_config_update/2,
+    observe_module_activate/2,
+    observe_module_deactivate/2,
+    observe_auth_checked/2,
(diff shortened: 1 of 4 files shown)


modules/mod_audit/mod_audit.erl
```diff
@@ -0,0 +1,136 @@
...
+
+observe_m_config_update(#m_config_update{module = Module, key = Key, value = Value}, Context) ->
+    log(
+        #log_audit{
+            action = config_update,
+            previous_value = m_config:get(Module, Key),
+            value = Value
+        },
+        Context
+    ).
```
ddeboer: All these observers are quite verbose, but I like that it’s very explicit.

modules/mod_audit/mod_audit.erl
```diff
@@ -0,0 +1,136 @@
...
+-export([
+    observe_m_config_update/2,
+    observe_module_activate/2,
+    observe_module_deactivate/2,
+    observe_auth_checked/2,
+    observe_auth_logoff/3,
+    observe_edge_insert/2,
+    observe_edge_delete/2,
+    observe_media_update_done/2,
+    observe_rsc_update_done/2
```
ddeboer: Any other notifications you’d like to see in the audit log?

mworrell:
Maybe we could hook into (the results of) `mod_ratelimit` and `mod_clamav` ?

ddeboer:
@mworrell Not yet, because they publish no notifications. 😉

mworrell:
It might be an idea to be able to hook into the `z_notifier` to observe the results of `first` or `map` notifications? Wondering what the most efficient method could be.

mmzeeman:
In the module we made it is possible to have an hierarchy of events so you can easily filter on authentication events, resource modification, etc. In our own codebase we have custom events.

What I don't have right now is an admin interface.

mworrell:
See also https://github.com/zotonic/zotonic/pull/2057 for some simple logging of system-events using the current logging module.

mworrell:
This is for an older 0.x - we have now moved to 1.x

I think this could also be done using a Logger module.

PR #2054, opened 2019-05-02
Additional virus scanning for spreadsheet files

mworrell:
### Description

Additional virus scanning for spreadsheet files.

 - [x] Filter for `externalLinks` in MS-Office files
 - [x] Sanitize uploaded CSV files
 - [x] Add translations

### Checklist

- [x] documentation updated
- [x] tests added
- [x] no BC breaks


doc/ref/modules/mod_clamav.rst
```diff
@@ -24,6 +24,11 @@ The following configs are available:
 ``clamav_max_size``
     The **StreamMaxLength** of ``clamd``, default to ``26214400`` (25M)
 
+``clamav_reject_msoffice_external_links``
+    Reject Microsoft Office documents containing ``externalLinks`` information.
+    If the Zotonic config is set to ``false`` then rejection can be forced by setting
+    the site’s config key ``mod_clamav.reject_msoffice_external_links`` to ``1``.
```
ddeboer: Missing default

modules/mod_clamav/support/z_clamav_msoffice.erl
```diff
@@ -0,0 +1,48 @@
...
+%% limitations under the License.
+
+-module(z_clamav_msoffice).
+
+-export([
+    scan_file/3
+]).
+
+scan_file(File, <<"application/vnd.openxmlformats-officedocument.", _/binary>>, Context) ->
+    case {z_config:get(clamav_reject_msoffice_external_links, true),
```
ddeboer: Should this not be converted to bool?

mworrell: It is a `zotonic.config` setting, question if we should be strict in typing there or not. From the site-config I usually am converting, as that is more-or-less user-input.

PR #2055, opened 2019-05-02
Add password reset token expiration

### LINKED ISSUE #1662: Expire reset password link
For security reasons, make it expire after 48 hours (by default), based on a configuration parameter.

ddeboer:
### Description

Fix #1662.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/modules/mod_authentication/controllers/controller_logon.erl
+++ b/modules/mod_authentication/controllers/controller_logon.erl
@@ -40,7 +40,7 @@
 
 -define(LOGON_REMEMBERME_COOKIE, "z_logon").
 -define(LOGON_REMEMBERME_DAYS, 365).
-
+-define(RESET_TOKEN_MAXAGE, 48*3600).
 
 init(DispatchArgs) -> {ok, DispatchArgs}.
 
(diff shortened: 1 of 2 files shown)


modules/mod_authentication/controllers/controller_logon.erl
```diff
@@ -690,7 +690,14 @@ delete_reminder_secret(Id, Context) ->
 
 
 get_by_reminder_secret(Code, Context) ->
-    case m_identity:lookup_by_type_and_key("logon_reminder_secret", Code, Context) of
+    Result = case m_config:get_value(mod_authentication, reset_token_maxage, Context) of
+        undefined ->
+            m_identity:lookup_by_type_and_key("logon_reminder_secret", Code, Context);
+        MaxAge ->
+            m_identity:lookup_by_type_and_key("logon_reminder_secret", Code, z_convert:to_integer(MaxAge), Context)
```
ddeboer: Should `MaxAge` also be shown in `email_password_reset.tpl`?

mworrell: Not needed in the email, we should just mention that it can be used for a short period of time.

mworrell:
Wondering if, instead of adding an extra function, we shouldn't calculate the `MaxAge` and check against the `created` date in the controller itself?

mworrell:
Changed this so that:

 * MaxAge is checked in `controller_logon`
 * Default MaxAge to 2 days
 * Check against `modified` of identity record
 * Update `modified` of identities when updating them
 * Delete *reminder secret* identity if check shows it is expired

PR #2056, opened 2019-05-03
Add config option to force new password on password reset

mworrell:
### Description

Adds config option `site.password_force_different` to force a different password on password resets.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


ddeboer:
Fixed spelling a bit, merging.

PR #2057, opened 2019-05-03
Logging of system/acl events in mod_logging

mworrell:
### Description

This re-uses mod_logging to log more system related events:

 - [x] Sudo as another user
 - [x] Config changes
 - [x] Rate limiter on auth or reset kicking in
 - [x] Error during file clamav checks on files
 - [x] User changes
 - [x] User role changes

### Known Issues

 * We don't know the user who removed a user-group member (as this is signaled via the edge update log, which is filled via a trigger)
 * Message if a MS-Office file is refused due to linked files. This can be added as a log entry after merging, as this logging branch does not have the office-filtering code yet.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/models/m_config.erl
+++ b/src/models/m_config.erl
@@ -188,7 +188,8 @@ set_value(Module, Key, Value0, Context) ->
                 "Configuration key '~s.~s' inserted, new value: '~s'",
                 [ z_convert:to_binary(Module), z_convert:to_binary(Key), Value ],
                 [ {module, ?MODULE}, {line, ?LINE} ],
-                Context);
+                Context),
+            ok;
         {update, OldV} ->
             z_depcache:flush(config, Context),
             z_notifier:notify(#m_config_update{module=Module, key=Key, value=Value}, Context),
@@ -196,7 +197,8 @@ set_value(Module, Key, Value0, Context) ->
                 "Configuration key '~s.~s' changed, new value: '~s', old value '~s'",
                 [ z_convert:to_binary(Module), z_convert:to_binary(Key), Value, OldV ],
                 [ {module, ?MODULE}, {line, ?LINE} ],
-                Context);
+                Context),
+            ok;
         {rollback,{no_database_connection, _Trace}} ->
             {error, no_database_connection};
         {rollback, {error, _} = Error} ->
--- a/src/models/m_config.erl
+++ b/src/models/m_config.erl
@@ -197,7 +197,7 @@ set_value(Module, Key, Value0, Context) ->
                 [ z_convert:to_binary(Module), z_convert:to_binary(Key), Value, OldV ],
                 [ {module, ?MODULE}, {line, ?LINE} ],
                 Context);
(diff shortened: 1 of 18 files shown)


mworrell:
@ddeboer Can you give this a test run?

src/models/m_config.erl
```diff
@@ -176,6 +225,11 @@ delete(Module, Key, Context) ->
     z_db:q("delete from config where module = $1 and key = $2", [Module, Key], Context),
     z_depcache:flush(config, Context),
     z_notifier:notify(#m_config_update{module=Module, key=Key, value=undefined}, Context),
+    z:warning(
```
ddeboer: Why `warning`? This should be `info` as well, as `warning` indicates something is wrong.

src/support/z_module_manager.erl
```diff
@@ -152,6 +160,12 @@ activate(Module, IsSync, Context) ->
                         end
                 end,
             1 = z_db:transaction(F, Context),
+            UId = z_acl:user(Context),
+            z:warning(
```
ddeboer: Again, these should not be warnings IMHO.

src/models/m_config.erl
mworrell: I was thinking that it is a config change, and more important than the other events. Same for modules.

src/support/z_module_manager.erl
mworrell: See also above, not really a "warning" as such, more an high priority informational message.

src/models/m_config.erl
```diff
@@ -191,7 +191,9 @@ set_value(Module, Key, Value0, Context) ->
                 "Configuration key '~s.~s' changed, new value: '~s', old value '~s'",
                 [ z_convert:to_binary(Module), z_convert:to_binary(Key), Value, OldV ],
                 [ {module, ?MODULE}, {line, ?LINE} ],
-                Context)
+                Context);
+        {rollback, _} ->
+            ok
     end,
```
mworrell: If we have a rollback - shouldn't we then retry?

ddeboer:
Just pushed a fix for the status page, which breaks at `z_db:transaction/2` because it has no database connection.

mworrell:
Which config was set by the status?  We could also check on having a db connection, and if not then return an error code from the config-set

ddeboer:
```
{rollback,{no_database_connection,[{m_config,set_value,4,[{file,"/Users/david/src/zotonic/src/models/m_config.erl"},{line,181}]},{mod_ratelimit,generate_device_secret, 
…
```

Checking db conn is probably better; feel free to change!

src/models/m_config.erl
mworrell: Made it a bit more restrictive and added ok/error return values to the function.

src/models/m_config.erl
mworrell: info it is.

PR #2077, opened 2019-05-16
Only allow admin to change admin 2fa settings

mworrell:
### Description

Only allow the admin (user 1) to change the 2FA settings of the admin.

 - [ ] Add translation

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@ddeboer maybe we should only allow people to set the 2FA TOTP code for themselves?  And not have anybody else be able to generate the new code?  That makes sense to me.

mworrell:
Only the user will be able to set their 2FA barcode.

PR #2080, opened 2019-05-20
Fix identification of files with a space in the name

mworrell:
### Description

This fixes an issue where ImageMagick identify result parsing would fail if the input filename had a space in it.

This only happens if identify is called directly with a file that is not uploaded via Zotonic, as Zotonic ensures filenames do not have spaces in them.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


ddeboer:
Spaced files, nice edge case. 😀

PR #2086, opened 2019-05-21
Log client ui errors

mworrell:
### Description

This adds a tab in the `admin/log` where you can see client-side Javascript errors.

The errors are caught in an `onerror` handler and sent with a xhr request to a custom controller in `mod_logging`. The error is then inserted in the ui-log database table and in the info lager log.

To prevent spamming of the database table there is a rate limit of 1 error per second.

The database log is also pruned after a week.

To do:

 - [x] Use a JS alert
 - [x] Also log the peer's IP address
 - [x] Show the user-agent in the log
 - [x] Translation of the new strings
 - [x] Document (and put in toc) `filter_log_format_stack`

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_logging/models/m_log_ui.erl
+++ b/modules/mod_logging/models/m_log_ui.erl
@@ -73,7 +73,8 @@ insert_event(Props, Context) when is_list(Props) ->
     end,
     Message = [
         {user_id, UserId},
-        {type, Type}
+        {type, Type},
+        {remote_ip, m_req:get(peer, Context)}
     ] ++ proplists:delete(type, Props2),
     MsgUserProps = maybe_add_user_props(Message, Context),
     z_db:insert(log_ui, MsgUserProps, Context).
@@ -192,6 +193,7 @@ install(Context) ->
                     rsc_id int,
                     user_id int,
                     type character varying(80) not null default ''::character varying,
+                    remote_ip character varying(32),
                     props bytea,
                     created timestamp with time zone not null default now(),
 
--- a/modules/mod_logging/models/m_log_ui.erl
+++ b/modules/mod_logging/models/m_log_ui.erl
@@ -1,0 +1,215 @@
+%% @author Marc Worrell
+%% @copyright 2019 Marc Worrell
+%% @doc Model for ui log messages.
+
+%% Copyright 2019 Marc Worrell
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


mworrell:
Screen Shot 2019-05-21 at 13 24 14

modules/mod_logging/models/m_log_ui.erl
```diff
@@ -0,0 +1,215 @@
...
+    m_value/2,
+    insert_event/2,
+    insert/2,
+    get/2,
+    search_query/2,
+    periodic_cleanup/1,
+    install/1
+]).
+
+
```
DorienD: Sometimes there are double newlines and sometimes single ones, is there a logic behind it?

modules/mod_logging/templates/admin_log_ui.tpl
```diff
@@ -0,0 +1,73 @@
+{% extends "admin_log_base.tpl" %}
+
+{% block title %}{_ Log ui events _}{% endblock %}
+
+{% block title_log %}{_ Log user-interface events _}{% endblock %}
```
DorienD: Why do you use a hyphen in between user and interface here?

modules/mod_logging/templates/admin_log_ui.tpl
```diff
@@ -0,0 +1,73 @@
...
+            <th>{_ Severity _}</th>
+            <th>{_ Message _}</th>
+            <th>{_ User _}</th>
+            <th>{_ Stack _}</th>
+        </tr>
+        <tr>
+            <td></td>
+            <td>
+                <select name="type" id="log_severity" class="form-control">
+                    <option {% if q.type == 'debug' %}selected{% endif %} value="debug">Debug</option>
```
DorienD: No translations here?

modules/mod_logging/models/m_log_ui.erl
mworrell: No, sometimes I just want to have a bit more "visual space" between functional different parts. Helps with reading and grouping.

modules/mod_logging/templates/admin_log_ui.tpl
mworrell: Hmmm, I guess double-dutch?  better without?

modules/mod_logging/templates/admin_log_ui.tpl
mworrell: Copied this from the other log screen, also no translations there.
We might want to add some, though I think this was left untranslated so that it matches the texts in the log rows underneath (those come from the database and are left as-is).

mworrell:
Made the stack trace more readable:

Screen Shot 2019-05-21 at 17 48 43

mworrell:
Build is passing, can be merged.

mworrell:
Question: right now this shows an alert on every JS error.

That might actually a bit too much, as sites also use external javascript which throw errors.

Maybe we should only show an alert when the error happens during a submit or postback?

mmzeeman:
Fixed a bug in the error handler. The error handler was crashing itself in most situations.

PR #2087, opened 2019-05-23
Fix race condition in rsc_gone

mworrell:
### Description

Fix a problem where a race condition in rsc_gone might end up canceling the delete transaction.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Only risk: if we have a rollback during the delete then we end up with a `rsc_gone` entry.

In psql >= 9.5 we can use `on conflict do nothing`, but right now earlier versions of psql are still in active use and default in (for example) Ubuntu 16.

PR #2091, opened 2019-05-29
mod_admin: Convert predicate to binary to fix matching

DorienD:
### Description

When using the find dialog (not the new_find dialog) there was an error on the matching of predicate. The predicate was not a binary as assumed in the case.


mworrell:
Thanks!

PR #2095, opened 2019-06-05
Exometer core config update

### LINKED ISSUE #2088: Exometer is not started correctly
Exometer has been split in different packages. We only need `exometer_core`, which is correctly listed in zotonic.app.src.

For normal applications this would work, but in order to start exometer correctly from the start the `exometer` application needs to be started as well. See: https://github.com/Feuerlabs/exometer_core#exometer-core-start

When this is not done all the exometer configuration in `erlang.config` is not read, and no reporters or standard metrics are loaded.

mmzeeman:
### Description

Fix #2088 

- Upgrade exometer to v1.5.7
- Display a log message when the default exometer config was not loaded.

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
@mmzeeman I assume this is ready for merging?

PR #2097, opened 2019-06-18
admin: show 'unique name' field in quick edit for all meta resources.

mworrell:
### Description

Replaces #1989 

Show the `name` field for all `meta` categories.
Make it required for `category` and `predicate`.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_admin/templates/_action_dialog_edit_basics.tpl
+++ b/modules/mod_admin/templates/_action_dialog_edit_basics.tpl
@@ -19,6 +19,8 @@
 	                <ul class="nav nav-pills">
 	                    <li class="active"><a data-toggle="tab" href="#{{ #main }}">{_ Main _}</a></li>
 	                    <li><a data-toggle="tab" href="#{{ #acl }}">{_ Access control _}</a></li>
+                        {% block tabbar_extra %}
+                        {% endblock %}
 	                </ul>
                 {% endblock %}
                 <div class="tab-content">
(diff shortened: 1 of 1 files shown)


modules/mod_admin/templates/_action_dialog_edit_basics.tpl
```diff
@@ -27,12 +29,23 @@
...
 	                        {% catinclude "_admin_edit_basics.tpl" id in_dialog is_editable=id.is_editable languages=languages show_header %}
 
-	                        {% if id.is_a.category or id.is_a.predicate %}
+	                        {% if id.is_a.meta %}
 	                            <div class="form-group row">
 		                            <label class="control-label col-md-3" for="{{ #unique }}">{_ Unique name _}</label>
                                     <div class="col-md-9">
 		                                <input class="form-control" type="text" id="{{ #unique }}" name="name" value="{{ id.name }}" />
-		                                {% validate id=#unique name="name" type={presence} %}
+                                        {% if id.is_a.category or id.is_a.predicate %}
```
DorienD: @mworrell I'm testing this and thinking should this if not be if `id.is_a.meta`?

mworrell: Only `category` and `predicate` are assumed to have a symbolic (and unique) name in the system.  Other meta categories can exist without an unique name.

Though having an unique name makes it always better when moving meta resources between sites of exporting/importing them from files.

DorienD: But `content_groups` and `user_groups` also need a unique name to work right?

mworrell: Only if you want to export / import the ACL definition. For the rest they just use the resource ids in the lookup tables.

DorienD: ah ok!

PR #2098, opened 2019-06-18
Support 'dmy' input format for dates.

mworrell:
### Description

Replaces #1963 

Support `dmy` as a date formatter in the properties normalization routines.

### Checklist

- [ ] documentation updated
- [x] tests added
- [x] no BC breaks


mworrell:
@linuss can you check this against your original use case?

PR #2100, opened 2019-06-18
Renamed all pgsql references to epgsql

### LINKED ISSUE #2096: Upgrade postgres driver
Our current postgres driver is five years old now. 

I would like to upgrade it because I want to use the JSONB datatypes for a side project I'm working on. The JSONB datatypes are now supported by Postgres 9.5 and up. (Basically all maintained versions of Postgres). The current driver we have (`2.0.0`) does not support this datatype.

`epgsql` supports the JSONB datatype from version `3.2.0` which is 3 years old already.

mmzeeman:
### Description

Fix #2096

Upgrades the epgsql driver we use to version 3.4.0. Please note that this is not the latest version of the driver. But there are slight incompatibilities in 4.x which require more work.

### Checklist

- [X] no BC breaks

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -45,7 +45,7 @@
   {gproc, ".*", {git, "git://github.com/uwiger/gproc.git", {branch, "master"}}},
   {emqtt, ".*", {git, "git://github.com/zotonic/emqtt.git", {branch, "zotonic"}}},
   {poolboy,   ".*",   {git, "git://github.com/devinus/poolboy.git", {tag, "1.5.1"}}},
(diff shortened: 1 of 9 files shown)


mmzeeman:
Nice... has to be done for master as well. :+1:

mworrell:
If this merge ok's on Travis then we can apply the changes there as well.

mworrell:
Master uses:   `{epgsql, "3.1.1"}`

Will try an update to `3.4.0`

mworrell:
Updated master to `3.4.0`

mmzeeman:
That was easy. I forgot that master already moved to epgsql.

PR #2103, opened 2019-06-20
mod_admin: do not toggle 'depiction' if image is embedded in tinymce

mworrell:
### Description

Fix a problem where a connected depiction was disconnected if it was added to the text using tinyMCE.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


DorienD:
Nice!

PR #2121, opened 2019-07-22
core: Fix pattern match on epgsql result

ddeboer:
### Description

Fix error when running sitetests

* Probably caused by #2100.
* Fix pattern match on epgsql result.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


ddeboer:
@mworrell There may be other places in the code this pattern needs to be changed.

ddeboer:
#2122 solves the breakage in a better way.

PR #2122, opened 2019-07-22
Use epgsql error record

mworrell:
### Description

Use the epgsql `#error{}` record instead of the `{error, ...}` tuple.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/db/z_db_pgsql.erl
+++ b/src/db/z_db_pgsql.erl
@@ -25,6 +25,7 @@
 -behaviour(z_db_worker).
 
 -include("zotonic.hrl").
+-include_lib("epgsql/include/epgsql.hrl").
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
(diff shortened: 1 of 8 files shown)


src/db/z_db_pgsql.erl
```diff
@@ -181,9 +182,9 @@ connect(Args, RetryCt) ->
                               [Hostname, Port, ?CONNECT_RETRY_SLEEP div 1000, self()]),
                 timer:sleep(?CONNECT_RETRY_SLEEP),
                 connect(Args, RetryCt+10);
-            {error, {error,fatal,<<"53300">>,_ErrorMsg,_ErrorArgs}} ->
+            {error, #error{ codename = too_many_connections }} ->
                 too_many_connections(Args, RetryCt);
-            {error, <<"53300">>} ->
+            {error, too_many_connections} ->
```
ddeboer: Can this one still match?

mworrell: I am actually a  bit unsure.  According to the type specs of the connect not though.

mworrell: Checked again:

```erlang
-type connect_error() ::
        #error{}
      | {unsupported_auth_method, atom()}
      | invalid_authorization_specification
      | invalid_password.
```

ddeboer: In that case, let’s remove the redundant patterns.

mworrell: Let's assume the type spec is valid and adapt our code.

PR #2125, opened 2019-07-24
Allow pager to accept a list for the 'result' arg.

### LINKED ISSUE #2124: Pagination with id.o.haspart
We would like to have pagers with the `id.o.haspart` (or other edges). The `moreresults` action seems to be able to handle `id.o.haspart` results as `m.search` results. But the pager does not.

mworrell:
### Description

Fix #2124

This allows a list to be the `result` argument of the pager scomp.

Also pass the `page` and `pagelen` as arguments if they are not default or not included in the request query arguments.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Rebased on 0.x

PR #2128, opened 2019-07-25
Make dialyzer happier (part 1 of N)

mworrell:
### Description

Remove dialyzer warnings.

This is an ongoing process and this MR will only reduce the number of warnings.
Subsequent MRs will reduce more warnings.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Going to merge this first part of the "make dialyzer happier" series.

We have now a warning by ranch:

```
Setting Ranch options together with socket options is deprecated. Please use the new map syntax that allows specifying socket options separately from other options.
```

This will be handled in a separate issue.

PR #2133, opened 2019-07-30
Dialyzer 3/3 (passes)

mworrell:
### Description

More dialyzer fixes.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
The last remaining warning is about a `{error, econnrefused}` returned by `epgsql_connect` but not mentioned in its type definition.

We might want to suppress this warning and get a type-spec fix merged into epgsql.

mworrell:
@ddeboer We got green on the Dialyzer build!

https://travis-ci.org/zotonic/zotonic/builds/566139318

Shall we enable Dialyzer in the OTP 19.3 step?
Then we have something like:

```yaml
matrix:
  include:
    - otp_release: 21.3
      env: ELVIS=true
    - otp_release: 20.3
      env: BUILD_DOCS=true
    - otp_release: 19.3
      env: DIALYZER=true
  fast_finish: true
```

Then we have something "special" in all three builds.

/cc @mmzeeman

mworrell:
The new travis config passes the CI testing:

https://travis-ci.org/zotonic/zotonic/builds/566200600

ddeboer:
Nice! Well done. 👍

Issue #2140, opened 2019-08-06
Documentation changes: nginx websocket config


mworrell:
From @karlsson:

```
        # websocket
        location /mqtt-transport {
                 proxy_pass       http://127.0.0.1:8000/mqtt-transport;
                 proxy_set_header Host      $host;
                 proxy_set_header X-Real-IP $remote_addr;
                 proxy_set_header X-Forwarded-For  $proxy_add_x_forwarded_for;
                 proxy_set_header X-Forwarded-Proto $scheme;
                 proxy_http_version 1.1;
                 proxy_set_header Upgrade $http_upgrade;
                 proxy_set_header Connection "upgrade";
                 proxy_set_header Sec-Websocket-Extensions $http_sec_websocket_extensions;
                 proxy_set_header Sec-Websocket-Key $http_sec_websocket_key;
                 proxy_set_header Sec-Websocket-Protocol $http_sec_websocket_protocol;
                 proxy_set_header Sec-Websocket-Version $http_sec_websocket_version;
        }
```

karlsson:
@mworrell 
OK, I can pick this one. When I try to generate docs (make docs) to check the generated html before pushing my pull request I get errors for missing directories or files though:
`export ZOTONIC_SRC=../ MODULES=../apps/zotonic_mod* ; for f in `find ref -name .generate-erlang`; do $f || exit 1; done
export APPS=../apps CORE=../apps/zotonic_core MODULES=../apps/zotonic_mod* ; for f in `find ref -name .generate`; do $f || exit 1; done
ref/templates/.generate: rad 18: ref/templates/meta-item_list-schema_org/types.rst: Filen eller katalogen finns inte
New doc: ref/templates/template_item_list-schema_org/types.rst`

Now the updates are not big so I can push anyway and have you review the pull request?

mworrell:
Closed in #2143

PR #2143, opened 2019-08-06
doc: nginx websocket config (#2140)

karlsson:
### Description

Fix #[2140]

Add documentation for proxying MQTT transport on websocket with nginx.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/doc/developer-guide/deployment/nginx.rst
+++ b/doc/developer-guide/deployment/nginx.rst
@@ -12,6 +12,10 @@ are planning to server (the ``user/sites/your_site/config``
 files). The ``hostname`` value should not contain any port number, if
 you run from port 80: ``{hostname, "test.zotonic.com"}``.
 
+Zotonic also makes use of a websocket connection for MQTT messages
+under the ``/mqtt-transport`` location and so you have to add an extra
+proxy directive for this.
+
 Below is a configuration file we use to proxy nginx to zotonic. Be
 sure to replace all occurrences of ``test.zotonic.com`` with your own
 hostname::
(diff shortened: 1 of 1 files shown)


doc/developer-guide/deployment/nginx.rst
```diff
@@ -12,6 +12,10 @@ are planning to server (the ``user/sites/your_site/config``
 files). The ``hostname`` value should not contain any port number, if
 you run from port 80: ``{hostname, "test.zotonic.com"}``.
 
+Zotonic also makes use of a websocket connection for MQTT messages
+under the ``/mqtt-transport`` location and so you have to add an extra
+proxy directive for this.
+
 Below is a configuration file we use to proxy nginx to zotonic. Be
 sure to replace all occurrences of ``test.zotonic.com`` with your own
 hostname::
```
mworrell: I am not sure why the doc tool suddenly complains, but apparently the default code-syntax is 'erlang'.

We need to change this with:

```
.. code-block:: nginx
```

(And replace the double ':' after "hostname::" with a single ':')

karlsson: Thanks Marc,
After checking the documentation on how to generate documentation (apt install sphinx-common) I am able to check before commit now. :smile: 
Will fix.

mworrell: Then you can do better than me, somehow my local docs build is a bit broken.

karlsson: Well I think I was a bit confused with different python versions and pip install as per the docs. So I ended up just installing sphinx-common (sudo apt-get install sphinx-common). And I think I am on python2.7 now. It complains a bit but generation seems to work OK, I think.
/home/mikael/.local/lib/python2.7/site-packages/sphinx/builders/html.py:1171: RemovedInSphinx20Warning: Now html_sidebars only allows list of sidebar templates as a value. Support for a string value will be removed at Sphinx-2.0.
  self.add_sidebars(pagename, ctx)
..
build succeeded.
The HTML pages are in _build/html.

mworrell:
Great, it passes the build.  Thanks!

PR #2147, opened 2019-08-11
docs: updated nginx proxy description

### LINKED ISSUE #1738: Nginx & SSL
I seem to have an issue with the `{ssl, true}` directive in dispatcher rules, and `{ssl_only, true}` in _zotonic.config_ when using Nginx as a proxy in front of Zotonic 1.0. Using the site as an anonymous user works fine, but when Zotonic attempts to redirect for an https connection, the page doesn't load. 

I have Nginx (1.13) configured for HTTP2, and followed the Zotonic cookbook reference for using Nginx with SSL termination. Regardless of whether the _proxy_pass_ directive in Nginx is set to `http://zotonicip:8000` with `ssl_listen_port` set to _none_ and `ssl_port` set to _443_, or `https://zotonicip:8443` with `ssl_listen_port` set to _8443_ and `ssl_port` set to _443_, any attempt by Zotonic (Cowboy?) to force an https connection breaks things. 

Any suggestions on what I am overlooking, and how to properly handle https redirects gracefully when using Nginx, other than editing all dispatch rules in all modules to be` {ssl, any}`?

karlsson:
### Description

Fix #1738

Rewrote the nginx proxy documentation taking into account that Zotonic always
redirects to SSL. Tested the configuration on a live server.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/doc/developer-guide/deployment/nginx.rst
+++ b/doc/developer-guide/deployment/nginx.rst
@@ -10,13 +10,41 @@ running on your system.
 When proxying, don't forget to check the config files of the sites you
 are planning to server (the ``user/sites/your_site/config``
 files). The ``hostname`` value should not contain any port number, if
-you run from port 80: ``{hostname, "test.zotonic.com"}``.
+you run from port 80/443: ``{hostname, "test.zotonic.com"}``.
 
-Zotonic also makes use of a websocket connection for MQTT messages
-under the ``/mqtt-transport`` location and so you have to add an extra
-proxy directive for this.
+Zotonic configuration
+---------------------
 
-Below is a configuration file we use to proxy nginx to zotonic. Be
+Example of ``~/.zotonic/[release]/zotonic.config`` ip/port settings when
+terminating SSL in nginx proxy and using plain HTTP towards the Zotonic
+backend:
+
+.. code-block:: erlang
+
+  %%% IP address on which Zotonic will listen for HTTP requests.
+  {listen_ip, any},
+
+  %%% Port on which Zotonic will listen for HTTP requests.
+  {listen_port, 8000},
+
+  %%% Port on which Zotonic will listen for HTTPS requests.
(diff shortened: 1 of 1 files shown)


doc/developer-guide/deployment/nginx.rst
```diff
@@ -10,20 +10,53 @@ running on your system.
...
 
-Below is a configuration file we use to proxy nginx to zotonic. Be
+  %%% Outside port on which Zotonic will listen for HTTP requests.
+  {port, 80},
+
+  %%% Outside port zotonic uses to receive incoming HTTPS requests.
+  {ssl_port, 443},
+
+  %%% Force sites to use SSL. Redirects http requests to https.
+  {ssl_only, true},
```
mworrell: On master we now always redirect to ssl, so we need to remove this setting.

mworrell: Checked the source code, we have removed it already, so we can also remove it from the docs here.

mworrell: Will remove it from the other docs as well

karlsson:
I will remove the ssl_only part from the description.
Do I need to mention anything about the ip_whitelist and X-Forwarded-For? Now I include X-Forwarded-For in nginx configuration, but nothing about ip_whitelist in zotonic.config, but maybe 127.0.0.1 is covered by default?

karlsson:
Also I am not able to test termination of http2 on the nginx client side since my site runs on an old Debian/nginx but I think it is just a matter of adding http2 on the listen ssl line:
listen 443 ssl http2;

mworrell:
127.0.0.1 is indeed on the whitelist, as are all non routable IP addresses.

I think your current changes are great and will merge them.

mworrell:
Thanks @karlsson !

Issue #2131, opened 2019-07-29
Self-Signed Certificates and switch Zotonic versions


mworrell:
Safari (and other browsers) have become more sensitive about arbitrary self-signed certificates.

Currently we generate these separately for every site.

Which means that switching between Zotonic versions, cleaning the site, etc. might cause a re-generation of the self-signed certificate.

This gives a problem, as the browser now thinks (rightfully) that the certificate doesn't match the site anymore.

We might be able to reduce this problem by having a single self-signed certificate for all the sites.

This might be placed in `~/.zotonic/ssl/...` (or whatever the Zotonic config directory is).

/cc @ddeboer @arjan @mmzeeman

mworrell:
Possibility:

 1. Add a self-signed certificate in `~/.zotonic/security/ssl/self-signed/...`
 2. Use this self-signed certificate as the root certificate for all self-signed site certificates.

Now each site can have the correct common name and we only need to add this single root certificate to the trust chain of the browser

ddeboer:
What are the arguments against the simplest solution, i.e. to use a single self-signed cert for all sites?

Security shouldn’t really be an issue, because browsers will still give a warning about using self-signed certs, so production sites will not be using them.

mworrell:
The same cert for all sites?  I am ok with that.  @mmzeeman mentioned something about that he prefers the *common name* fits the name of the site.  If we would like to have that then we need a cert per site (of per hostname?).

In principle it is ok to have the same self-signed certificate for all sites on a server. The main issue is switching the certificate when switching Zotonic versions.  Especially on development machines.

mworrell:
Would we like to store all certificates together in a single directory, or use the current setup with a sub-directory in the `priv` directory?

mmzeeman:
I don't know if it will work if you have a certificate with no common name or alternatives in it. My guess was that browsers use the domain-name of the site to select a certificate from the keychain tool.

The main problem is that Safari currently doesn't show the [Yes] [No] dialog anymore. This means that you have to manually add the certificate to your keychain with the Keychain Access Tool.

This is now the error message Safari displays:

<img width="403" alt="Screenshot 2019-07-30 at 10 19 55" src="https://user-images.githubusercontent.com/1024972/62113012-1e141200-b2b4-11e9-92a0-93b1d8cd74ff.png">

As developer you now have to locate the crt file, and drag to the keychain access window. After this, the certificate is not trusted. (red cross)

[...]

PR #2149, opened 2019-08-12
Replace .dev with .test as Google set strict-transport-security on .dev

mworrell:
### Description

Replace the default top-level site domain with ".test" instead of ".dev".

See the comment by @karlsson at https://github.com/zotonic/zotonic/issues/2131#issuecomment-520219958

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


karlsson:
Nice.
One thing about the default value of the schema in the addsite command. Now it is "public". If I do more than one addsite command both sites will have public as default schema and will clash.
What about setting the default name of the schema to site_name instead?
Also when I add the site from the status site web interface the default values are set so that the values in ~/.zotonic/1/zotonic.config are used (if I understand right). Wouldn't it be good to have the same behaviour when using the addsite cli command too?

mworrell:
Yes, schema named after the site sounds right, and is also how I do it manually.

Will check the script and the addsite in zotonic_status.

mworrell:
@karlsson I continued the script changes in #2150

PR #2150, opened 2019-08-13
Cleanup shell scripts

### LINKED ISSUE #1742: OTP-ification: use hex for module installation
~This can be done after #1717 has been merged~

Remove the *install module* scripts and use *hex* instead

mworrell:
### Description

Clean up the zotonic shell scripts and remove functionality that is now covered by rebar3, hex and git.

Basic idea is to remove the shell script commands and replace them with RPC calls into the running Zotonic.

All module and site related functions must move to the Zotonic code (launcher) so that we don't duplicate code between Erlang and the shell scripts.

Only start/stop/debug/etc (which don't need a running Zotonic) should still be managed by scripts.

Fix #1742

 - [x] Default 'addsite' dbschema to the site name.
 - [ ] Add `zotonic_launcher:cmd/1` function for handling script functions (most require zotonic to be running anyway)
 - [ ] Use the `zotonic_status_addsite` Erlang module to add a site
 - [ ] Remove module scripts (use Hex et al instead)
 - [ ] Add new module script (use `zotonic_launcher:cmd/1` and echo modules etc)
 - [ ] Update documentation

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_launcher/bin/zotonic-addsite
+++ b/apps/zotonic_launcher/bin/zotonic-addsite
@@ -37,9 +37,10 @@ function usage {
     echo "  -u <user>    Database user (default: $DBUSER)" 1>&2
     echo "  -P <pass>    Database password (default: $DBPASSWORD)" 1>&2
     echo "  -d <name>    Database name (default: $DBDATABASE)" 1>&2
-    echo "  -n <schema>  Database schema (default: $DBSCHEMA)" 1>&2
+    echo "  -n <schema>  Database schema (default: ${DBSCHEMA:="<site_name>"})" 1>&2
     echo "  -a <pass>    Admin password (default: $ADMINPASSWORD)" 1>&2
     echo 1>&2
+
     #echo 1>&2
     exit 1
 }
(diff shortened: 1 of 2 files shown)


apps/zotonic_launcher/bin/zotonic-addsite
```diff
@@ -165,6 +166,10 @@ function check_options {
     done
 
     SITE=${!OPTIND}
+
+    if [ "$DBSCHEMA" == "" ]; then
```
karlsson: DBSCHEMA=${DBSCHEMA:=$SITE}

karlsson: # OK don't bother since the script shall be replaced anyway.

mworrell:
See also #1863 - I asked about the status of that work there as it could be good replacement of all the shell scripts.

mworrell:
In #2184 we are now reworking the erlscripts.

mworrell:
This is overruled in #2198 

Closing this issue (though the `addsite` command needs to be re-added to the changes in #2189.

PR #2156, opened 2019-08-14
Pguyot patch 3

### LINKED ISSUE #1143: Why does zotonic_setup export functions?

mworrell:
### Description

Fix #1144
Fix #1143

Rebased version of #1144 by @pguyot 

(diff also contains ssl docs commit that has been merged already)
 
### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_launcher/bin/helpers/zotonic_setup
+++ b/apps/zotonic_launcher/bin/helpers/zotonic_setup
@@ -24,7 +24,6 @@ function require_zotonic_running {
         exit 1
     fi
 }
-export -f require_zotonic_running
 
 function require_zotonic_not_running {
     PONG=$($ZOTONIC_CALL -a "zotonic ping" 2>&1)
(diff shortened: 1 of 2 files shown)


mworrell:
@pguyot a late merge. Thanks for your fixes!

PR #2157, opened 2019-08-14
Update rebar3 on make (#1876)

### LINKED ISSUE #1876: On Make - check minimal rebar3 version
Sometimes we use new capabilities of rebar3.
But an older version of rebar3 is not replaced.

We should check the installed version of rebar3, and if needed replace it with a newer version.

See also #1875

karlsson:
### Description

Fix #1876 

Check in Makefile if rebar3 version on https://s3.amazonaws.com/rebar3/rebar3 is updated compared to the latest downloaded for Zotonic build.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ erl_crash.dump
 
 .rebar
 rebar3
+rebar_etag
 rebar3.crashdump
 compile_commands.json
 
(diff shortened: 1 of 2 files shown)


mworrell:
Would it be an idea if we move this _rebar3 check_ to something like `make pull` ?
Then we could also run `git pull` to fetch the newest zotonic.

The _old rebar3_ issue is mostly at play when updating zotonic (or when having a borked rebar3, but then manually removing it would also help).

karlsson:
Yes moving to something like `make pull` would be better I think and "least surprise" for the user.
As it is now it always makes a HTTP request for the HEAD information when you make and even though I have a 2 s timeout if it fails it still complicates the build chain.

mworrell:
Indeed "least surprise".  Could you move the _rebar3 update_ code to something like `make pull`

mworrell:
Thanks you!

PR #2173, opened 2019-08-28
For docker use erlang.config.in as-is. Fixes #2166

### LINKED ISSUE #2166: docker base /etc/zotonic/erlang.config error
Снимок экрана 2019-08-21 в 12 04 05

mworrell:
### Description

Fix #2166 

For Docker, use the `erlang.config.in` file as-is without modification.

If we need an adapted configuration for Docker then we should make a separate config file.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
@ddeboer can you check this?  And maybe add a special docker `erlang.config.docker.in` file?

ddeboer:
This breaks because no `erlang.config.docker` exists. Let’s add that.

mworrell:
@ddeboer do you know what should be in `erlang.config.docker` ?

PR #2179, opened 2019-09-09
Configurable same site attribute

### LINKED ISSUE #2114: Add the possibility to configure CSRF option on the session cookie.
I would like to have the possibility to change the CSRF option (same-site) for a site.

The current setting is a great default. But I would like to override it does not set the same-site property.

Does anyone want an option to set it to strict?

mmzeeman:
### Description

Added a way to configure the `SameSite` attribute of session cookies. 
Fix #2114 

Adds the possibility to remove the `SameSite` attribute, set it to `strict`, or set it to `lax` explicitly.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/doc/ref/configuration/site-configuration.rst
+++ b/doc/ref/configuration/site-configuration.rst
@@ -131,6 +131,15 @@ The following options can be configured:
   The domain the Zotonic session-id and page-id cookies will be set
   on. Defaults to the main hostname.
 
+``{session_cookie_name, <name>}``
+  Configure the name used for the session cookie. Default it is set to
+  ``z_sid``.
+
+``{session_cookie_same_site, <options>}``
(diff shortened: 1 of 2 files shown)


mmzeeman:
Chrome 80 will change the default behaviour of cookies. Cookies which don't have the SameSite attribute will be set to `SameSite: lax`. Chrome 80 will be released early 2020.

This will affect this PR, because setting the attribute to none will remove the SameSite attribute. This will cause Chrome browsers to use the `lax` setting. Note that setting the attribute to `none` will trigger a Safari bug (it will then use `strict`).

Should we close this PR, and don't allow configuration of the session cookies?

mmzeeman:
Closing this PR. Chrome will change the behaviour of cookies without a `SameSite` attribute. Leaving this attribute unconfigurable is better.

PR #2197, opened 2019-10-03
Add dispatch so pager tpl can use this

DorienD:
### Description

Add dispatch so the pager template can use this information.


mworrell:
Funny, in master we had this already https://github.com/zotonic/zotonic/blob/master/apps/zotonic_mod_base/src/scomps/scomp_base_pager.erl#L126

PR #2203, opened 2019-10-23
Less noisy UI logging

### LINKED ISSUE #2202: ui_log_disabled configuration parameter does not work
It looks like the check on [`ui_log_disabled`](
https://github.com/zotonic/zotonic/blob/master/apps/zotonic_mod_logging/src/mod_logging.erl#L117) should be inverted (or the property renamed to `ui_log_enabled`).

Also, even if `is_ui_ratelimit_check/1` returns `false` [a ratelimit message is logged](https://github.com/zotonic/zotonic/blob/master/apps/zotonic_mod_logging/src/controllers/controller_log_client.erl#L65), which still spams our logs.

mworrell:
### Description

Fix #2202

No lager info logging if is_ui_ratelimit_check returns false.
Also fixes the mod_logging.ui_log_disabled config check

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_logging/controllers/controller_log_client.erl
+++ b/modules/mod_logging/controllers/controller_log_client.erl
@@ -58,9 +58,8 @@ log(Body, Context) ->
                     m_log_ui:insert_event(Props, Context),
                     lager:info("UI event: ~s", [Body1]);
                 false ->
-                    lager:info("[ratelimit] UI event: ~s", [Body1])
-            end,
-            ok;
+                    ok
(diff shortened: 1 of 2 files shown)


mworrell:
If reviewed then needs separate merge in master.

ddeboer:
Thanks @mworrell!

mworrell:
Also merged into master with 39d6edf

PR #2212, opened 2019-10-31
translation with undefined value crashes

loetie:
translation with undefined value crashes when merging pages.

```{stop_request,500,{function_clause,[{filter_stringify,stringify_1,[{badarg,[{erlang,iolist_to_binary,[[[<<"nl">>,58,32,undefined,<<"\n">>]]],[]},{admin_rsc_diff,format_value,3,[{file,"zotonic/modules/mod_admin/support/admin_rsc_diff.erl"},{line,172}]}```

if value is converted to binary this does not occure.


mworrell:
Interesting.  That `undefined` value for one of the translations is actually illegal data. It should always be a binary.  Maybe we should sanitize that during the resource update.

Did you check where this `undefined` came from?

mworrell:
Fixed in a bit different way.

PR #2236, opened 2019-11-19
Wip blackhole controller

### LINKED ISSUE #2226: log-client-event dispatch is optional but built-in into core.
The dispatch to log client event is located in `mod_logging`, this module is optional. But the logging itself is built into core `zotonic.js`. This will result in `404` responses when client errors are logged.

Is it ok to make a dispatch for this path in `mod_base` which just returns `204` no-content?

Screenshot 2019-11-07 at 15 21 50

Screenshot 2019-11-07 at 15 20 24

mmzeeman:
### Description

Fix #2226

A controller was added which returns 204 on every request it receives. A dispatch rule was added to mod_base to the `log-client-event` dispatch found in `mod_logging`. When the module is not enabled, the client errors are ignored.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_base/controllers/controller_nocontent.erl
+++ b/modules/mod_base/controllers/controller_nocontent.erl
@@ -34,7 +34,6 @@
 -include_lib("zotonic.hrl").
 
 init(_Args) ->
-    ?DEBUG(nocontent),
     {ok, []}.
 
 resource_exists(ReqData, _Context) ->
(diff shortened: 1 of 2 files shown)


mworrell:
We also need to merge this into the master (1.0).

mmzeeman:
So I leave the issue open and mention that this has to be integrated for 1.0.

PR #2241, opened 2019-11-20
Add _checkouts to the code paths

### LINKED ISSUE #2240: _checkouts is missing from the path
The `_checkouts` directory is missing from the path. When code is compiled for modules in `_checkouts` it will not move to the `_build` directory, but it will stay in the `_checkouts`. Zotonic needs to be able to find the modules.

mmzeeman:
### Description

Fix #2240

Adds `_checkouts` to the path, so compiled modules are found.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_launcher/src/command/zotonic_command.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_command.erl
@@ -214,6 +214,7 @@ code_paths_test() ->
 
 code_paths() ->
     [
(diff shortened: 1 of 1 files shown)


mworrell:
I was wondering if we could do something with the code path in the `code` application.
But for now this will do 👍

PR #2243, opened 2019-11-21
start-nodaemon vs start_nodaemon fixes

### LINKED ISSUE #2242: docker start_nodaemon vs start-nodaemon: cannot start
In various places, there is mixed use of "start-nodaemon", but the actual command is "start_nodaemon" (hyphen vs underscore).  This prevents docker images from starting.

Known places this can be found:
* Dockerfile
* Dockerfile.heavy
* apps/zotonic_launcher/bin/zotonic
* docker/s6-service/zotonic/run

There is a related bug in docker/docker-entrypoint.sh which checks for a .erl file being executable, but does not directly execute that file, it simply needs to check if it exists.

To reproduce, attempt to start a docker container (either built yourself or published at dockerhub): specifically zotonic-heavy is the one in use here.  I am using git master and building images myself, on Ubuntu Linux/x86-64.

jontow:
### Description

Fix #2242 

This PR changes references to start-nodaemon to start_nodaemon, but also allows start-nodaemon in the event someone accidentally uses that.  Also addresses the '-x'(executable) vs '-e'(existing) check.

### Checklist

- [-] documentation updated
- [-] tests added
- [?] no BC breaks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,4 +15,4 @@ RUN apk add --no-cache --virtual build-deps $BUILD_APKS \
 # Use dumb-init to reap zombies, catch signals, and all the other stuff pid 1 should do.
 ENTRYPOINT ["/usr/bin/dumb-init", "-c", "--", "/opt/zotonic/docker/docker-entrypoint.sh"]
 
-CMD ["start-nodaemon"]
+CMD ["start_nodaemon"]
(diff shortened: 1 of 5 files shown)


mworrell:
Good catch!  Thanks.

PR #2268, opened 2019-12-05
Do not emit the link-... event if an edge changes

### LINKED ISSUE #1838: Javascript z_event error while adding media item in tinymce
If you add a media item in the tinymce editor this error will be logged in the console. `z_event: no registered event named: 'links-_some id_-depiction'`

Although it logs this error, it doesn't seem to break anything.

Steps to reproduce
- Login the admin 
- Create new page
- add 'zotonic media item'
- select/upload item

Also, once you have the selected item present in the 'Attached media' block, the error won't be logged. So the event seems to be registered at that point.

mworrell:
### Description

Fix #1838  

Suppress JS console message.
The list of connections is now updated lazily so we don't need to emit the event

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_admin/lib/js/apps/admin-common.js
+++ b/modules/mod_admin/lib/js/apps/admin-common.js
@@ -115,9 +115,6 @@ After a page connection is done. Calls a named wire (that must exist).
 See: _admin_edit_content_page_connections_list.tpl
 */
 window.zAdminConnectDone = function(v) {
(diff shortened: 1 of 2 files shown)


mworrell:
@DorienD can you check this one?

DorienD:
Works!

PR #2270, opened 2019-12-05
mod_admin: Fix crop center view in rsc_edge_media.tpl  Fix #2269

### LINKED ISSUE #2269: Javascript error in admin for crop centered images as depiction
For example:

A text page has a couple of images linked as depictions and one of those images has a specified crop_center value. If you then link or unlink any of the other images there is a growl/javascript error. After that you need to refresh page to do any editing.

This is because the crop_center that is being searched by javascript is id `#crop_center` and not id `#crop_center_(id)`. The value of id `#crop_center` is undefined and then the regex does not work and the javascript error occurs in function setCropFromValue in z.cropcenter.js .
The value in config.hiddenInput is then `#crop_center` instead of `#crop_center_(id)` .

mworrell:
### Description

Fix #2269

Fix crop-center preview in depiction list on admin edit page.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@loetie can you check with this branch?

loetie:
@mworrell seems to be working perfectly

PR #2271, opened 2019-12-06
Add block around language options

loetie:
Add block around language options so it is possible to overwrite/extend this template in modules/sites with inherit possibilities.

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/modules/mod_translation/templates/_translation_edit_languages.tpl
+++ b/modules/mod_translation/templates/_translation_edit_languages.tpl
@@ -1,3 +1,4 @@
+{% block language_options %}
 {% with m.rsc[id].language as r_lang %}
 <div class="form-group">
     <div id="admin-translation-checkboxes">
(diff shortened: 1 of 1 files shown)


mworrell:
I am a bit curious, why a block around the whole template?  Is that more useful for you than just overwriting the template?

loetie:
The idea is that it is possible to add something after the current code without overriding the current code (inherit) so if the original template is updated those still automatically work. 

Example:
`{% overrules%}

{% block language_options %}

{% inherit %}

{% button
    text=_"Auto translate this page"
    class="btn--secondary"
    postback={auto_translate id=id}
%}

{% endblock %}`

mworrell:
Completely valid reason 👍

mworrell:
Also merged into master via 1a80978cb23afe6d0ac38b583ab0b408e0ebd2ae

PR #2272, opened 2019-12-09
mod_filestore: retry upload / download errors

### LINKED ISSUE #2187: mod_filestore: errors are never retried
It seems as if when mod_filestore receives an error when trying to access files stored in S3, it stores that fact in the database and then never retries to retrieve the file after the error is fixed.

I don't really see the point of storing remote error conditions; I would argue it is best to remove this functionality.

mworrell:
### Description

Fix #2187

In mod_filestore, retry upload and download errors after 10 minutes.

This fixes a problem where downloads would not be retried after a transient error happened.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_filestore/mod_filestore.erl
+++ b/modules/mod_filestore/mod_filestore.erl
@@ -154,19 +154,20 @@ manage_schema(Version, Context) ->
     m_filestore:install(Version, Context).
 
 lookup(Path, Context) ->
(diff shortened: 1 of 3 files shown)


mworrell:
@CyBeRoni can you give this branch a try with your setup and see if it fixes your problems?

mworrell:
Ping @CyBeRoni

PR #2278, opened 2019-12-20
mod_wires: The error object is optional, carefully access it. Fixes #2276

### LINKED ISSUE #2276: Zotonic onerror handler can crash.
It is possible that the error object is null when the window onerror handler is called. When that happens the error hander crashes, and the error is not sent to the server.

Screenshot 2019-12-11 10 25 56

Screenshot 2019-12-11 10 26 09

mmzeeman:
### Description

Fix #2276

Access the error object more carefully it is optional.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Also updated 0.x in b198646aa000446601b10d46d45e030c1d085d00

PR #2291, opened 2020-01-10
2290 session active

### LINKED ISSUE #2290: Make it possible for pages to tell the session to stay active
We use zotonic as the basis of a chat system. We have one page where agents pick up incoming chats. Currently the agents are automatically logged off after being inactive for one hour. Some of the agents are very specialised and are not contacted regularly. They usually have the page open minimized and receive the incoming chat request via a desktop notification.

The current way the sessions is kept active can't be influenced from the page. I would like to add a mechanism to make it possible that pages can tell the session to stay active.

mmzeeman:
### Make it possible to ignore user activity for keeping the session alive

Fix #2290

  - Refactored the code which checks for activity. It now takes the whole interval between pings into account.
  - Made the event handlers monitoring user activity passive. This will not block the main ui thread.
  - Add a function `z_activity_ignore` which removes the event handlers which track user activity. This will cause `z_is_active()` to always return `true`.
  - Add a tag `ignore_inactivity` which can be used to indicate that session expiration by user
inactivity should be prevented.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/doc/ref/scomps/scomp_ignore_activity.rst
+++ b/doc/ref/scomps/scomp_ignore_activity.rst
@@ -1,0 +1,18 @@
+.. highlight:: django
+.. include:: meta-ignore_activity.rst
+
+This tag disables zotonic's user activity tracking. Normally zotonic
+indicates to the server if the user is active or not. Normally sessions
+are stopped after 1 hour of inactivity.
+
+Sometimes however it is important to keep the user session alive. This is
+possible with this tag.
+
+Example::
+
+  {# Somewhere in a template #}
+  {% ignore_activity %}
+  ...
+
+This will make sure that the session is kept alive as long is the user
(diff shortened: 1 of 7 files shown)


OldhamMade:
Could the term `ignore_inactivity` be ambiguous? Are there any other types of activity, other than user activity, that could be ignored? Would it be better to be explicit and make the term `ignore_user_inactivity`?

mmzeeman:
Updated the name

mworrell:
@mmzeeman ready for merge?

mmzeeman:
Yes

mworrell:
The master doesn't implement the 'activity' tracking yet - so no merging needed (yet).

PR #2296, opened 2020-01-29
core: Don't close the connection in a beforeunload, but later in an unload. Make sure all ajax requests are async

### LINKED ISSUE #2294: New Chrome 80 behaviour conflicts with zotonic transport logic
Chrome 80 now errors when sending synchronous ajax requests in page unloading state.

Normally all zotonic's ajax requests are done asynchrounously to prevent blocking in the user interface. But when the page goes to unload state, because of a `beforeunload` event being triggered, zotonic drops the websocket connection, and only sends synchronous ajax requests. We did this to reliably send data data to the server at this stage, and also to prevent requests from being cancelled. This was introduced in zotonic version `0.13.6`

https://github.com/zotonic/zotonic/commit/62956bfe839e500fe55d70d7b51c58bacf784368

However, Chrome 80 now errors when you try to send synchronous ajax requests at this stage. My Chrome 79 browser somehow had this feature enabled.

https://www.chromestatus.com/feature/4664843055398912

How to fix this?

  1) Just send async ajax requests, and take the cancelled requests for granted.
  2) It might also be possible to queue the messages, and send them together with the unload beacon message. This is sent asynchrounously and can also carry extra data.

mmzeeman:
Don't close the connection in a beforeunload, but later in an unnload

### Description

Fix #2294

This PR keeps the connection open after a `pageunload` and makes sure all ajax
requests are asynchronous. This prevents the browser from sending synchronous
ajax requests after a `pageunload` and prevents the throw/deprecation warning
in Chrome.

Note: The connection will not be automatically restarted when the page is unloading
state after a `pageunload`.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_base/lib/js/apps/zotonic-1.0.js
+++ b/modules/mod_base/lib/js/apps/zotonic-1.0.js
@@ -147,15 +147,9 @@ function z_set_page_id( page_id, user_id )
     $(window).bind('beforeunload', function() {
         z_page_unloading = true;
 
-        // Stop the websocket. This prevents a connection interrupted error.
-        z_websocket_stop();
-
-        // Abort an open comet connection
-        if (z_comet) {
-            try { z_comet.abort(); } catch(e) { }
-            z_comet = undefined;
-        }
-
+        // Keep the connection open, but because the unloading flag is set
+        // the connection will not be automatically restored if it is dropped.
+        
         setTimeout(function() {
             z_page_unloading = false;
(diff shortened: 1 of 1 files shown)


mworrell:
Any changes needed in master as well?

mworrell:
Checked, nothing to change in master.

Merging this for the next 0.x release

PR #2297, opened 2020-01-29
1329 set hsts headers

### LINKED ISSUE #1329: There should be a way to set HSTS headers

mmzeeman:
### Configure HSTS headers

Issue #1329

This PR makes it possible to add a HSTS header to your site.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
We should also add this to master

PR #2299, opened 2020-01-31
mod_ssl: Use rfc7919 dhe parameters

### LINKED ISSUE #2293: Use of custom DHE key exchange parameters is marked as "insufficient" (SSL)
Zotonic uses custom DHE key exchange parameters. Automatic checking tools mark the use of parameters like these as insufficient. RFC7919 mentions that clients are allowed to drop the connection if they find that servers use parameters which they can't check, or are not known strong.

I would like to change zotonic so it doesn't generate DHE parameters anymore, and uses known strong DHE parameters. I would like to make it possible that zotonic can easily switch between parameters because of security reasons.

RFC7919 - https://tools.ietf.org/html/rfc7919#section-3.1

Screenshot from a site review done by https://internet.nl

Screenshot 2020-01-16 at 14 25 42

mmzeeman:
### Description

Fix #2293 

This PR removes the code which generates DHE key exchange parameters. This process can take a long time, and it is not recommended. Some automated tools check for the usage of
published DHE parameters.

I added the published RFC7919 parameters to the source. When zotonic starts it checks if it 
needs to write a new parameter file to the `ssl` directory. If it does not find find it it saves 
the parameters published as `ffdhe3072`. It is possible to specify which parameter to use by
configuring this as a site setting `default_dh_param_name`. It can be set to `ffdhe2048`, `ffdhe3072` or `ffdhe4096`. Note that it is still possible to supply your own parameter set
by simply store them in the file named `dh-params.pem` in the `ssl` directory.

I tested the parameters with https://internet.nl

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_ssl_self_signed/mod_ssl_self_signed.erl
+++ b/modules/mod_ssl_self_signed/mod_ssl_self_signed.erl
@@ -70,8 +70,9 @@ check_certs(Context) ->
             {true, true} ->
                 case check_keyfile(KeyFile) of
                     ok ->
-                        case check_dhfile(proplists:get_value(dhfile, Certs),
-                                          m_config:get_value(site, default_dh_param_name, ffdhe3072, Context)) of
+                        ParamName = z_convert:to_atom(
+                                      m_config:get_value(site, default_dh_param_name, ffdhe3072, Context)),
+                        case check_dhfile(proplists:get_value(dhfile, Certs), ParamName) of
                             ok -> {ok, Certs};
                             {error, _}=E -> E
                         end;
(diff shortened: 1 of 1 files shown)


mmzeeman:
Also needs to be done on master

modules/mod_ssl_self_signed/mod_ssl_self_signed.erl
```diff
@@ -90,15 +92,16 @@ check_keyfile(Filename) ->
...
+check_dhfile(Filename, Default) ->
     case filelib:is_file(Filename) of
         true ->
             case is_dhfile(Filename) of
                 true -> ok;
                 false -> {error, {no_dhfile, Filename}}
             end;
         false ->
-            generate_dhfile(Filename)
+            FileData = dh_params(Default),
```
mworrell: Need to map the default to an atom (or to a binary and then use that below), as the `m_config:get_value/3` will return a binary.

mmzeeman: Missed that one. I only tested it with a site config file.

mworrell:
Shall we add this into the 0.54 release?
I guess you need this @mmzeeman ?

mmzeeman:
Yes, thanks...

PR #2304, opened 2020-02-05
1337 ssl application configuration

### LINKED ISSUE #1335: SSL application configuration
When you have a busy site with lots of ssl connections the default configuration options of the ssl session manager don't really work. It may keeps ssl sessions for 24 hours, but has just room for 1000 sessions in the table. When the table is full, all sessions in the table are invalidated. When you have a busy site you see this happen often and the ssl_manager's mailbox is filled with hundreds of messages.

Something needs to be written about this somewhere, and the erlang.config.in needs good ssl session defaults which works for web-servers. Maybe something like this?

```
%%% Ssl application defaults. 
 {ssl, [
  {session_cache_server_max, 20000}, % Increase when you have a lot of connections 
  {session_lifetime, 300}, % 5 minutes (in seconds).
  {ssl_pem_cache_clean, 300000} % Clean pem cache every 5 minutes. (in milliseconds)
 ]},
```

mmzeeman:
### Add better ssl application config to erlang.config.in

Fix #1335 

  - Adds better config values to erlang.config.in
  - Add a warning to the console when zotonic starts with when ssl application is not configured.
  - Add a warning to the admin status page (`admin/status`) when not configured.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_admin/models/m_admin_status.erl
+++ b/modules/mod_admin/models/m_admin_status.erl
@@ -36,6 +36,11 @@
 ]).
 
 %% @spec m_find_value(Key, Source, Context) -> term()
+m_find_value(is_ssl_application_configured, #m{value=undefined}, _Context) ->
+    case application:get_env(ssl, session_lifetime) of
+        undefined -> false;
+        {ok, _} -> true
+    end;
 m_find_value(session_count, #m{value=undefined}, Context) ->
     session_count(Context);
 m_find_value(page_count, #m{value=undefined}, Context) ->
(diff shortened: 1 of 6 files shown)


mmzeeman:
Where should the documentation go on where you should configure this? Or is this good enough?

mmzeeman:
The console:

Screenshot 2020-02-05 at 15 43 36

The status page

Screenshot 2020-02-05 at 15 43 04

mworrell:
Shall we add a link to the Zotonic docs in that warning?

mmzeeman:
That is a good idea... I would also like to point to the `erlang.config` file.

mmzeeman:
Ah... it is possible to get to the config files used by using `init:get_arguments()`. Looking at the other information you can get from there I think it is probably handy to show this information on the status page. Then this message can just say.. erlang.config.

Other interesting information is:

  - root (the installation directory of otp)
  - home (the home directory)
  - config (the config files used).

mworrell:
I assume we also need to move this to the master?

PR #2307, opened 2020-02-20
mod_translation: add option to make languages available in de admin/e…

mworrell:
### Description

Add option to enable languages for editing without enabling them for viewing in the site.

Screenshot 2020-02-20 at 10 40 28

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_translation/mod_translation.erl
+++ b/modules/mod_translation/mod_translation.erl
@@ -301,11 +301,18 @@ event(#postback{message={language_delete, Args}}, Context) ->
             z_render:growl_error(?__("Sorry, you don't have permission to change the language list.", Context), Context)
     end;
 
-event(#postback{message={language_enable, Args}}, Context) ->
+event(#postback{message={language_status, Args}}, Context) ->
     case z_acl:is_allowed(use, ?MODULE, Context) of
         true ->
             {code, Code} = proplists:lookup(code, Args),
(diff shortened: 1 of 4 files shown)


modules/mod_translation/mod_translation.erl
```diff
@@ -301,11 +301,18 @@ event(#postback{message={language_delete, Args}}, Context) ->
             z_render:growl_error(?__("Sorry, you don't have permission to change the language list.", Context), Context)
     end;
 
-event(#postback{message={language_enable, Args}}, Context) ->
+event(#postback{message={language_status, Args}}, Context) ->
     case z_acl:is_allowed(use, ?MODULE, Context) of
         true ->
             {code, Code} = proplists:lookup(code, Args),
-            language_enable(Code, z_convert:to_bool(z_context:get_q("triggervalue", Context)), Context),
+            case z_convert:to_binary(z_context:get_q("triggervalue", Context)) of
```
DorienD: Remove newline?

mworrell: Which newline?

DorienD: There is an newline between `case of` and `<<enabled>`?

DorienD: There is not!

Issue #2306, opened 2020-02-13
width/height on generated img tags


mworrell:
In the past we removed the width/height from the generated `<img>` tags. This to accommodate to the css implementation of browsers from that time.

Now browsers are (will) be using the width/height, in connection to the css, to calculate the size (aspect ratio) of the image and reserve the correct placeholder, even if the image didn't load yet.

Explanation video:
https://www.youtube.com/watch?v=4-d_SoCHeWE

**Proposal 1:** Set the calculated width/height on the img tags.

**Proposal 2:** Use a template to generate the img tag.

The last proposal is more something for master.

mworrell:
@DorienD especially your comments please!

seannaswell:
Very helpful to know, will soon be adding a photo section to a site and would be nice to implement...

mworrell:
If @DorienD is ok then we can add this back in (it used to be there).

DorienD:
Im gonna dig into this

mworrell:
@DorienD I have made a 0.x branch in #2318

You can play around with this branch and see the effects.

I did change the admin css to add:

```css
img {
    max-width: 100%;
    height: auto;
}
```

For the admin this seems to be enough to make everything work again as expected.

mworrell:
Reopened, as we need to merge into the 0.x as well.

mworrell:
Has been merged

PR #2318, opened 2020-02-28
core: generate image width and height attributes. Issue #2306

mworrell:
### Description

Issue #2306

Add `width` and `height` attributes to the generated `img` tags. This enables the browser to calculate the aspect ratio of the image before it is loaded, thus removing any jenk when loading a page with (non cached) images.

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


DorienD:
I will test this week.

mworrell:
Great, thanks for all the checking with the existing sites!

We will merge this after the 0.55 has been released, which is probably today.

PR #2330, opened 2020-03-05
mod_admin: new page dialog, swap columns and small changes

mworrell:
### Description

Modifications to the new page dialog.

 - Swap columns
 - For consistency, use _page_ instead of _content_

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/doc/ref/filters/filter_flatten_value.rst
+++ b/doc/ref/filters/filter_flatten_value.rst
@@ -1,0 +1,4 @@
+
+.. include:: meta-flatten_value.rst
+
+.. todo:: Not yet documented.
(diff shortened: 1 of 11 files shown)


mworrell:
As discussed with @loetie

mworrell:
@loetie let me know if/when this is ok - then we can ship it as part of the 0.55 for your customer.

loetie:
Hi @mworrell,

1. Is it possible to add blocks around the modal-footer div and header "Make a new page or media"? (needed for OR project)
2. I see the cat_exclude parameter now has a comma seperated value as string, the search however does not seem to do anything with it. Should it be added to _action_dialog_connect_tab_find_results.tpl?

The other improvements seem to work very nicely! :)

mworrell:
@loetie - done, you can check if all is ok for your customer

PR #2333, opened 2020-03-06
Add a lock to the pivot table to prevent race conditions

mmzeeman:
### Add a pivot table lock to prevent race conditions

The table `rsc_pivot_queue` will be locked in row share exclusive mode to prevent race
conditions when updating a resource, or adding edges to the resource.

Removed the catch clause. It looks like it was there to prevent race condition errors from propagating. These kind of errors can no longer happen because of the lock.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/support/z_pivot_rsc.erl
+++ b/src/support/z_pivot_rsc.erl
@@ -146,21 +146,16 @@ queue_all(Context) ->
 
 %% @doc Insert a rsc_id in the pivot queue
 insert_queue(Id, Context) ->
-    case z_db:q("update rsc_pivot_queue
-                 set serial = serial + 1
-                 where rsc_id = $1", [Id], Context) of
-        1 ->
-            ok;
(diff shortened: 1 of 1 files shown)


mmzeeman:
I think `share row exclusive mode` is the right mode for this. See: https://www.postgresql.org/docs/9.5/sql-lock.html

mworrell:
merged, also in master.

mmzeeman:
Nice.. I'm wondering if there are other places which need this treatment.

On master we probably also want to upgrade `epgsql` at some point.

mworrell:
On master we use epgsql 3.4.0, don't know what changed in the 4.3.0.

mmzeeman:
The changelog mentions small changes, new error messages, in the connect call. They now have typespecs, that could help with finding problems.

PR #2334, opened 2020-03-07
Upgrade epgsql

mmzeeman:
Upgraded the epgsql dependency to 4.3.0.

It is mostly compatible with the previous version, but it has a lot of improvments.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
I read something about using `null` versus `undefined` for `NULL` columns.

Do we need to change anything to keep getting `undefined`?

mmzeeman:
Missed that one... 

But I still get undefined back in the normal locations

```
> m_rsc:p(1, page_path, z:c(blog)).
undefined
```

mmzeeman:
Aha.. our own driver translates the null values.

See: https://github.com/zotonic/zotonic/blob/master/apps/zotonic_core/src/db/z_db_pgsql.erl#L327

mworrell:
Ah that is good, then it fits what we do with JSON. Having a single value for "this is nothing" is a lot better than `undefined`, `null`, `none`, `nil`, ...

PR #2335, opened 2020-03-08
Add site and dispatch-rule info to metrics user data

mmzeeman:
### Add stats and logging

WIP, requires cowboy 2.7.0 to work.

Fix statistics panels in admin.

Add global stats topics.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
We are currently using Cowboy 2.8.0 - so the Cowboy version should be ok.

mworrell:
These changes are now cherry picked into #2571

PR #2352, opened 2020-03-13
Remove jQuery migrate

mworrell:
### Description

Change some JS code so that jQuery migrate can de removed.

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
@mmzeeman please add any other cleanup you need 👍

mmzeeman:
Didn't find any other things. I was wondering if we should give a warning in the release notes about the removal. It could be that older sites are affected by the removal.

mworrell:
Yes, we should add a warning in the release notes.

mworrell:
Added a preliminary 0.56 release note to the docs.

PR #2377, opened 2020-04-01
Use z.lang cookie to persist language preference

### LINKED ISSUE #2374: Persistent tables used for client language
If the language is changed it is saved into the persistent storage.
This gives an update statement when the session is persisting itself.

The language setting can also be saved on the client, as the master code is already doing.
The master is using the `z.lang` cookie for this.

Replace the persistent language storage with the `z.lang` mechanism in 0.x.

mworrell:
### Description

Fix #2374

This removes storage of the language settings in the persistent storage and replaces it with the same `z.lang` cookie used in the master.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/modules/mod_translation/mod_translation.erl
+++ b/modules/mod_translation/mod_translation.erl
@@ -53,6 +53,9 @@
 -include("zotonic.hrl").
 -include_lib("modules/mod_admin/include/admin_menu.hrl").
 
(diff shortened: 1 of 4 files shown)


mworrell:
Closing, as this merge request included another unrelated commit.

PR #2382, opened 2020-04-02
core: Add a database pool usage warning

mmzeeman:
### Description

Whenever a database connection is checked back into the pool, a check is done. When the check finds that the pool usage is high, it will log a message.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_logging/mod_logging.erl
+++ b/modules/mod_logging/mod_logging.erl
@@ -30,6 +30,7 @@
 -export([start_link/1]).
 -export([
     observe_search_query/2,
+    observe_tick_1m/2,
     observe_tick_1h/2,
     pid_observe_zlog/3,
     observe_admin_menu/3,
@@ -77,11 +78,15 @@ pid_observe_zlog(Pid, #zlog{ props = #log_email{} = Msg }, _Context) ->
 pid_observe_zlog(_Pid, #zlog{}, _Context) ->
     undefined.
 
+observe_tick_1m(tick_1m, Context) ->
+    check_db_pool_health(Context).
 
 observe_tick_1h(tick_1h, Context) ->
     m_log:periodic_cleanup(Context),
     m_log_email:periodic_cleanup(Context),
     m_log_ui:periodic_cleanup(Context).
+
+
 
 
 observe_admin_menu(admin_menu, Acc, Context) ->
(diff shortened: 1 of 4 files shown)


mworrell:
How can we have this warning log a bit less?

I guess when we have high load then we need to warn about the problem, but if we do so on every check-in then we might end up with hundreds of messages.

Maybe we should write the usage percentage to exometer and then have a periodic reporter log the status?

mmzeeman:
Good idea... count the number of high water mark events in a time frame, and raise an alarm when that happens.

mmzeeman:
Maybe something like this to raise alarms... https://github.com/esl/elarm (For 1.0 I think)

mworrell:
Alarms sounds good - though I am a bit uncertain about when to clear the alarm in this case.

Here it is like a trigger - if set then keep it for a certain time (with counter how many times it happened during the last period).

With the db pool it might drop below the alarm level and then up again, as the load fluctuates quickly.

mmzeeman:
We can setup a spiral counter, https://github.com/Feuerlabs/exometer/blob/master/doc/README.md#exometer_spiral-probe. It collects data for 60 seconds. If something happens within that timeframe we can log.

mworrell:
yes, something like that.  we need a reporter for this stat though

modules/mod_logging/mod_logging.erl
```diff
@@ -201,6 +206,33 @@ code_change(_OldVsn, State, _Extra) ->
...
+                    Context);
+        {ok, _} ->
+            ok;
+        {error, not_found} ->
+            ok
+    end,
+
+    case exometer:get_value([zotonic, Site, db, pool_high_usage], one) of
+        {ok, [{one, HighCounts}]} when HighCounts > 0 ->
+            z:info("Database pool usage is close to exhaustion, ~s", [Advice],
```
mworrell, line 228: What if the pool is full - can `z:info` handle that?
Or do we need to add a check in the db part of the logger to not log to the db if the pool is exhausted?

mmzeeman: Hmm... the pool usage can fluctuate, depending on what happens. I assume I can log here, as the event is, on average 30 second old.

PR #2386, opened 2020-04-06
Reorg site supervisor (0.x)

mworrell:
### Description

Reorganize the site supervisor.

This places the services like pivot and edge-log underneath their own one-for-one supervisor.

These services are allowed to crash and restart without affecting the rest of the running site.

This reorg is done because of a crash autopsy where the pivot server crashed due to a slow database.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
After this has been merged a separate merge request for master will be made.

PR #2406, opened 2020-04-12
Binary keys and maps rsc and medium

### LINKED ISSUE #2405: Change (stored) rsc property keys to binaries
This is a somewhat bigger issue.

Currently we are using a property list to store all properties of resource.
All properties are using an atom for their key.

As atoms are in short supply, we check on update if property keys map to known atoms. This prevents exhaustion of the atom tables.

This worked because the atoms are defined in the templates and code of the site. Compiling the templates and loading the code defined all atoms.

In the future this might not work anymore. The interface is made more dynamic and the updates are relayed through the API (MQTT or HTTP) from non-template code.

To make it possible to define new properties we need to move to binaries as property keys.

Positive:

 * Endless property names, atom space is safe
 * Compatible with next phase where we want to store the properties as JSONB

Negative:

 * More memory used for storage
 * Slower overall as more data (the binary keys) is copied than with atoms (single word)

/cc @mmzeeman @ddeboer

mworrell:
### Description

Fix #2405

Use binaries for model `m_get` path, rsc and medium properties.

Also depends on https://github.com/zotonic/template_compiler/pull/13 (**Merged**)

### Checklist

- [x] documentation updated
- [x] tests added
- **BC breaks**


apps/zotonic_mod_admin_predicate/src/mod_admin_predicate.erl
```diff
@@ -146,18 +146,17 @@ take([Id|L], N, Acc) ->
...
         true ->
-            Subjects = proplists:get_all_values(predicate_subject, Props),
-            Objects  = proplists:get_all_values(predicate_object, Props),
+            Subjects = maps:get(<<"predicate_subject">>, Props, []),
+            Objects  = maps:get(<<"predicate_object">>, Props, []),
             m_predicate:update_noflush(Id, Subjects, Objects, Context),
 
-            Props1 = proplists:delete(predicate_subject,
-                        proplists:delete(predicate_object, Props)),
+            Props1 = maps:remove(predicate_subject,
```
mmzeeman, line 158: Shouldn't this be a binary too?

mworrell: Indeed, good catch, updating the code!

PR #2424, opened 2020-05-04
Use zotonic qdate localtime

### LINKED ISSUE #2423: Users which have set their tz to Asia/Tehran can't logon
Due to a problem in `qdate_localtime` it is not possible for users in this timezone to logon. The reason is that there is an omission in the tz database for this timezone. See: https://github.com/choptastic/qdate_localtime/issues/2

I have forked `qdata_localtime` and made a rule of the thumb rounding rule.

mmzeeman:
### Description

Zotonic's qdate_localtime contains a rounding rule for "Asia/Tehran" 

Fix #2423 

Change the rebar.config and lock files so zotonic/qdate_localtime is used instead of choptastic/qdate_localtime

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman:
Yep... PR for master is ready too.

PR #2429, opened 2020-05-09
mod_survey bugfixes

arjan:
Found some bugs while creating a survey. Also updated the dutch translation strings.


modules/mod_survey/templates/_admin_survey_editor_results.tpl
```diff
@@ -1,4 +1,4 @@
-{% with q.id|default:id as id %}
+{% with q.id|to_integer|default:id as id %}
```
mworrell, line 1: Or should we use `m.rsc[q.id].id|default:id` here?

mworrell:
Thanks! Merging this for today's release.

PR #2450, opened 2020-06-08
Added encoding and decoding of json values

mmzeeman:
### Description

Added logic for storing `props` in jsonb format.

Added logic to store extra props in json format when the table has a `props_json` column. 

- [x]  Use `jsxrecord` to be able to handle erlang records.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


src/db/z_db.erl
```diff
@@ -503,17 +540,33 @@ select(Table, Id, Context) ->
...
+                         PropsCol when is_list(PropsCol) ->
+                             lists:keydelete(props, 1, R) ++ PropsCol;
+                         _ -> R
+                     end,
+
+            case proplists:get_value(props_json, Props1, not_found) of
+                undefined -> lists:keydelete(props_json, 1, Props1);
+                Bin when is_binary(Bin) ->
+                    JP = json_decode(Bin),
+                    Normalised = [{maybe_to_atom(K), V} || {K, V} <- JP],
```
mworrell, line 554: This might be tricky - as it depends on the loaded modules which props are defined.  In the master we force more erlang modules (esp. models) to be loaded, so that more atoms are defined.  Maybe you need to do something similar here as well. Otherwise you might end up with atoms and binaries for the same keys, depending when they are loaded.

mworrell:
Merging.

Please note the remark about the atoms.

PR #2451, opened 2020-06-09
Shortlink and Canonical links should be full URLs

OldhamMade:
### Description

While both Shortlink and Canonical links _can_ be relative, it is both
common practice and expected by search engines to use full URLs,
especially when no <base> tag is used.


mworrell:
Thanks!

One note: in the 1.x the argument `use_absolute_url` has been renamed to `absolute_url`

(I don't know why... remembering it from the upgrading notes)

OldhamMade:
Thanks for the notice. I've pushed an update.

mworrell:
Great and thanks!

PR #2457, opened 2020-06-15
Added support for https_proxy in GNUMakefile

MayamaTakeshi:
### Description

GNUMakefile doesn't work if the host is behind an https proxy

This PR adds calls to 'httpc:set_options()' before calls to 'httpc:request()' in case env var $https_proxy is set.


mworrell:
Thanks! Nice addition.

Wondering if we should extend this to other http(s) routines inside Zotonic itself.  But for now it is great that we can build.

PR #2467, opened 2020-06-27
Update system security for shell scripts

crownedgrouse:
- Protect shell command of path with dash and space with --
        - rewrite variables from $VAR to ${VAR}
        - update missing brackets surrounding some variables


mworrell:
Thanks!

Good changes, merging.

PR #2474, opened 2020-07-02
OAuth authentication and signup

### LINKED ISSUE #2404: Check / re-enable OAuth logins
The OAuth logins via LinkedIn, Facebook, Twitter etc needed `mod_server_storage`.
As this module is now available we can add the login again.

The page(/controller) where the external redirect needs to be adapted so that first the page is rendered, and then a server request is done. This because otherwise the server is not receiving the `z.auth` cookie.

Same for the signup, which also used _secret_ temporary storage.

mworrell:
### Description

Fix #2404 

 * Signup using server side storage
 * Enable OAuth authentication with Facebook, Twitter and LinkedIn.
 * Remove Instagram (they removed their API)
 * More generic OAuth service modules

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
(rebased on master for testing)

PR #2488, opened 2020-07-17
Fix a problem where setting default language could not enable the language

### LINKED ISSUE #2487: Problem if one language
if you set 1 language, then there is no way to edit the title of the page

mworrell:
### Description

Fix #2487 

 - If a default language is set, force it to be enabled
 - On admin edit page, ensure a language tab is selected

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_config.erl
+++ b/apps/zotonic_core/src/models/m_config.erl
@@ -298,37 +298,54 @@ set_value_db(Module, Key, Value0, Context) ->
     end.
 
 %% @doc Set a "complex" config value.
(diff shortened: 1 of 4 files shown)


heiheshang:
admin
nothing has changed for me

mworrell:
In the language settings, on admin > translation, try changing the language to en and then back to ru.

There was a problem with writing the config, so the config needs to be written again.

heiheshang:
yes, this is how it works

PR #2489, opened 2020-07-18
Let output valid YAML for config

crownedgrouse:
### Description
	Add # comment on titles
	Add dynamic underlining depending node name length

Let output valid YAML for config. This allow external tools to validate or use Zotonic YAML config.


mworrell:
Nice!

Thanks and merged 👍

PR #2501, opened 2020-08-05
Use sassc for .sass compilation.

### LINKED ISSUE #2500: Merge sass handling from 0.x
sass file handling has been changed to use `sassc`

Merge changes from https://github.com/zotonic/zotonic/pull/2499

mworrell:
### Description

Also better https listener errors.

Fix #2500

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_config.erl
+++ b/apps/zotonic_core/src/support/z_config.erl
@@ -249,7 +249,8 @@ default(dbpassword) -> "zotonic";
 default(dbdatabase) -> "zotonic";
 default(dbschema) -> "public";
 default(filewatcher_enabled) -> true;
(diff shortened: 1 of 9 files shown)


mworrell:
@mmzeeman should we use `sassc` for `.less` files or keep using `lessc` for that?

mworrell:
And same question for `.scss` @mmzeeman

mworrell:
I see that `.scss` files are indeed handled with `sassc` now.

Looking at the structure of _bootstrap_ we might also want to check parent directories for the non-underscore .sass/.scss file?

mworrell:
Now checking the parent directories for any non-underscore file.

PR #2502, opened 2020-08-06
Add JSON as export format and some fixes [ADAPT FOR 1.x]

### LINKED ISSUE #2497: The id is not included in the common_properties of a resource.
When looking at the export routines I saw that the id was not included in the common_properties of a resource. Is this an omission, or is there a reason behind that?

### LINKED ISSUE #2498: Export query only exports the first 20 resources
When you define a content query, and use it to export data via `/export/query/csv/<id>`, only the first 20 resources are exported.

mmzeeman:
### Add a json exporter.

Fix #2497 
Fix #2498

- Added a json exporter which can be used by other modules too.
- Added a json export button to the resource sidepanel
- Added the id of a resource to the properties.
- Fixed the limit of 20 exported resources for admin content queries.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/include/zotonic.hrl
+++ b/include/zotonic.hrl
@@ -234,6 +234,9 @@
 %% @doc Call the translate function, 2nd parameter is context
 -define(__(T,Context), z_trans:trans(T,Context)).
 
+%% @doc Extra trans record definition to ease JSON mapping of translatable strings
+-record(trans, { tr = [] :: list( {atom(), binary()} )}).
+
 %% The name of the session request parameter
 -define(SESSION_PAGE_Q, "z_pageid").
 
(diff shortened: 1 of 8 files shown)


mworrell:
I didn't include this in the 0.58.0, is it ready for inclusion in the next release?

mworrell:
@mmzeeman ready to be merged?

mmzeeman:
Maybe we should not include this in 0.x.

mworrell:
I have set the milestone to 'roadmap' so that we can have a look what we want to include in master.

mworrell:
Just checked the 1.x - there we do a `m_rsc:get/2` and then dump the contents of that map as a JSON data value.

I think it might be better for the 1.x to use the output of the m_rsc_export model.  Then we also export the edges.

Also, the output is now a map of `id => rsc`.  I suspect a list is better, then we keep the order of the results.

@mmzeeman  What do you think?

mworrell:
See the changes in #3965

PR #2523, opened 2020-09-07
Log errors on resource update

rl-king:
Prevents silent failures if throw is never caught,
which happens in controllers.

### Description

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Thanks!

Makes a lot of sense.

Note that in the master (1.x and later) the exceptions are removed and we are using the normal Erlang method of returning an ok or error tuple.

PR #2525, opened 2020-09-15
Added an extra timeout before the page reload to give the browser mor…

### LINKED ISSUE #2524: Logon problem when user has multiple open tabs
There is a logon problem when the user has two open tabs and is not authenticated.

Steps to reproduce:

1) Open two tabs
2) Go to `/admin` in both tabs
3) Logon to admin on one of the tabs and wait.
4)  Now two things can happen.
     1) Both tabs go to `/admin` *expected behaviour*
     2) One tab stays on the logon window, the other briefly logs in and then logs out again.

In Safari one of the tabs will reload because the session got invalidated, but it uses the old `z_sid` cookie value causing a reloa (because the session is no longer there and a new one will be started).
This will overwrite the `z_sid` value of the authenticated tab, causing it to logoff.

I can't reproduce this on Chrome, but both Firefox and Safari sometimes have this behaviour. It looks like it is slow to synchronise the new cookie value with other tabs.

mmzeeman:
…e time to synchronize the cookie between tabs

### Description

Fix #2524

Added an extra timeout to the place where the page detects an auth change in the session. This gives the browser more time to sync the auth cookie between tabs. This prevents that the reload is being done with the previous cookie. When this happens, zotonic creates a new session, because the old session has got a new id. Zotonic then sends a new cookie, which is synced with the other tabs, effectively immediately logging off a user. This PR fixes that.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_base/lib/js/apps/zotonic-1.0.js
+++ b/modules/mod_base/lib/js/apps/zotonic-1.0.js
@@ -576,10 +576,12 @@ function z_transport_session_status(data, msg)
                         z_session_status_ok(data.page_id, data.user_id);
                         break;
                     case 'auth_change':
-                        // The user-id of the session is changed.
-                        // A new session cookie might still be on its way, so wait a bit
                         if (data.page_id == z_pageid) {
-                            z_session_restart(z_pageid);
+                            // The user-id of the session is changed.
+                            // A new session cookie might still be on its way, so wait a bit
+                            setTimeout(function() {
+                                z_session_restart(z_pageid);
+                            }, 1000);
                         }
                         break;
                     default:


mworrell:
Did you test this with the Safari 14?

modules/mod_base/lib/js/apps/zotonic-1.0.js
```diff
@@ -576,10 +576,12 @@ function z_transport_session_status(data, msg)
...
                     case 'auth_change':
-                        // The user-id of the session is changed.
-                        // A new session cookie might still be on its way, so wait a bit
                         if (data.page_id == z_pageid) {
-                            z_session_restart(z_pageid);
+                            // The user-id of the session is changed.
+                            // A new session cookie might still be on its way, so wait a bit
+                            setTimeout(function() {
+                                z_session_restart(z_pageid);
+                            }, 1000);
```
mworrell, line 584: Could the timeout be shorter?  Like 500 msec?  Or is this from testing?

mmzeeman: This was also tested with Safari 14. FF also had problems with the too fast reload. I thought 1 second would be more than fast enough for users, but especially for the browsers.

PR #2529, opened 2020-10-05
Bump mochiweb version to use better TLS cipher suite.

mmzeeman:
### Description

Upgrade to mochiweb, which creates a safer TLS cipher suite based on the available algorithms in OTP and OpenSSL.

See: https://github.com/zotonic/mochiweb/pull/12 for details.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Is this ok for the upcoming 0.58?

mmzeeman:
Yes.. we run it in production for a couple of months now. The ssl check site https://internet.nl is happy.

PR #2531, opened 2020-10-20
Use zotonic_ssl 1.0.3

mmzeeman:
### Description

Push collecting the save tls socket options to `zotonic_ssl`. This library knows which options are safe to use for the OTP release the user is using.

- [x] Update rebar.lock with new hex hash of `zotonic_ssl` 1.0.3

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/rebar.config
+++ b/apps/zotonic_core/rebar.config
@@ -49,7 +49,7 @@
 %    {template_compiler, "1.2.0"},
     {dispatch_compiler, "1.0.0"},
     {cowmachine, "1.5.0"},
-    {zotonic_ssl, "1.0.2"},
+    {zotonic_ssl, "1.0.3"},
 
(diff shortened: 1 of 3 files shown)


mworrell:
The test seems to crash on calling:

```erlang
zotonic_ssl_option:get_safe_tls_server_options([], [])
```

mmzeeman:
Yes, I want to check out OTP 23.1 too. It allows session tickets, this makes reconnecting tls sockets much faster.

mworrell:
Best to rebase on master - there were some dep updates there as well

mmzeeman:
It looks like there are no conflicts with master. Just a clean patch on the hash and version numbers.

mworrell (approved):

PR #2537, opened 2020-10-27
Fix a problem with OAuth flow on Safari 14. Issue #2536

### LINKED ISSUE #2536: OAuth flow problems on Safari 14
Safari 14 (iOS, iPadOS and macOS) introduces new cookie security.

Because of a bug in these changes the session cookie is not passed during an OAuth flow.

mworrell:
### Description

Fix #2536

Fix a problem with Safari 14 and the OAuth authentication flow.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Failed checks were due to Travis problems, one test ok'd so the basics are good.

PR #2540, opened 2020-10-28
Generate lang switch URLs with the correct slug

mworrell:
### Description

Change the language switch to generate URLs with the correct slug if an id is known.

This prevents the use of URLs with slugs belonging to another language.
Especially Google is prone to find those "wrong" URLs as it is using Javascript to crawl sites.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Merging, as we need it and Travis is having problems.

PR #2548, opened 2020-11-19
core: Fix m_rsc_export for person resources

robvandenbogaard:
### Description

Privacy filtering on person resources looks into proplist values
recursively, because export data is nested. Also fixes the list
fold parameter ordering.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Thanks!

PR #2563, opened 2020-12-30
Only warn about db pool when waiting period increases.

mworrell:
### Description

This allows serializing db pool requests during peak load, without emitting warnings.

If the db connection-pool takes more than 10msec to return a connection there will be a warning about the pool being close to full.

If the db connection-pool returns `full` then the warning about the pool exhaustion is raised.

Also a statistic `zotonic.[site].db.connection_wait` is added.
This is a histogram with the µsec wait time needed to get a connection from the pool.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/db/z_db.erl
+++ b/apps/zotonic_core/src/db/z_db.erl
@@ -240,16 +240,17 @@ has_connection(Context) ->
 
 
 %% @doc Transaction handler safe function for fetching a db connection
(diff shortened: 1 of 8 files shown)


mworrell:
@mmzeeman if you are ok, then you can merge.

The stats should be cleaned up in another merge request.

mmzeeman:
It is looking fine.

mworrell:
Thanks for checking, going to merge.

PR #2602, opened 2021-02-19
Cotonic startup improvements.

mworrell:
### Description

This fixes a problem with lost messages during startup.
Also, by pre-connecting the mqtt connection the bridge is started earlier.

 * Use cotonic.ready promise
 * Use / topics
 * Pre-connect the websocket
 * Catch early topic clicks and submits, replay on cotonic.ready

See also the companion pull request https://github.com/cotonic/cotonic/pull/45

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Looks like a way more robust way to initialise everything.

PR #2604, opened 2021-02-20
Add cotonic head init in base

mmzeeman:
### Description

The cotonic init code which is loaded with a all include head, was not present in base.tpl. This PR adds an include in _base.tpl.


mworrell:
We could also just do the standard all include of "_html_head.tpl"

That would make this a better example of how to make a base template.

PR #2605, opened 2021-02-21
Fix #1628: Use summary filter for meta description

### LINKED ISSUE #1628: Use |summary in mod_seo
Can we change the way the meta description is being used in the _html_head.tpl from mod_seo. I would prefer to use the |summary filter so it will take the body text into account if there is no summary.

DorienD:
### Description

Fix #1628

Use summary filter for the meta description tag

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/modules/mod_seo/templates/_html_head_seo.tpl
+++ b/modules/mod_seo/templates/_html_head_seo.tpl
@@ -21,7 +21,7 @@
 					{% else %}
 						<meta name="keywords" content="{% for predicate in id.op %}{% if predicate /= "depiction" %}{% for oid in id.o[predicate] %}{{ oid.title }}, {% endfor %}{% endif %}{% endfor %}{{ keywords }}" />
 					{% endif %}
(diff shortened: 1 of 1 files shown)


modules/mod_seo/templates/_html_head_seo.tpl
```diff
@@ -21,7 +21,7 @@
 					{% else %}
 						<meta name="keywords" content="{% for predicate in id.op %}{% if predicate /= "depiction" %}{% for oid in id.o[predicate] %}{{ oid.title }}, {% endfor %}{% endif %}{% endfor %}{{ keywords }}" />
 					{% endif %}
-					<meta name="description" content="{{ id.seo_desc|default:id.summary|default:description|escape }}" />
+					<meta name="description" content="{{ id.seo_desc|default:id|summary }}" />
```
mworrell, line 24: To prevent applying summary to the seo_desc, I think we might need:

```
{{ id.seo_desc|default:(id|summary) }}
```

PR #2606, opened 2021-02-23
Refactor permissions to check use of mod_logging

DorienD:
### Description

Refactor ACL check `is_admin` to check permission of use of the mod_logging module. 

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


robvandenbogaard (approved):
Looks preferable like this to me too indeed.

mworrell (approved):

PR #2614, opened 2021-02-28
fix address widget everywhere (#2613)

### LINKED ISSUE #2613: address widget always visible + wrong address feature checkbox in category page
Address widget is always visible on an edit page, even if its category has the  `Show address on edit page` flag disabled.

Steps to reproduce the problem:

- Create a page in any category which has not the `Show address on edit page` flag enabled (e.g. `Text`).

Zotonic version: current master.

First of all, in the category page you will see a wrong value for the `Show address on edit page` checkbox, because of this default value which makes the flag always checked:

https://github.com/zotonic/zotonic/blob/041419f2042a178cef4001f05003aad9a3d1cf71/apps/zotonic_mod_admin/priv/templates/_admin_features.category.tpl#L6

Moreover, here is another default value which makes the widget always visible:
https://github.com/zotonic/zotonic/blob/041419f2042a178cef4001f05003aad9a3d1cf71/apps/zotonic_mod_admin/priv/templates/_admin_edit_main_parts.tpl#L7

Removing the `` |if_undefined:`true` `` filter seems to fix this issue.

udanieli:
### Description

Fix #2613

This PR removes a couple of default values:

- one default which makes the address widget always visible in admin
- another default which makes the `show address` feature always checked in the category edit page

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_mod_admin/priv/templates/_admin_edit_main_parts.tpl
+++ b/apps/zotonic_mod_admin/priv/templates/_admin_edit_main_parts.tpl
@@ -4,7 +4,7 @@
 {% catinclude "_admin_edit_basics.tpl" id show_header %}
 {% all catinclude "_admin_edit_content.tpl" id %}
 
-{% if id.category_id.feature_show_address|if_undefined:`true` %}
+{% if id.category_id.feature_show_address %}
     {% catinclude "_admin_edit_content_address.tpl" id %}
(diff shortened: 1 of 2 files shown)


mworrell:
Good catch - the empty value ("false") is removed on save, so is never present.

mworrell:
Thanks!

PR #2619, opened 2021-03-01
Rename 'feature_..' to 'is_feature_..'

### LINKED ISSUE #2613: address widget always visible + wrong address feature checkbox in category page

mworrell:
### Description

Fix #2613

Rename the feature flags of the categories to "is_..." as then a proper boolean value is 

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_mod_admin/priv/templates/_admin_edit_main_parts.tpl
+++ b/apps/zotonic_mod_admin/priv/templates/_admin_edit_main_parts.tpl
@@ -4,8 +4,15 @@
 {% catinclude "_admin_edit_basics.tpl" id show_header %}
 {% all catinclude "_admin_edit_content.tpl" id %}
 
-{% if id.category_id.feature_show_address %}
+{% if id.category_id.is_feature_show_address|if_undefined:true %}
     {% catinclude "_admin_edit_content_address.tpl" id %}
+{% endif %}
+
+{% if id.category_id.is_feature_show_geodata
+         |if_undefined:(id.category_id.is_feature_show_address)
+         |if_undefined:true
+      or id.location_lat
+      or id.location_lng %}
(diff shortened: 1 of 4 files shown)


udanieli:
This way the address widget is everywhere again and the checkbox in every category page is checked, even if that value is not set. I guess this is the behavior you expect.

mworrell:
Indeed, though now it is possible to disable the features for categories.

In the past we had address only on the person, event and institution resources. But then we also needed it on media and articles... so we ended up enabling it by default.

The error you found was that the `feature_...` couldn't be deselected, as the save routines would strip the empty value. By using the prefix `is_` we let the sanitizer know that it should be a Boolean value, and then on empty values `false` is stored.

I did have a look at the installer code, to add better defaults to some categories. But to do that we might want to refactor that code first.

PR #2649, opened 2021-04-14
Removed server generated sid, replace with cotonic-sid.

mworrell:
### Description

Replace the `sid` as generated by the authentication code with the `cotonic-sid` as managed by Cotonic.

This fixes an issue where the `sid` could not be synchronized between MQTT and HTTP requests.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Companion Cotonic changes https://github.com/cotonic/cotonic/pull/54

PR #2659, opened 2021-05-13
New sitemap building using urlsets.

mworrell:
### Description

This adds support for sitemaps for:
- systems with many pages
- systems with external sources for pages

The sitemap is served as separate url sets.
The Zotonic page URLs are tracked in a seo_sitemap table.

TODO:
- [ ] Check with systems using external sources
- [x] Add License headers to files

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


modules/mod_seo_sitemap/templates/sitemap_urlset.tpl
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+{% for url in m.seo_sitemap.urlset[q.s][q.o][q.c] %}
+   <url>
+      <loc>{{ url.loc|escapexml }}</loc>
+      {% if url.lastmod %}<lastmod>{{ url.lastmod|date:"c" }}</lastmod>{% endif %}
+      {% if url.changefreq %}<lastmod>{{ url.changefreq|escapexml }}</changefreq>{% endif %}
```
robvandenbogaard: Shouldn't `<lastmod>` here be `<changefreq>`? (now unbalanced tag)

modules/mod_seo_sitemap/templates/sitemap_urlset.tpl
```diff
@@ -0,0 +1,11 @@
...
+      {% if url.priority|is_defined %}<lastmod>{{ url.priority|to_binary|escapexml }}</priority>{% endif %}
```
robvandenbogaard: Same here: I guess `<lastmod>` should be `<priority>`

doc/ref/modules/mod_seo_sitemap.rst
```diff
@@ -9,6 +9,59 @@ the Google bot to index your site. By default, its generated
...
+The sitemaps follow the specification at https://www.sitemaps.org/protocol.html
+
+Maintaining the sitemap index
+-----------------------------
+
+The ``seo_sitemap`` model maintains an index of all pages (and their URLs) that are publicly readable.
+
+There is a button in the System > Status menu to rebuild the sitemap index. This is needed if there
+are (big) changes in the access control affecting the publicly viewable pages or if languages are
+changed.
```
robvandenbogaard, line 21: Would it be problematic or is it difficult to cover these cases automatically? To schedule a rebuild when any ACL or language setting are changed (theoretically, a change there yields a set of affected Resources, from which the required diff on the sitemap can be derived, but we would need an incremental sitemap build infrastructure for that).

mworrell: Language changes are now quite well covered (at least in the master version).

The ACL changes I am not so sure about, as going through 1M records because _something_ changed might be a bit much...  and indeed, we don't have a method to reverse engineer what records might be affected by a change (categories, content groups, etc.)

robvandenbogaard: Yeah indeed, it seems good for now, at least good to have discussed it here, the manual reindex button is needed as a fallback anyway.

modules/mod_seo_sitemap/models/m_seo_sitemap.erl
```diff
@@ -0,0 +1,478 @@
...
+            catch _:_ -> ?URLSET_SIZE
+            end
+    end,
+    Limit2 = erlang:min(?URLSET_SIZE, Limit1),
+    seo_sitemap:urlset(Source, Offset, Limit2, Context).
+
+%% @doc Return the number of sitemap entries
+-spec count( z:context() ) -> non_neg_integer().
+count( Context ) ->
+    z_db:q1("select count(*) from seo_sitemap", Context).
```
robvandenbogaard, line 75: Shouldn't this take into account `publication_end`, language and resource visibility as noted below for `slice`?

robvandenbogaard: Ah that would not be efficient in the case of visibility because that is calculated per item.

modules/mod_seo_sitemap/models/m_seo_sitemap.erl
```diff
@@ -0,0 +1,478 @@
...
+max_id( Context ) ->
+    z_db:q1("select max(id) from seo_sitemap", Context).
+
+
+%% @doc Return a slice from the sitemap entries. Note that this might
+%% give duplicate entries in sitemaps, as the ids might change during
+%% sitemap discovery and consumption by the crawler.
+%% All locations are made into absolute urls with the correct hostname.
+%% As entries are filtered on publication_end, their language and resource
+%% visibility, this might return less entries than the Limit amount requested.
```
robvandenbogaard, line 87: Agreed that this is a workable situation. To easily prevent concurrency issues, we could use a snapshot view on the sitemap data, which would be natural practice in a versioned database. For the current implementation, the omission is ok to keep it simple.

modules/mod_seo_sitemap/models/m_seo_sitemap.erl
```diff
@@ -0,0 +1,478 @@
...
+            z_db:transaction(
+                fun(Ctx) ->
+                    lists:foreach(
+                        fun({_Lang, Loc}) ->
+                            z_db:q(
+                                "delete from seo_sitemap
+                                 where rsc_id = $1
+                                   and loc = $2
+                                   and source = 'rsc'",
+                                [ Id, Loc ],
```
robvandenbogaard, line 324: Shouldn't `lang` be included here too? Or included in the update on line 357?

modules/mod_seo_sitemap/models/m_seo_sitemap.erl
mworrell: Indeed - those are filtered out when the urlset is requested.

modules/mod_seo_sitemap/models/m_seo_sitemap.erl
mworrell: The `lang` is also assumed to be part of the url.

PR #2694, opened 2021-07-01
Move logs and data to separate directories

### LINKED ISSUE #2693: Location of data and logs
The recent update of lager moved our logs to the `log` directory in the current working dir.

Before this it was stored in `priv/logs`

After some investigation @mmzeeman and I concluded that it might be better to keep the `priv` directory read-only and only used for assets (lib, lib-src, templates, translations, etc.)

We can then move the `files` and `mnesia` directory to a `data` directory.
And all logs to a `log` directory.

The root of these directories should be configurable.

Discussion:

 - [ ] Do we want to use `logs` ?
 - [ ] Do we want to use `data` ?
 - [ ] Do we want to default the base directory for both to `.`  (the root zotonic directory or release dir when we start using releases) or to some directory outside the source tree dir?
 - [ ] Shall we make two configs?  One for the log root and one for the data root dir?

/cc @mmzeeman @ddeboer @CyBeRoni @mmzeeman

mworrell:
### Description

Fix #2693

The `priv` directory should be considered read-only.

This moves the data and logs to other places and makes those places configurable.

Configs and data (except mnesia) on old locations will be found and used.

Mnesia and log files are always moved.

- [x] **decide** on usage of `filename:basedir/2`
- [x] Use the system basedir `user_config` for security and zotonic config files
- [x] Change default erlang config for lager and mnesia
- [x] Update zotonic_core for the mnesia directory
- [x] Initialize `setup:data_dir/0` and `setup:log_dir/0` to the configured `data_dir` and `log_dir`
- [x] Add configs for data root dir in zotonic.config
- [x] Add site config for data root dir
- [x] Change `z_config` to add the new configs
- [x] Change `z_path` to use the new files directories

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Proposal, for the directories use `filename:basedir/2`

http://erlang.org/doc/man/filename.html#basedir-2

For the paths, see below.

**Good** thing is that we then store our data on the places a system is expected to store its data.
**Bad** thing is that the data is scattered all over  the system, making it harder to clean up and/or backup.

`filename:basedir(user_config, "foo").`

 * Linux: filename:basedir(user_config, "foo").
 * macOS: "/Users/marc/Library/Application Support/foo"

`filename:basedir(user_cache, "foo").`

 * Linux: "/home/zotonic/.cache/foo"
 * macOS: "/Users/marc/Library/Caches/foo"

`filename:basedir(user_data, "foo").`

 * Linux: "/home/zotonic/.local/share/foo"
 * macOS: "/Users/marc/Library/Caches/foo"

`filename:basedir(user_log, "foo").`

 * Linux: "/home/zotonic/.cache/foo/log"
 * macOS: "/Users/marc/Library/Logs/foo"

There is also `site_config` and `site_data`, but those locations (`/usr/local/share/..` and `/Library/Application Support/...`) are usually only writeable by root.

mmzeeman (approved):

PR #2718, opened 2021-08-06
Use current_timestamp instead of current_date so the time is also used in the query.

mmzeeman:
### Description

The time was not used in the query. This caused events which just finished to disappear from the results.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Thanks!

PR #2723, opened 2021-08-11
Better export/import for resource pubsub

mworrell:
### Description

This adds functionality for exporting and importing resources between Zotonic instances.

The default resource export is now using the `rsc_export` model instead of `mod_export`. This gives a more complete export. The `shortlink` link in the `<head>` redirects to the `rsc_export`.

To do:

 - [x] format of the `menu` property (it is now made into a map, should be a list)
 - [x] Map embedded ids in the received body, menu and blocks (create placeholders)
 - [x] Recursively import placeholders (option for max depth)
 - [x] Add function to make rsc authoritative (update rsc-id, add entry to rsc_gone)

This limits imports to non `meta` categories.  There are also some hard coded predicates that the importer refuses to import and the exporter refuses to export. This for security of the ACL user groups model.

Importing things into the `meta` category, and allowing ACL related edges, needs some more consideration about the options and when it should be allowed to happen by whom.

This also introduces a change to how menus are stored in the `menu` property.  Previously it was stored as a list of tuples

    {1234, [ {5678, [ ... ]}, ...}

This doesn't allow for conversion to JSON, so the structure has been changed to use records:

    #rsc_tree{ id = ...,  tree = [ ... ]}

This allows for serialization to JSON using jsxrecord.

### Checklist

- [x] documentation updated
- [x] tests added
- [ ] no BC breaks


apps/zotonic_core/src/models/m_rsc_import.erl
```diff
@@ -18,96 +17,1130 @@
...
+
+
+%% @doc Map all ids in a menu to local (stub) resources. Skip all tree entries
+%% that could not be mapped to local ids.
+map_menu([ #rsc_tree{ id = RemoteId, tree = Sub } | Rest ], UriTemplate, Acc, ImpAcc, Options, Context) ->
+    case map_id(RemoteId, UriTemplate, ImpAcc, Options, Context) of
+        {ok, {LocalId, ImpAcc1}} ->
+            {Sub1, ImpAcc2} = map_menu(Sub, UriTemplate, [], ImpAcc1, Options, Context),
+            Acc1 = [ #rsc_tree{ id = LocalId, tree = Sub1 } | Acc ],
+            map_menu(Rest, UriTemplate, Acc1, ImpAcc2, Options, Context);
```
robvandenbogaard, line 728: I was wondering why you would need to reverse the list at the [] case of map_menu; it is to avoid having to concatenate lists here, so simple prepends can be used. This is more efficient because single reverses on list ends cost less than appending several times within the lists.

apps/zotonic_core/src/models/m_rsc_import.erl
```diff
@@ -18,96 +17,1130 @@
...
+map_menu([ _ | Rest ], UriTemplate, Acc, ImpAcc, Options, Context) ->
+    map_menu(Rest, UriTemplate, Acc, ImpAcc, Options, Context);
+map_menu([], _UriTemplate, Acc, ImpAcc, _Options, _Context) ->
+    {lists:reverse(Acc), ImpAcc}.
+
+
+%% @doc Map all ids in blocks to local (stub) resources. Remove blocks that can not
+%% have their ids mapped.
+map_blocks([ B | Rest ], UriTemplate, Acc, ImpAcc, Options, Context) ->
+    OptionsNoEdges = proplists:delete(import_edges, Options),
```
robvandenbogaard, line 742: I am not sure I understand why it is called `OptionsNoEdges` while the value holds what is in `import_edges` - the tandem implies a boolean, but I'm not sure how the semantics of `import_edges = true` would match the naming of `OptionsNoEdges`; I'd guess the contrary.

apps/zotonic_core/src/models/m_rsc_update.erl
```diff
@@ -1245,16 +1250,7 @@ props_filter(<<"category_id">>, CatId, Acc, Context) ->
...
-    case m_rsc:is_a(CgId, content_group, Context)
-        orelse m_rsc:is_a(CgId, acl_collaboration_group, Context)
-    of
-        true ->
-            Acc#{ <<"content_group_id">> => CgId };
-        false ->
-            lager:error("Ignoring unknown content group '~p' in update.", [CgName]),
-            Acc
-    end;
+    props_filter(<<"content_group_id">>, CgName, Acc, Context);
```
robvandenbogaard, line 1253: I guess the code for this case could be removed because of the updates to id handling in m_rsc (m_rsc:rid is already used downstream?)

apps/zotonic_mod_menu/src/mod_menu.erl
```diff
@@ -226,88 +265,82 @@ create_new(List, Context) ->
...
     end.
 
-	validate([], Acc) ->
-		lists:reverse(Acc);
-	validate([{_M,_L} = M|Ms], Acc) ->
-		validate(Ms, [M|Acc]);
-	validate([M|Ms], Acc) ->
-		validate(Ms, [{M,[]}|Acc]).
+validate(undefined, Acc) ->
+    Acc;
```
mmzeeman, line 328: Shouldn't this be a reverse too? Like the line below.

apps/zotonic_mod_menu/src/mod_menu.erl
```diff
@@ -447,16 +461,18 @@ manage_schema(install, Context) ->
     #datamodel{
         categories = [
               {menu, categorization,
-               [{title, <<"Page menu">>}]
+                    #{
+                        <<"title">> => <<"Page menu">>
```
mmzeeman, line 465: I think that in English normally capitalisation is used for titles so "Page Menu"

apps/zotonic_mod_menu/src/mod_menu.erl
```diff
@@ -447,16 +461,18 @@ manage_schema(install, Context) ->
...
         ],
         resources = [
             {main_menu,
                 menu,
-                [
-                    {title, <<"Main menu">>},
-                    {menu, default_menu(Context)}
-                ]
+                #{
+                    <<"title">> => <<"Main menu">>,
```
mmzeeman, line 473: Same capitalisation.

mmzeeman (approved):
Looks very nice. Made some minor comments.

apps/zotonic_mod_base/src/controllers/controller_file.erl
```diff
@@ -142,11 +142,30 @@ process(_Method, _AcceptedCT, _ProvidedCT, Context) ->
...
+            Context4 = z_context:set_resource_headers(maybe_id(Context3), Context3),
+            {z_file_request:content_stream(Info, cowmachine_req:resp_content_encoding(Context4)), Context4}
     end.
 
 %%%%% -------------------------- Support functions ------------------------
 
+maybe_id(Context) ->
+    case z_context:get(?MODULE, Context) of
+        {error, _} ->
+            {false, Context};
```
mmzeeman, line 154: Erlang does not really have an "option/maybe" type.. But I think this function should either return an id or undefined. It looks like the spec from `set_resource_headers` also expects that.

mworrell: Good one.  That should have been `false` without Context.

apps/zotonic_mod_menu/src/mod_menu.erl
mworrell: Indeed, I see that the MS also uses capitals for both words.

https://www.microsoft.com/en-us/language/Search?&searchTerm=main%20menu&langID=155&Source=true&productid=undefined

apps/zotonic_mod_menu/src/mod_menu.erl
mworrell: Changed

apps/zotonic_mod_menu/src/mod_menu.erl
mworrell: Yes, probably better.

apps/zotonic_core/src/models/m_rsc_update.erl
mworrell: Indeed. Because of our more omni-present `rid` lookups this turned out to be a near duplicate of the `content_group_id` handling.

apps/zotonic_core/src/models/m_rsc_import.erl
mworrell: Because we go recursively through the lists and prepend each item the lists gets reversed in the accumulator.
That is why at the end the accumulator is reversed.

apps/zotonic_core/src/models/m_rsc_import.erl
mworrell: Yes, better would be `OptionsWithoutEdges`

mworrell: Changed it.

PR #2731, opened 2021-08-27
Move all dependencies to Hex

mworrell:
### Description

Move all dependencies to Hex so that the Zotonic apps can be published to Hex.

 - [x] Exif (now `erlang_exif`)
 - [x] Keyserver
 - [x] LetsEncrypt
 - [x] OAuth

Actions/Status:

 - [x] Publish to erlangpack erlang_exif
 - [x] OAuth - file an issue
 - [x] Copy the code we are using from LetsEncrypt to zotonic_mod_ssl_letsencrypt
 - [x] Test letsencrypt

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
OAuth is now also on Hex

mworrell:
We are running into a type issue in the `shotgun` library.

```
Line 85 Column 1: Function decode/2 will never be called
Line 136 Column 9: The pattern {'ok', Response = #{'headers':=RHeaders}} can never match the type {'error',_} | {'ok',reference() | #{'body'=>binary(), 'header'=>map(), 'status_code'=>integer()}}
```

The `header` comes from the type definition in shotgun, but it returns `headers`.

https://github.com/inaka/shotgun/issues/168

As this issue is open for 4 years, even with a fix on some branch, I propose to remove the usage of shotgun altogether.

PR #2733, opened 2021-09-02
Publish all apps as packages to Hex

### LINKED ISSUE #1718: Publish Zotonic modules to Hex.pm
1. Publish all Zotonic _core_ modules as packages to Hex.pm, so they can be used separately. This depends on #1717. I’ve already published [one example](https://hex.pm/packages/zotonic_mod_acl_user_groups). This allows users to build their own Zotonic distribution, picking the modules they need.
2. Introduce some naming convention for _third-party_ Zotonic modules. Perhaps `mod_zotonic_thing` or (probably better) `namespace_mod_zotonic_thing`, so we get `ginger_mod_zotonic_foundation`, `ginger_mod_zotonic_github`, `arjan_mod_zotonic_geo` etc. I know these names are longer (and less elegant) than we’re used to, but they have the advantage of being very clear. We can [search for these conventionally named packages on the Hex API](https://hex.pm/api/packages?search=mod_zotonic).
3. We can then replace the (probably outdated) module repository at http://modules.zotonic.com with our Hex integration, showing the module name, description, maintainer(s) and latest version. Even better, we can add a ‘Search for Zotonic modules’ in the admin that will automatically use Hex to add the module as a dependency (write to `rebar.config` deps or `zotonic.config`) and install it.

Depends on #1717.

mworrell:
### Description

Changes to prepare for publishing all core apps and modules to Hex.

Fix #1718

 - [x] Modify all .app.src files
 - [x] Ensure rebar.config files are ok
 - [x] Check VERSION file -- generated by rebar3 scripts using last found git tag
 - [x] Always sync versions of zotonic packages in the apps/*/rebar.config
 - [x] Add scripts to publish
 - [x] Add GH action to publish
 - [x] Fix interdependency between filehandler and core (z_notifier call)
 - [x] Add zotonic app that has a `debs` for all other apps
 - [x] Trial run for complete publish

Steps to publish:

 1. Compile Zotonic (this also updates VERSION)
 2. Run test and xref (prevent publication of erroneous releases)
 3. Publish in order:
     - zotonic_core
     - ...

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@ddeboer and @mmzeeman can you have a look at this?

Good starting point is the GH action: `publish.yml`
The publication scripts are in `release/*`

mmzeeman (approved):
Nice clean update to get everything available as separate deps.

PR #2735, opened 2021-09-21
Do the edge log check via a notification 

mmzeeman:
This makes sure the edge log check will be done after the transaction finishes.

### Description

The edge log check was done inside a transaction, however when edges are inserted by the user in a transaction the edge log information was not visible yet because it was set in a nested transaction. This caused edge notifications to be fired much later. Using zotonic's notification mechanism makes sure the check is done after the transaction is committed.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_edge_log_server.erl
+++ b/apps/zotonic_core/src/support/z_edge_log_server.erl
@@ -43,8 +43,7 @@
 
 %% @doc Force a check, useful after known edge operations.
 check(Context) ->
-    Name = z_utils:name_for_site(?MODULE, Context),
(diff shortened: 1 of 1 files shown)


mworrell:
With a notifier via pid we pass the whole Context, which might contain a lot of data. Maybe we can use a mfa observer, and then a cast?

On a side note, maybe we could also add context-less _pid_ observers?

mmzeeman:
Sounds like a good idea. I changed the setup to a MF observer.

mworrell (approved):

PR #2743, opened 2021-09-27
Add language, tz and peer_ip to mqtt context

### LINKED ISSUE #2742: Can't connect with MQTT client on local systems.
When you connect an MQTT client on localhost and use admin/admin as username password, zotonic crashes. This is because the peer_ip address is missing from the context when the username password is validated.

Because the peer ip is missing from the context, it can also lead to other problems. For instance with mod_ratelimit.

```erlang
08:51:16.362 [error] CRASH REPORT Process <0.12365.1> with 0 neighbours crashed with reason: {{badmap,undefined},[{erlang,map_get,[cowmachine_remote_ip,undefined],[]},{cowmachine_req,peer_ip,1,[{file,"/Users/mmzeeman/Work/git/zotonic/master/zotonic/_build/default/lib/cowmachine/src/cowmachine_req.erl"},{line,345}]},{m_identity,is_peer_allowed,1,[{file,"/Users/mmzeeman/Work/git/zotonic/master/zotonic/apps/zotonic_core/src/models/m_identity.erl"},{line,682}]},{m_identity,check_username_pw_1,3,[{file,"/Users/mmzeeman/Work/git/zotonic/master/zotonic/apps/zotonic_core/src/models/m_ident..."},...]},...]}
```

mmzeeman:
### Description

Fix #2742

Add language, timezone and peer_ip to the mqtt context.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_mqtt_sessions_runtime.erl
+++ b/apps/zotonic_core/src/support/z_mqtt_sessions_runtime.erl
@@ -137,21 +137,15 @@ set_language(Lang, Context) ->
 
 set_connect_context_options(Options, Context) ->
     Prefs = maps:get(context_prefs, Options, #{}),
-    Context2 = case maps:get(language, Prefs, undefined) of
-        undefined -> Context;
-        Language -> z_context:set_language(Language, Context)
+    Context1 = maybe_set_language(Prefs, Context),
+    Context2 = maybe_set_timezone(Prefs, Context1),
+    Context3 = z_context:set(auth_options, maps:get(auth_options, Prefs, #{}), Context2),
+    Context4 = maybe_set_peer_ip(Options, Context3),
+    Context5 = case maps:get(cotonic_sid, Prefs, undefined) of
+        undefined -> Context4;
+        Sid -> z_context:set_session_id(Sid, Context4)
     end,
-    Context3 = case maps:get(timezone, Prefs, undefined) of
-        undefined -> Context2;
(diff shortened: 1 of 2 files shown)


apps/zotonic_core/src/support/z_mqtt_sessions_runtime.erl
```diff
@@ -97,13 +107,26 @@ maybe_set_language(#{ <<"preferences">> := #{ <<"language">> := Lang } }, Contex
         error:undef ->
             z_context:set_language(Lang, Context)
     end;
+maybe_set_language(#{ language := <<>> }, Context) ->
+    Context;
+maybe_set_language(#{ language := Lang }, Context) ->
+    try
```
mmzeeman, line 113: @mworrell Is this construct still needed? It looks like the function is always defined.

mworrell: I suspect it was a protection against illegal timezones.

mworrell: Are the language/timezone not always in the preferences?

mmzeeman: mod_translation could also be switched off. But in that case mod_translation is still callable. No `{error, undef}` is returned in that case.

PR #2746, opened 2021-10-04
Add is_site_url filter.

mworrell:
### Description

Add the filter `is_site_url`.

This filter can be used to check if an URL is local to the current site.
It is used for testing redirects on authentication pages.

Move all URL related filters together in a doc section

### Checklist

- [x] documentation updated
- [x] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_context.erl
+++ b/apps/zotonic_core/src/support/z_context.erl
@@ -44,6 +44,7 @@
     prune_for_database/1,
     prune_for_scomp/1,
 
+    is_site_url/2,
     abs_url/2,
 
     pickle/1,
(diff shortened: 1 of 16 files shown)


apps/zotonic_core/src/support/z_context.erl
```diff
@@ -385,6 +386,27 @@ prune_envdata(Env) ->
         cowmachine_resp_body => <<>>
     }.
 
+
+%% @doc Check if the URL is an URL of the local site
+-spec is_site_url( undefined | string() | binary(), z:context() ) -> boolean().
+is_site_url(undefined, _Context) -> false;
+is_site_url(<<"//", _/binary>> = Url, Context) -> is_site_url_1(Url, Context);
```
robvandenbogaard, line 393: Would a `//` prefixed path ever result in a `true` in `is_site_url_1`? If so, how does it differ from a single `/` prefix?

mmzeeman: A double prefix is there to use the same protocol as the page. It can contain a different domain name. So if you page is at https://example.com/foo a url "//google.com/foo" will use https as protocol.

mworrell: The `//` is always followed by a hostname, which we then check against the hostnames known by the sites dispatcher.

mworrell: Example:

```
(zotonic@PoToi)4> z_sites_dispatcher:get_site_for_url(<<"//zotonicwww2.test/foo/bar">>). 
{ok,zotonicwww2}
```

robvandenbogaard: Aha, thanks for explaining!

PR #2757, opened 2021-10-11
Filter out empty ts_query. Fixes #2756

### LINKED ISSUE #2756: Zotonic return error 500 when searching with single unicode char
There is a problem with the ts_query when a single unicode character is entered.

Screenshot 2021-10-11 at 22 06 50

mmzeeman:
### Description

Fix #2756 

Filter out empty results from `to_tsquery/2`.

Also removed old code with a workaround for postgres versions older than 8.4. These are no longer supported.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman:
BTW... Should we support searching for emoji?

PR #2785, opened 2021-11-01
Swap the width and height of portrait oriented images

### LINKED ISSUE #2784: Media previews of images with vertical orientation are not correctly made.
The media previews of images with a vertical orientation in the exif information are not correctly made. This leads to distorted images.

The media preview in the admin of this image:

2222a1fb_5d07_4568_bc28_52c85e519609

is:

2222a1fb_5d07_4568_bc28_52c85e519609-781868595 jpg(mediaclass-admin-media 0f32931a1bec98e31996c4d06a03dbdc8c7ce6d7)

So it is rotated correctly, but its width and height are swapped.

mmzeeman:
### Description

Fix #2784

The width and height of portrait oriented images must be swapped when calculating its preview size.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_media_preview.erl
+++ b/apps/zotonic_core/src/support/z_media_preview.erl
@@ -619,8 +619,8 @@ calc_size(#{ image_height := 0 } = S) ->
 calc_size(#{ image_width := 0 } = S) ->
     calc_size(S#{ image_width => 1 });
 
-calc_size(#{ orientation := Orientation } = S) when Orientation >= 5 ->
-    calc_size(S#{ orientation => 1 });
+calc_size(#{ image_height := H, image_width := W, orientation := Orientation } = S) when Orientation >= 5 ->
+    calc_size(S#{ orientation => 1, image_height := W, image_width := H});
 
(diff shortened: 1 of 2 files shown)


mmzeeman:
After this fix, the media preview of the image mentioned in the issue is:

2222a1fb_5d07_4568_bc28_52c85e519609-706656798 jpg(mediaclass-admin-media 0f32931a1bec98e31996c4d06a03dbdc8c7ce6d7)

mworrell:
Shall we add a test case for this?

Fix looks good.

mworrell:
Or just add an issue to add test cases for the image manipulation in general.  Then we can add more than just this one.

mmzeeman:
Good one... it is quite tricky.... One thing I noticed is that the width and height of the image in case of an orientation swap is  now wrong in the meta-data of the image. I'm not sure how to handle that one.

See top of the screenshot:

Screenshot 2021-11-04 at 10 59 48

mworrell:
Technically the display is correct.... that is how the data is stored in the image.  It is just that there is an extra rotation added during resizing/display.

mmzeeman:
Check... I was indeed wondering how to approach that.

mmzeeman:
I'll include some unit tests for this later today 👍

PR #2790, opened 2021-11-09
Make it possible to add extra_args via the special with keyword in templates

mmzeeman:
### Description

Make it possible to add `extra_args` as a map in templates. The extra args can be used to add elements to the url via the page dispatcher.

Example:

```
{{ id.page_url with extra_args = %{ extra_thing: "yes" } }}
```

When `extra_thing` is used in the dispatch the value will be added to the url

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman:
We should probably document this. But where?

mworrell (approved):

PR #2799, opened 2021-11-12
Faceted search

mworrell:
### Description

Adds simple faceted search to the postgresql queries.

This uses a `pivot/facet.tpl` to generate the different facets.
The list of facets is taken from the template blocks.
A table is generated using the block definitions. A new table can be generated if  the blocks change.
At resource pivot, a record is added using the templates.

The facet table entries can also be used a search terms.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_mod_search/priv/templates/pivot/facet.tpl
```diff
@@ -0,0 +1,33 @@
...
+{# characters for their column index. Full text blocks are using    #}
+{# two columns. One for the value, and one for a normalized value.  #}
+{# A trigram index is added on the normalized column.               #}
+
+{# Date(time)s are stored as-is, no additional timezone conversion  #}
+{# is done, so be careful with the timezone used to fill the block. #}
+
+{% block category_id %}{{ id.category_id }}{% endblock %}
+
+{% block foo_ft %}{% endblock %}
```
robvandenbogaard, line 31: Maybe having an example field like country (as available in address - not sure it ) here would help understanding which type of free text fields would be a good fit (not title for instance, because it is expected to have a range of nearly unique values).

apps/zotonic_mod_search/src/support/search_facet.erl
```diff
@@ -0,0 +1,351 @@
...
+    block :: atom(),
+    type :: facet_type(),
+    is_range :: boolean()
+    }).
+
+-type facet_type() :: text
+                    | ft
+                    | int
+                    | float
+                    | dt
```
robvandenbogaard, line 32: Personally I prefer non-abbreviated names like datetime and fulltext. I had to read the rest of the code to realise I was slow on getting that dt means datetime ;)

apps/zotonic_mod_search/priv/templates/pivot/facet.tpl
mworrell: I added some examples.

apps/zotonic_mod_search/src/support/search_facet.erl
mworrell: I made the names longer.

mworrell:
@robvandenbogaard can you have another look?  I think it is now good enough to merge.

We will need to add some additional documentation describing the facets in detail.

Also need to add the faceting of the array elements. That is a bit more tricky, but very useful for keywords (where you have multiple values per resource).

mmzeeman (commented):
Nice.

mmzeeman (approved):

mworrell:
I think this can still go wrong with group-by queries. There we need a slightly different select clause to fetch the rsc.id values.

Will check into that.

mworrell:
The search_query does not produce any `group_by` statements, so that is ok.

PR #2807, opened 2021-11-17
Add support for importing RDF triples.

mworrell:
### Description

This adds an additional import format:

```erlang
#{
    <<"uri">> => <<"....">>,
   <<"rdf_triples">> => [ zotonic_rdf:rdf_triple() ]
}
```

Also allows modules to resolve URIs for fetching import data.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_admin/src/support/z_admin_rsc_import.erl
+++ b/apps/zotonic_mod_admin/src/support/z_admin_rsc_import.erl
@@ -32,8 +32,7 @@ event(#postback{ message={import_refresh, Args} }, Context) ->
         {ok, {_Id, _ObjectIds}} ->
             case proplists:get_all_values(on_success, Args) of
                 [] ->
-                    Context1 = z_render:wire(OnError, Context),
-                    z_render:growl(?__("Succesfully imported page from the remote server.", Context1), Context1);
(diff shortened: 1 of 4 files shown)


apps/zotonic_mod_admin/src/support/z_admin_rsc_import.erl
```diff
@@ -26,16 +26,19 @@
...
 event(#postback{ message={import_refresh, Args} }, Context) ->
+    OnError = proplists:get_value(on_error, Args),
     {id, Id} = proplists:lookup(id, Args),
     case m_rsc_import:reimport_recursive_async(Id, Context) of
         {ok, {_Id, _ObjectIds}} ->
             case proplists:get_all_values(on_success, Args) of
                 [] ->
-                    z_render:growl(?__("Succesfully imported page from the remote server.", Context), Context);
+                    Context1 = z_render:wire(OnError, Context),
+                    z_render:growl(?__("Succesfully imported page from the remote server.", Context1), Context1);
```
robvandenbogaard, line 36: Should it be an error or success in this case - the onerror seems to contradict the growl.

mworrell: Ooop, that is indeed an ok!

PR #2828, opened 2021-12-13
Fix an issue where the live scomp could not use client topics.

mworrell:
### Description

The topics for the wires should be seen from the client, for server topics add a 'origin/bridge/'.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Nice that the topics are now generated as local to the page.

PR #2829, opened 2021-12-13
Use native date/time pickers in the admin.

mworrell:
### Description

This removes the jQuery date/time pickers and replaces them with native HTML5 date and time inputs.

Problem: when an input is empty the current date or time is displayed in light gray. I would prefer it to be empty in those cases.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Nice that we can use the native time pickers now.

PR #2833, opened 2021-12-23
Faceted search with alternative counts per facet.

mworrell:
### Description

This changes the `facets` search, so that for each facet the count for the other values of the facet is shown.

The faceted search where the counts per facet in the result set (_drill down_) are shown is now called `subfacets`

For this change the search query handling has been modified.

This gives a BC break for the pivot search.

The `custompivot` search term has been removed.
Instead of this refer in the `filter` to a custom pivot with `pivotname` as:  `pivot.pivotname.field`
Pivot fields in the `rsc` table can be referred as: `pivot.field`
Facet fields can be referred as: `facet.field`

There is are new search terms:

 * `filter.pivot.pivotname.field`
 * `pivot.pivotname.field`
 * `pivot.field`
 * `filter.facet.field`

These filter terms accept the same kind of values as the `facet.field` term.

The values are coerced into the correct type for the referred column.

The difference with the `facet.field` terms is that the latter are used for the counts per facet. The former `filter.facet.field` restricts the rows for all facet counts.

TODO:

 * [x] Check tests
 * [x] Update documentation

### Checklist

- [x] documentation updated
- [ ] tests added
- **BC breaks**


mmzeeman (approved):
Very nice functionality upgrade, and one which also still lets it stay SQL instead of something like an or-mapping.

mworrell:
Indeed, quite some SQL options (array, full text, etc.) are not supported by those SQL-query-builders.  This is a middle-of-road approach.

PR #2847, opened 2022-01-19
mod_filestore: fix a problem with 'busy' sql connections

mworrell:
### Description

The filestore is querying the database during the start of the uploader.
If the database is slow then the supervisor might experience a timeout during start_link.

Somehow the database connection is reused and the exact same query is retried on the 
same connection process. Why is still unclear to me and needs further investigation.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/db/z_db.erl
+++ b/apps/zotonic_core/src/db/z_db.erl
@@ -309,7 +309,8 @@ get_connection(#context{dbc=undefined} = Context) ->
     case has_connection(Context) of
         true ->
             set_dbtrace_flag(Context),
-            z_db_pool:get_connection(Context);
+            DbDriver = z_context:db_driver(Context),
+            DbDriver:pool_get_connection(Context);
         false ->
             {error, nodatabase}
     end;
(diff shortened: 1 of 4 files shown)


mmzeeman (approved):
Interesting. Does this happen in other area's of zotonic?

mworrell:
There was also a report of a 'busy' connection in the pivot process.

My theory of that one is that it was a timeout in one query, and as queries are done in a transaction, that the next queries reused the same transaction connection for the next queries. Have to check the code though, to see how a query timeout is handled in that code.

mworrell:
I am still at a loss how the connection of the long-running can be reused by the file-uploader supervisor to start the next uploader.

The connection is not stored in the context, as it is only used by the `z_db:with_connection/2` I can't really see how it ends up being used for another query.

The connection process is checked out from poolboy, so can't be checked out again. And there isn't a transaction that could provide the connection (like with the pivot routines).

mworrell:
Our working hypothesis is now that a busy connection was checked into the pool and then re-used.

This is now circumvented by letting the worker do the checkin, so that it can refuse the checkin iff the worker is busy.

PR #2849, opened 2022-01-20
core: optimize memory usage

### LINKED ISSUE #2848: Memory usage - default number of processes / atoms
Currently we have these defaults:

 * Processes: 10M
 * Ports: 100K
 * Atoms: 1M

I think the process and atom limits are a bit high.
We could put them lower and use less memory.

What would be good defaults?

In general we should check if there are processes holding memory when they don't need to.

/cc @mmzeeman @arjan @CyBeRoni

mworrell:
### Description

Fix #2848

Lower memory use.

Points are added below depending on the results of observer and code review.

 - [x] Hibernate trans server after loading translations
 - [x] Change mod_search to use pid_observe_ for watches
 - [x] Let z_sites_dispatcher hibernate after updates
 - [x] Let module_indexer force gc after index action
 - [x] Check memory use of z_module_sup (force gc?)
 - [x] Check memory use of z_module_manager (force gc?)
 - [x] Check max processes / atoms

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/i18n/z_trans_server.erl
+++ b/apps/zotonic_core/src/i18n/z_trans_server.erl
@@ -118,7 +118,7 @@ handle_cast({load_translations, Trans}, State) ->
     List = maps:fold(F, [], Trans),
     sync_to_table(List, State#state.table),
     z_template:reset(State#state.site),
-    {noreply, State};
+    {noreply, State, hibernate};
 
 %% @doc Trap unknown casts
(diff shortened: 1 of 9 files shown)


mworrell:
Before changes:

Screenshot 2022-01-20 at 16 00 06

After changes:

Screenshot 2022-01-20 at 15 58 16

PR #2850, opened 2022-01-21
mod_admin: add task queue status to admin/status

mworrell:
### Description

This also moves the error_count column from the task props to a database column.

Screenshot 2022-01-21 at 09 52 50

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/install/z_install_update.erl
+++ b/apps/zotonic_core/src/install/z_install_update.erl
@@ -258,6 +258,7 @@ upgrade(C, Database, Schema) ->
     ok = publication_start_nullable(C, Database, Schema),
     ok = key_changes_v1_0(C, Database, Schema),
     ok = rsc_language(C, Database, Schema),
+    ok = task_queue_error_count(C, Database, Schema),
     ok.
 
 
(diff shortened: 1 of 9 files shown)


apps/zotonic_core/src/install/z_install_update.erl
```diff
@@ -734,8 +735,48 @@ rsc_language(C, Database, Schema) ->
...
+                fun({TaskId, Props}) ->
+                    case task_error_ct(z_db_pgsql:decode_value(Props)) of
+                        N when is_integer(N), N > 0 ->
+                            {ok, _} = epgsql:equery(C, "
+                                            update pivot_task_queue
+                                            set error_count = $1
+                                            where id = $2
+                                            ",
+                                            [ N, TaskId ]);
+                        _ ->
```
robvandenbogaard, line 769: Nice to keep track of error count per task

robvandenbogaard, line 769: Nice to keep track of error count per task

apps/zotonic_core/src/install/z_install_update.erl
```diff
@@ -734,8 +735,48 @@ rsc_language(C, Database, Schema) ->
 
             {ok, _} = epgsql:equery(C, "
                             insert into pivot_task_queue (module, function, key)
-                            values ('z_install_update_task', 'init_language', $1)
+                            values ('z_install_update_task', 'init_language', '')
                             ",
-                            [ z_ids:id() ]),
+                            []),
```
robvandenbogaard, line 740: Will the key be inserted at an other point?

mworrell: There is an unique SQL index per module/function/key.
In this case there can only be a single task for this, so we can leave the key empty.

mworrell:
```sql
    CREATE TABLE pivot_task_queue
    (
        id serial NOT NULL,
        module character varying(80) NOT NULL,
        function character varying(64) NOT NULL,
        key character varying(100) NOT NULL DEFAULT ''::character varying,
        due timestamp with time zone ,
        props bytea,
        error_count integer not null default 0,

        CONSTRAINT pivot_task_queue_pkey PRIMARY KEY (id),
        CONSTRAINT pivot_task_queue_module_function_key_key UNIQUE (module, function, key)
    )
```

PR #2882, opened 2022-02-14
core: add support for img decoding attribute.

mworrell:
### Description

Adding `decoding="async"` makes the page load faster, as sync img decoding is disabled.

This also solves an issue with some Safari versions where only the top part of an image is shown.

More changes:

 - Add prefix `mediaclass-` to the mediaclass when added the img class attribute
 - Add `rotate` to the list of known img options
 - Fix a problem where the mediaclass was not added as a css class if there was an alt attribute defined

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Looks like good idea.

PR #2921, opened 2022-03-17
use https to access github after disabling git protocol

CyBeRoni:
### Description

GitHub has disabled the git:// protocol for access, breaking builds. This PR swaps all rebar dependencies to https.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Thanks!

PR #2941, opened 2022-04-13
Fix syntax error in zotonic.config.in

cdfa:
### Description

Fix syntax error in `zotonic.config.in`.

### Checklist

- [x] documentation updated
- [x] tests added
- [x] no BC breaks

### DIFF
--- a/priv/zotonic.config.in
+++ b/priv/zotonic.config.in
@@ -123,10 +123,6 @@
    %% {clamav_port, 3310},
    %% {clamav_max_size, 26214400},
 
-%%% Password for the sites administration site (zotonic_status). Will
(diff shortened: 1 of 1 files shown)


mworrell:
That `,` isn't there as it is the last term in the list.
All other terms are commented out.

An option could be to move the `,` before all terms, so that uncommenting them would not introduce a syntax error.

cdfa:
Ahh, so the syntax error was caused by uncommenting deps. I like the style with leading `,`, but I think the issue would also be resolved if the key is moved to the end of the list.

mworrell:
Thanks  :-)  En gemerged!

PR #2946, opened 2022-04-14
core: support export/import of ids in nested values.

mworrell:
### Description

This adds export and import of resource id references in nested values.

For example when a resource id is referred to (with an `_id` key) in a list or map.

Also upgrades zotonic_stdlib to sanitize keys like `foo_id_list`, so that it will handle all items in the list as `id` values (eg. resource references).

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_core/src/models/m_rsc_import.erl
```diff
@@ -953,6 +1000,15 @@ cleanup_map_ids(RemoteRId, Rsc, UriTemplate, ImportedAcc, Options, Context) ->
     end,
     {Rsc5, ImportedAcc1}.
 
+%% @doc Check if the value looks like an exported id description.
+is_id_map(#{
+        <<"id">> := Id,
+        <<"uri">> := Uri,
+        <<"is_a">> := [ Cat |_]
```
vkatsuba, line 1007:
```suggestion
        <<"is_a">> := [ Cat |_ ]
```

apps/zotonic_core/rebar.config
```diff
@@ -44,7 +44,7 @@
     {filezcache, "1.1.0"},
     {yamerl, "0.8.1"},
 
-    {zotonic_stdlib, "~> 1.6.0"},
+    {zotonic_stdlib, "~> 1.6"},
```
vkatsuba, line 47: Was downloaded `1.7.0` - maybe make sense to set here version `1.7`.

mworrell: Yes we want 1.7 as the minimum version.

PR #2948, opened 2022-04-15
mod_base: on API GET call, pass query args as the payload. 

### LINKED ISSUE #2947: On search model API call, pass http query arguments as the search terms
Now you can pass query arguments in the MQTT call like this:

```javascript
cotonic.broker.call("bridge/origin/model/search/get/query", { text: "rationale" } ).then((m) => console.log(m));
```

This is not possible with the HTTP call.

/cc @vkatsuba

mworrell:
### Description

On HTTP GET API calls, pass the Qs args as props in the payload.

Also suppress the "zotonic" arguments when parsing POST forms as payload.

Fixes #2947

API Calls like `curl -i -k 'https://zotonicwww2.test:8443/api/model/search/get/query?text=rationale'` now work.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_context.erl
+++ b/apps/zotonic_core/src/support/z_context.erl
@@ -81,6 +81,7 @@
     get_q_all/2,
     get_q_all_noz/1,
     get_q_validated/2,
+    get_q_map/1,
     set_q_all/2,
 
     q_upload_keepalive/2,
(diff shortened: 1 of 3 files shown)


mmzeeman (commented):
What is the '*' argument?

I know, it is only three lines, but is it possible to have a logical name for creating a map from a query string in one function and use that in both locations? Maybe even a function of z_context? `z_context:get_q_map` or something?

vkatsuba (approved):

mworrell:
The * argument is the matching path in the dispatch rule (where you can also write '*')

In the API this is the topic for the API call.

mmzeeman (approved):

PR #2949, opened 2022-04-19
mod_search: add the search term 'id', to limit result to one or more ids

mworrell:
### Description

This completes the `id_exclude` with a search term `id`.

Useful when mapping textual searches like `"id:1234" to query terms.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_search/src/support/search_query.erl
+++ b/apps/zotonic_mod_search/src/support/search_query.erl
@@ -160,6 +160,7 @@ request_arg(<<"filter.facet.", F/binary>>)-> {facet, F};
 request_arg(<<"filter.", F/binary>>)   -> {filter, F};
 request_arg(<<"pivot.", _/binary>> = F)-> {filter, F};
 request_arg(<<"pivot_", F/binary>>)    -> {filter, <<"pivot.", F/binary>>};
+request_arg(<<"id">>)                  -> id;
 request_arg(<<"id_exclude">>)          -> id_exclude;
 request_arg(<<"hasobject">>)           -> hasobject;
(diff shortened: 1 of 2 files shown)


apps/zotonic_mod_search/src/support/search_query.erl
```diff
@@ -312,6 +325,33 @@ qterm({id_exclude, Id}, Context) ->
...
+        ],
+        args = [ RscIds ]
+    };
+qterm({id, Id}, Context) ->
+    case m_rsc:rid(Id, Context) of
+        undefined ->
+            [];
+        RscId ->
+            #search_sql_term{
+                where = [ <<"rsc.id = ">>, '$1'],
```
vkatsuba, line 351:
```suggestion
                where = [ <<"rsc.id = ">>, '$1' ],
```

apps/zotonic_mod_search/src/support/search_query.erl
```diff
@@ -300,8 +301,20 @@ qterm({content_group, ContentGroup}, Context) ->
 qterm({id_exclude, Ids}, Context) when is_list(Ids) ->
     %% id_exclude=resource-id
     %% Exclude an id or multiple ids from the result
-    Es = lists:map(fun(Id) -> {id_exclude, Id} end, Ids),
-    qterm(Es, Context);
+    RscIds = lists:filtermap(
+        fun(Id) ->
```
vkatsuba, line 305: This function and at line 331 - look the same, maybe it should be moved to a separate function or macro

PR #2953, opened 2022-05-03
[Translation] Add icon options on the language switch toggle

DorienD:
Add icon options on language switch template.


apps/zotonic_mod_translation/priv/templates/_language_switch.tpl
```diff
@@ -1,10 +1,11 @@
+{% with icon_after|default:'<i class="caret"><span class="sr-only">{_ Open language menu _}</span></i>' as icon_after %}
```
mworrell, line 1: I think we need to change that `{_ ... _}`, as it won't be picked up by the translation module. It is now a string inside a string of the template language.

We can make the entire string translatable:

```
default:_'<i class="caret"><span class="sr-only">Open language menu</span></i>'
```

PR #2954, opened 2022-05-03
FIX: Corrected -p option on addsite

alvaropag:
### Description
-p option was setting dbhost instead of dbport

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Good catch, thanks!

PR #2955, opened 2022-05-04
Fix addsite default skel and docs

williamthome:
### Description

The docs says "Skeleton site (one of ‘blog’, ‘empty’, ‘nodb’; default: blog)", but when a site is created with no options the default one is "empty". This PR fixes this using the already defined ?SKEL macro, but changed from atom to string.
So with this PR, running addsite command without options the site skel will be blog.
Also fixed the addsite docs. The doc says that the default schema is "public", but running the addsite command the default one is <site_name>.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Thanks!

PR #2956, opened 2022-05-05
Update cookbook-custom-model docs

williamthome:
### Description

Looking at `zotonic_model` and docs, I found some mismatches.
Maybe `Keys` should be `[atom() | binary()]` instead of just `list()`? Or what's the best type? Maybe [pos_integer() | binary()]? Or can be [term()]?

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Due to template-model lookup the path can also contain arbitrary data. For example the search model accepts something like this:

```django
{% print m.search.paged[{query cat=`news`}] %}
```

Here the tuple `{query, [{cat, news}]}` is passed in the path.

PR #2957, opened 2022-05-05
Use init:stop to make sure the zotonic node shuts down cleanly

mmzeeman:
### Make sure started applications are stopped before shutting down the node

`erlang:halt(0)` immediately stops the node, without waiting for applications to stop
cleanly. Applications which need to do teardown work before stopping will not do this. The `init` application, which coordinates system startup, has a stop function which cleanly handles shutting down the node, and manage heart when needed. See: https://www.erlang.org/doc/man/init.html#stop-1

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
We added the rather brutal kill as the `init:stop()` tries to unload beam files, which takes very long on some systems. This seems to be fixed in OTP25 (as you mentioned).

Maybe we can stop all applications instead?

mmzeeman:
Something like?

``` [ application:stop(App) || {App, _, _} <- application:which_applications() ],```

**Comment by mworrell** [comment]
Maybe, I guess we don't have a dependency graph for a clean shutdown?

**Comment by mmzeeman** [comment]
Here is the original issue regarding the unloading behaviour. I guess we had this problem even on faster computers because zotonic loads a lot of modules. https://github.com/erlang/otp/issues/5031

**Comment by mworrell** [comment]
Yes, I got some complaints from people, especially on non MacBook Pro machines ...

**Comment by mmzeeman** [comment]
I don't know if the application module maintains some sort of order.  It could be that they just return the applications in reverse order they where started. That would be good I guess.

**Comment by mworrell** [comment]
Looking at the application list it seems to be that way.

File: apps/zotonic_launcher/src/zotonic.erl [hunk 1]
```diff
@@ -112,9 +112,25 @@ stop() ->
...
+    % Stop all other running applications. Note: on OTP 25 we can simply use init:stop(0).
+    % On earlier OTP versions this is problematic because all modules are unloaded before
+    % the node is stopped. This can take a long time.
+    application:stop(zotonic_launcher),
+    application:stop(exometer),
+    application:stop(jobs),
+    application:stop(mnesia),
+    application:stop(epgsql),
+
+    [ application:stop(A) || {A, _, _} <- application:which_applications() ],
```
mworrell, line 128: Should we first stop the other applications and then the base apps like exometer, jobs and mnesia?

It might be that the other apps want to save something in the databases.

mmzeeman: Ah, yes you are right. I noticed the order is important. When it is wrong it can lead to unexpected hangs. Maybe we should just use `init:stop/0`. OTP 25 is just around the corner for anybody who complains about slowness.

mworrell: It is _really_ slow.

We might also use `init:stop/0` if the OTP version is >= 25.

And the other shutdown on older versions.

mworrell:
@mmzeeman can you check my latest change?

mmzeeman:
It is reporting a lot of errors.

```
(zotonic@Tukkal)2> zotonic:stop().
2022-05-23 10:50:05 ERROR <0.1917.0> [:] ▸ text="Error in process <0.1917.0> on node zotonic@Tukkal with exit value:\
{badarg,[{ets,lookup,\
              ['m:z_depcache$site_cafe',333],\
              [{error_info,#{cause => id,module => erl_stdlib_errors}}]},\
         {depcache,get_concurrent,5,\
                   [{file,\"/Users/mmzeeman/Work/git/zotonic/master/zotonic/_build/default/lib/depcache/src/depcache.erl\"},\
                    {line,630}]},\
         {depcache,get_ets,2,\
                   [{file,\"/Users/mmzeeman/Work/git/zotonic/master/zotonic/_build/default/lib/depcache/src/depcache.erl\"},\
                    {line,350}]},\
         {depcache,get,3,\
                   [{file,\"/Users/mmzeeman/Work/git/zotonic/master/zotonic/_build/default/lib/depcache/src/depcache.erl\"},\
                    {line,265}]},\
...
```

and stopping again, after it has already stopped.

```
(zotonic@Tukkal)3> zotonic:stop().
** exception error: bad argument
     in function  erlang:send/2
        called as erlang:send(heart,{<0.1993.0>,set_cmd,"echo ok"})
        *** argument 1: invalid destination
     in call from heart:set_cmd/1 (heart.erl, line 109)
     in call from zotonic:stop/0 (/Users/mmzeeman/Work/git/zotonic/master/zotonic/apps/zotonic_launcher/src/zotonic.erl, line 120)
```

Will look at it later. The order in which things are taken down is probably wrong.

mmzeeman:
Hmm... strange, now it won't stop at all.

```
> ./bin/zotonic stop
Stopping zotonic zotonic@Tukkal ................................. Still not stopped after 30 seconds
```

mworrell:
Maybe we should just do the brute force stop on OTP <= 24, and the nice stop on OTP 25+

mworrell:
@mmzeeman Now that we have moved on to new(er) OTP versions, shall we check your initial fix again?

mmzeeman:
Yes, it helps with properly shutting down erlang applications. I have an experimental site stats tracker which uses duckdb (it has a nice interface for bulk appending data). At the moment it does not properly shut down, so all data stays in the write-ahead-log instead of in the main database.

mworrell:
@mmzeeman Only problem I see now is that `erlexec` gives a crash when shutting down.

I think the app shutdown is also waiting for all the children of the exec supervisor, and then does a hard kill.

```
2024-07-03 16:26:54 ERROR <0.436.0> [supervisor:do_terminate/2:913] ▸ report=[{supervisor,{local,exec_app}},{errorContext,shutdown_error},{reason,killed},{offender,[{pid,<0.437.0>},{id,exec},{mfargs,{exec,start_link,[[]]}},{restart_type,permanent},{significant,false},{shutdown,10000},{child_type,worker}]}] label={supervisor,shutdown_error}
```

Could there be a way for a nice (and quick) shutdown of erlexec?

PR #2959, opened 2022-05-05
Change controller_template to not search for a rsc

### LINKED ISSUE #2958: Issue with id in controller_template
Hi!

The [controller_template:process/4](https://github.com/zotonic/zotonic/blob/cc382f408517309034a28a59bf242aea827d2af6/apps/zotonic_mod_base/src/controllers/controller_template.erl#L60) always checks for a rsc_id. My question is, why?
Shouldn't this be only valid for controllers like [controller_id](https://github.com/zotonic/zotonic/blob/master/apps/zotonic_mod_base/src/controllers/controller_id.erl) or [controller_page](https://github.com/zotonic/zotonic/blob/master/apps/zotonic_mod_base/src/controllers/controller_page.erl)?
I'm using custom database tables and I don't wanna check for the id in rsc table in the controller template.
I created a custom controller as a workaround, but I think the controller_template does not need to check for an id.

williamthome:
### Description

Closes #2958

The controller_template:process/4 always checks for an rsc by id.
I think this is not a job for this controller, instead, `controller_id` can be used.
The id check in rsc table breaks a custom database implementation.
And, if the rsc is not found, a warning is displayed with the message

> "Failed to get page url path. Is the 'page' dispatch rule missing?".

This PR removes the rsc check and the warning does not appear anymore.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_mod_base/src/controllers/controller_template.erl
+++ b/apps/zotonic_mod_base/src/controllers/controller_template.erl
@@ -58,17 +58,11 @@ is_authorized(Context) ->
 
 
 process(_Method, _AcceptedCT, _ProvidedCT, Context) ->
-    Id = z_controller_helper:get_id(Context),
-    Context0 = z_context:set_noindex_header(m_rsc:p_no_acl(Id, seo_noindex, Context), Context),
-    Context1 = z_context:set_resource_headers(Id, Context0),
-    Context2 = set_optional_cache_header(Context1),
-    Template = z_context:get(template, Context2),
-    Vars = [
-        {id, Id}
-        | z_context:get_all(Context2)
-    ],
-    Rendered = z_template:render(Template, Vars, Context2),
-    z_context:output(Rendered, Context2).
+    Context1 = set_optional_cache_header(Context),
+    Template = z_context:get(template, Context1),
+    Vars = z_context:get_all(Context1),
+    Rendered = z_template:render(Template, Vars, Context1),
+    z_context:output(Rendered, Context1).
 
 set_optional_cache_header(Context) ->
     case z_context:get(max_age, Context) of


mworrell:
I have checked some existing projects (from various companies), and there this "feature" is used to pass an `id` to a template without giving an error (404) if that id is not present.  The `id` is then passed in the controller options.

If we would only fetch the id from the controller options, then I think we still support existing systems and don't get in a conflict with query args.

williamthome:
Another workaround, besides creating a custom controller, is to use any other value than `id` as params in the dispatch, but I think this isn't a good one.

mworrell:
I propose to just get the (optional) id from the controller_template dispatch rule options, and ignore any other inputs.

williamthome:
But think I have a table named foo that isn't linked to the rsc table, my dispatch rules are '{{foo, ["foo", id], controller_template, [{template, "foo.tpl"}]}}'. The controller_template will check for the rsc, but in this case I don't wanna the rsc, but the foo data. If any rsc has the foo id, the page title is changed to the rsc category. I haven't checked, but maybe more things are changed if any rsc is found. How can I ignore this? Maybe then a new controller should be created?
My apologies if I didn't understand your proposal.

alvaropag:
We used id on the dispatch rule with controller_template without knowing it would be checked against rsc.
We could change the id to foo_id, but this would require us to change all the templates that use it (and break a lot of places in the system)
Maybe we should add an explicit ignore to the options? Something like [{rsc_id_check, false}]? 
`{{foo, ["foo", id], controller_template, [{template, "foo.tpl"}, {rsc_id_check, false} ]}}` ?
Just my 2 cents.

mworrell:
In this example, "id" is in the path, so is a query argument.
That would not be handled in my proposal.

Only the id in the options would, for example:

```
{foo, [ "foo" ], controller_template, [ {id, page_foo} ]}
```

mworrell:
I have an alternative change here:  https://github.com/zotonic/zotonic/pull/2960

Could you check that against your code?

mworrell:
Merged alternative in https://github.com/zotonic/zotonic/pull/2960

williamthome:
Works like a charm! Thanks for this!
My apologies for the late response.

mworrell:
Thanks for your contributions to make Zotonic better!

PR #2963, opened 2022-05-11
Fix OAuth module name in site config

williamthome:
### Description

Using the `addsite` command, I got an error about OAuth for Facebook. Checking the active modules, the OAuth was disabled. These changes fix this by changing the OAuth module name in the `zotonic_site.config` from `mod_oauth` to `mod_oauth2`, according to the [zotonic_mod_oauth2](https://github.com/zotonic/zotonic/tree/master/apps/zotonic_mod_oauth2).

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Good one, we indeed replaced the 0.x mod_oauth in 1.x with mod_oauth2.

Thanks!

PR #2968, opened 2022-05-13
core: ensure To email is properly formatted.

mworrell:
### Description

Also add name of signup user to To for emails.

This fixes an issue with improperly formatted `To` addresses in verify emails.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_signup/mod_signup.erl
+++ b/modules/mod_signup/mod_signup.erl
@@ -266,7 +266,9 @@ send_verify_email(UserId, Ident, Context) ->
         {email, Email},
         {verify_key, Key}
     ],
-    z_email:send_render(Email, "email_verify.tpl", Vars, z_acl:sudo(Context)),
+    {Name, _NameCtx} = z_template:render_to_iolist("_name.tpl", [{id, UserId}], z_acl:sudo(Context)),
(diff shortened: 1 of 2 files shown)


modules/mod_signup/mod_signup.erl
```diff
@@ -266,7 +266,9 @@ send_verify_email(UserId, Ident, Context) ->
         {email, Email},
         {verify_key, Key}
     ],
-    z_email:send_render(Email, "email_verify.tpl", Vars, z_acl:sudo(Context)),
+    {Name, _NameCtx} = z_template:render_to_iolist("_name.tpl", [{id, UserId}], z_acl:sudo(Context)),
+    To = z_email:combine_name_email(z_string:trim(iolist_to_binary(Name)), Email),
```
vkatsuba, line 270: If `Name` will have by some reason unicode `iolist_to_binary` will be crashed.

mworrell: In the 0.x all strings are utf8 byte binaries or lists.  So we can easily serialize.

mworrell: Also the "iolist" in the function name should be "iodata"....  but alas, history.

vkatsuba (approved):

PR #2969, opened 2022-05-13
mod_signup: use the name of the person for the To in the verify email.

mworrell:
### Description

This for a lower spam score and a nicer recipient in the received email.

If an email to an resource `Id` and then let the z_email routines add the name and email address of the recipient to the email.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Handig

PR #2970, opened 2022-05-13
Site application proposal

williamthome:
### Description

Add "-S" as a boolean option to the `addsite` command (default false). If true, `<site>_app.erl` (application behaviour) and `<site>_sup.erl` (supervisor behaviour) files are included in the site dir, and, when the site starts, the application and root supervisor starts to.
e.g. 
```
bin/zotonic addsite -S true foo
```

Running `observer:start()` in the zotonic shell the site appears in the applications list:
image

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
@@ -45,7 +45,7 @@ usage() ->
     io:format(" -d <name>     Database name (default: ~s) ~n", [ z_config:get(dbdatabase) ]),
     io:format(" -n <schema>   Database schema (defaults to <site_name>) ~n"),
     io:format(" -a <pass>     Admin password~n"),
-    io:format(" -A <app>      If true, initializes a site app and a root app when the site starts (default: ~p)~n~n", [?APP]).
+    io:format(" -A <app>      If true, initializes a site app and a root supervisor when the site starts (default: ~p)~n~n", [?APP]).
 
 run(Args) ->
     case zotonic_command:get_target_node() of
--- a/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
@@ -192,11 +192,20 @@ parse_args([ "-n", Schema | Args ], Acc) ->
     parse_args(Args, Acc#{ dbschema => Schema });
 parse_args([ "-a", Pw | Args ], Acc) ->
     parse_args(Args, Acc#{ admin_password => Pw });
-parse_args([ "-A", "true" | Args ], Acc) ->
-    parse_args(Args, Acc#{ app => true });
-parse_args([ "-A", "false" | Args ], Acc) ->
-    parse_args(Args, Acc#{ app => false });
+parse_args([ "-A", App | Args ], Acc) ->
+    case re:run(App, "^(true|false|1|0)$", [caseless, {capture, none}]) of
+        match ->
+            Bool = string_to_boolean(string:to_lower(App)),
+            parse_args(Args, Acc#{ app => Bool });
+        nomatch ->
+            {error, App}
+    end;
 parse_args([ "-" ++ _ = Arg | _ ], _Acc) ->
     {error, Arg};
 parse_args(Rest, Acc) ->
     {ok, {Acc, Rest}}.
+
+string_to_boolean("true") -> true;
+string_to_boolean("false") -> false;
+string_to_boolean("1") -> true;
+string_to_boolean("0") -> false.
--- a/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
@@ -27,25 +27,25 @@
 ]).
 
 -define(SKEL, "blog").
--define(SUP, false).
+-define(APP, false).
 
 info() ->
     "Add a site and install database".
 
 usage() ->
     io:format("Usage: zotonic addsite [options] <site_name> ~n~n"),
-    io:format(" -s <skel>          Skeleton site (one of 'blog', 'empty', 'nodb'; default: ~s~n", [?SKEL]),
-    io:format(" -H <host>          Site's hostname (default: <site_name.test>) ~n"),
-    io:format(" -L                 Create the site in the current directory and add a symlink to zotonic app_user~n"),
-    io:format(" -G <url>           Clone from this Git url, before copying skeleton~n"),
-    io:format(" -h <host>          Database host (default: ~s) ~n", [ z_config:get(dbhost) ]),
-    io:format(" -p <port>          Database port (default: ~p) ~n", [ z_config:get(dbport) ]),
-    io:format(" -u <user>          Database user (default: ~s) ~n", [ z_config:get(dbuser) ]),
-    io:format(" -P <pass>          Database password (default: ~s) ~n", [ z_config:get(dbpassword) ]),
-    io:format(" -d <name>          Database name (default: ~s) ~n", [ z_config:get(dbdatabase) ]),
-    io:format(" -n <schema>        Database schema (defaults to <site_name>) ~n"),
-    io:format(" -a <pass>          Admin password~n"),
-    io:format(" -S <supervisor>    If true, initializes a site app and a root supervisor when the site starts (default: ~p)~n~n", [?SUP]).
+    io:format(" -s <skel>     Skeleton site (one of 'blog', 'empty', 'nodb'; default: ~s~n", [?SKEL]),
+    io:format(" -H <host>     Site's hostname (default: <site_name.test>) ~n"),
+    io:format(" -L            Create the site in the current directory and add a symlink to zotonic app_user~n"),
+    io:format(" -G <url>      Clone from this Git url, before copying skeleton~n"),
+    io:format(" -h <host>     Database host (default: ~s) ~n", [ z_config:get(dbhost) ]),
+    io:format(" -p <port>     Database port (default: ~p) ~n", [ z_config:get(dbport) ]),
+    io:format(" -u <user>     Database user (default: ~s) ~n", [ z_config:get(dbuser) ]),
+    io:format(" -P <pass>     Database password (default: ~s) ~n", [ z_config:get(dbpassword) ]),
+    io:format(" -d <name>     Database name (default: ~s) ~n", [ z_config:get(dbdatabase) ]),
+    io:format(" -n <schema>   Database schema (defaults to <site_name>) ~n"),
+    io:format(" -a <pass>     Admin password~n"),
+    io:format(" -A <app>      If true, initializes a site app and a root app when the site starts (default: ~p)~n~n", [?APP]).
 
 run(Args) ->
     case zotonic_command:get_target_node() of
@@ -165,7 +165,7 @@ parse(Args) when is_list(Args) ->
     Options = #{
         hostname => undefined,
         skeleton => ?SKEL,
-        supervisor => ?SUP
+        app => ?APP
     },
     parse_args(Args, Options).
 
@@ -192,10 +192,10 @@ parse_args([ "-n", Schema | Args ], Acc) ->
     parse_args(Args, Acc#{ dbschema => Schema });
 parse_args([ "-a", Pw | Args ], Acc) ->
     parse_args(Args, Acc#{ admin_password => Pw });
-parse_args([ "-S", "true" | Args ], Acc) ->
-    parse_args(Args, Acc#{ supervisor => true });
-parse_args([ "-S", "false" | Args ], Acc) ->
-    parse_args(Args, Acc#{ supervisor => false });
+parse_args([ "-A", "true" | Args ], Acc) ->
+    parse_args(Args, Acc#{ app => true });
+parse_args([ "-A", "false" | Args ], Acc) ->
+    parse_args(Args, Acc#{ app => false });
 parse_args([ "-" ++ _ = Arg | _ ], _Acc) ->
     {error, Arg};
 parse_args(Rest, Acc) ->
--- a/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
@@ -195,7 +195,7 @@ parse_args([ "-a", Pw | Args ], Acc) ->
 parse_args([ "-A", App | Args ], Acc) ->
     case re:run(App, "^(true|false|1|0)$", [caseless, {capture, none}]) of
         match ->
-            Bool = string_to_boolean(string:to_lower(App)),
+            Bool = z_convert:to_bool(string:to_lower(App)),
             parse_args(Args, Acc#{ app => Bool });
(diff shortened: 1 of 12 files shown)


mmzeeman (approved):
Side note: Funny that we have a template language on board, and not use it for this. Probably historic when this was done in python.

williamthome:
Should it be better if the `-S <supervisor>` option be  `-A <app>` like the `rebar3 new app` command?

apps/zotonic_mod_zotonic_site_management/src/zotonic_status_addsite.erl
```diff
@@ -278,21 +278,15 @@ copy_skeleton_dir(From, To, Options, Context) ->
...
 
-copy_file("SITE" ++ Ext, FromPath, ToPath, Options) when Ext == ".erl"; Ext == ".app.src" ->
-    % Replace with the site name
-    ToPath1 = filename:join([
-                    filename:dirname(ToPath),
-                    z_convert:to_list(proplists:get_value(site, Options)) ++ Ext
-                ]),
-    case filelib:is_file(ToPath1) of
+copy_file("SITE" ++ Ext = Filename, FromPath, ToPath, Options) when Ext == "_sup.erl"; Ext == "_app.erl" ->
+    case proplists:get_value(supervisor, Options) of
```
vkatsuba, line 282:
```suggestion
    case proplists:get_value(supervisor, Options, false) of
```

apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
```diff
@@ -189,6 +192,10 @@ parse_args([ "-n", Schema | Args ], Acc) ->
     parse_args(Args, Acc#{ dbschema => Schema });
 parse_args([ "-a", Pw | Args ], Acc) ->
     parse_args(Args, Acc#{ admin_password => Pw });
+parse_args([ "-S", "true" | Args ], Acc) ->
+    parse_args(Args, Acc#{ supervisor => true });
+parse_args([ "-S", "false" | Args ], Acc) ->
```
vkatsuba, line 197: If set `false` by default in place where this value is expected - will no make sense to have set `false` here, because `false` should be value by default.

apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
```diff
@@ -189,6 +192,10 @@ parse_args([ "-n", Schema | Args ], Acc) ->
     parse_args(Args, Acc#{ dbschema => Schema });
 parse_args([ "-a", Pw | Args ], Acc) ->
     parse_args(Args, Acc#{ admin_password => Pw });
+parse_args([ "-S", "true" | Args ], Acc) ->
```
vkatsuba, line 195: Maybe make sense to:
* Not handle for `"true"` value exactly in this place
* Handle `"TRUE"`/`"true"`/`1`/`"FALSE"`/`"false"`/`"0"`/"fAlSe"`/`"tRUe"` as `true`/`false`

apps/zotonic_mod_zotonic_site_management/priv/skel/blog/src/SITE_sup.erl
```diff
@@ -0,0 +1,50 @@
...
+
+%% API functions
+-export([ start_link/0 ]).
+
+%% Supervisor callbacks
+-export([ init/1 ]).
+
+-define(SERVER, ?MODULE).
+
+%%%=============================================================================
```
vkatsuba, line 31: I know that this is default `rebar3` code generating but at the same time in `zotonic` is not used similar separators, maybe make sense to remove similar comments in all places here.

apps/zotonic_mod_zotonic_site_management/priv/skel/blog/src/SITE_sup.erl
```diff
@@ -0,0 +1,50 @@
...
+%%% API functions
+%%%=============================================================================
+-spec start_link() -> supervisor:startlink_ret().
+
```
vkatsuba, line 35: Please, remove here extra new line here or add at line 43 - just for keeping one code style.

apps/zotonic_mod_zotonic_site_management/priv/skel/blog/src/SITE_app.erl
```diff
@@ -0,0 +1,29 @@
...
+%% limitations under the License.
+
+-module(%%SITE%%_app).
+-author("%%FULLNAME%%").
+
+-behaviour(application).
+
+-export([ start/2, stop/1 ]).
+
+start(_StartType, _StartArgs) ->
```
vkatsuba, line 25: I suppose it need to either add them everywhere or get rid of them.

vkatsuba (commented):
Great idea I suppose. Left a few recommendations :upside_down_face:.

apps/zotonic_mod_zotonic_site_management/src/zotonic_status_addsite.erl
```diff
@@ -278,21 +278,15 @@ copy_skeleton_dir(From, To, Options, Context) ->
...
-            ok;
+            copy_site_file(Filename, FromPath, ToPath, Options);
         false ->
-            % Replace the module name, write the site Erlang module.
-            {ok, SiteErl} = file:read_file(FromPath),
-            Outfile = replace_tags(SiteErl, Options),
-            file:write_file(ToPath1, Outfile)
+            ok
     end;
+copy_file("SITE" ++ Ext = Filename, FromPath, ToPath, Options) when Ext == ".erl"; Ext == ".app.src"; Ext == "_sup.erl"; Ext == "_app.erl" ->
```
vkatsuba, line 288:
```suggestion
copy_file("SITE" ++ Ext = Filename, FromPath, ToPath, Options) ->
```

apps/zotonic_mod_zotonic_site_management/priv/skel/blog/src/SITE_app.erl
williamthome: You mean the specs?

apps/zotonic_mod_zotonic_site_management/src/zotonic_status_addsite.erl
williamthome: These guards are a must because only erlang files have tags to replace.
In [this function](https://github.com/zotonic/zotonic/blob/35143344ce2eac224e8ce1d91696f84a15fb3e37/apps/zotonic_mod_zotonic_site_management/src/zotonic_status_addsite.erl#L316) the files are just copied.
Maybe could the `copy_site_file` function be renamed to `copy_file_replacing_tags`, `copy_erl_file` or something like that for a better understanding.

apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
```diff
@@ -192,11 +192,20 @@ parse_args([ "-n", Schema | Args ], Acc) ->
     parse_args(Args, Acc#{ dbschema => Schema });
 parse_args([ "-a", Pw | Args ], Acc) ->
     parse_args(Args, Acc#{ admin_password => Pw });
-parse_args([ "-A", "true" | Args ], Acc) ->
-    parse_args(Args, Acc#{ app => true });
-parse_args([ "-A", "false" | Args ], Acc) ->
-    parse_args(Args, Acc#{ app => false });
+parse_args([ "-A", App | Args ], Acc) ->
```
vkatsuba, line 195: Interesting approach. You can try simplify it. I suppose expected default value `false` for `app`. If yes, then it can be done like:

```erlang
parse_args([ "-A", App | Args ], Acc) ->
   parse_args(Args, Acc#{ app => app_to_boolean(string:to_lower(App)) });
%....

app_to_boolean(App) when App =:= "true"; App =:= "1" -> true;
app_to_boolean(_) -> false.
```

apps/zotonic_mod_zotonic_site_management/priv/skel/blog/src/SITE_app.erl
vkatsuba: Oh.. I miss this word :facepalm:. Yes, `spec`'s.

apps/zotonic_mod_zotonic_site_management/src/zotonic_status_addsite.erl
vkatsuba: Yep, all is ok here.

apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
vkatsuba: Hm.. I see function `z_convert:to_bool/1`, maybe you can try reuse it :thinking:?
```erlang
parse_args([ "-A", App | Args ], Acc) ->
    parse_args(Args, Acc#{ app => `z_convert:to_bool(string:to_lower(App)) });
```

williamthome: Good catch! `z_convert:to_bool/1` works.

apps/zotonic_mod_zotonic_site_management/src/zotonic_status_addsite.erl
```diff
@@ -278,21 +278,17 @@ copy_skeleton_dir(From, To, Options, Context) ->
...
-copy_file("SITE" ++ Ext, FromPath, ToPath, Options) when Ext == ".erl"; Ext == ".app.src" ->
-    % Replace with the site name
-    ToPath1 = filename:join([
-                    filename:dirname(ToPath),
-                    z_convert:to_list(proplists:get_value(site, Options)) ++ Ext
-                ]),
-    case filelib:is_file(ToPath1) of
+copy_file("SITE" ++ Ext = Filename, FromPath, ToPath, Options)
+    when Ext == "_sup.erl"; Ext == "_app.erl" ->
+    case proplists:get_value(app, Options, false) of
```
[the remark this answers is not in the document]
williamthome: The `addsite` command will always set it.

apps/zotonic_mod_zotonic_site_management/priv/skel/blog/src/SITE_sup.erl
```diff
@@ -0,0 +1,43 @@
...
+-spec start_link() -> supervisor:startlink_ret().
+start_link() ->
+    supervisor:start_link({local, ?SERVER}, ?MODULE, []).
+
+-spec init([]) -> {ok, {supervisor:sup_flags(), [supervisor:child_spec()]}}.
+init([]) ->
+    SupFlags = #{
+        strategy => one_for_all,
+        intensity => 0,
```
mmzeeman, line 39: This means that the site is not restarted in case of a crash. That is probably not what we want. It is nice though that an erroring site will not take down the whole server. FYI, for that problem zotonic's site supervisor had to take extra precautions.

Maybe OTP's defaults work better here... max 1 restarts in 5 seconds.

mmzeeman (approved):

apps/zotonic_mod_zotonic_site_management/priv/skel/blog/src/SITE_sup.erl
williamthome: Changed to use supervisor defaults for intensity and period in 9612ffe.

apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
```diff
@@ -189,6 +192,14 @@ parse_args([ "-n", Schema | Args ], Acc) ->
     parse_args(Args, Acc#{ dbschema => Schema });
 parse_args([ "-a", Pw | Args ], Acc) ->
     parse_args(Args, Acc#{ admin_password => Pw });
+parse_args([ "-A", App | Args ], Acc) ->
+    case re:run(App, "^(true|false|1|0)$", [caseless, {capture, none}]) of
```
vkatsuba, line 196: I suppose this function can be simplified like(without using `re:run/3`):
```erlang
parse_args([ "-A", App | Args ], Acc) ->
    parse_args(Args, Acc#{ app => z_convert:to_bool(string:to_lower(App)) });
```

apps/zotonic_mod_zotonic_site_management/priv/skel/blog/src/SITE_sup.erl
```diff
@@ -0,0 +1,40 @@
...
+
+%% API functions
+-export([ start_link/0 ]).
+
+%% Supervisor callbacks
+-export([ init/1 ]).
+
+-define(SERVER, ?MODULE).
+
+start_link() -> supervisor:start_link({local, ?SERVER}, ?MODULE, []).
```
vkatsuba, line 31:
```suggestion
start_link() ->
    supervisor:start_link({local, ?SERVER}, ?MODULE, []).
```

apps/zotonic_mod_zotonic_site_management/priv/skel/empty/src/SITE_sup.erl
```diff
@@ -0,0 +1,40 @@
...
+
+%% API functions
+-export([ start_link/0 ]).
+
+%% Supervisor callbacks
+-export([ init/1 ]).
+
+-define(SERVER, ?MODULE).
+
+start_link() -> supervisor:start_link({local, ?SERVER}, ?MODULE, []).
```
vkatsuba, line 31:
```suggestion
start_link() ->
    supervisor:start_link({local, ?SERVER}, ?MODULE, []).
```

apps/zotonic_mod_zotonic_site_management/priv/skel/empty/src/SITE_sup.erl
```diff
@@ -0,0 +1,40 @@
...
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+
+-module(%%SITE%%_sup).
+-author("%%FULLNAME%%").
+
+-behaviour(supervisor).
+
+%% API functions
```
vkatsuba, line 23: Could you please remove comments `%% API functions` and `%% Supervisor callbacks` here and in other `SITE_sup.erl`? Although they can be left. But since there isn't a lot of code here and it looks pretty self-explanatory, I think comments like this can be a bit redundant. But it's up to you to decide - it's just a recommendation :upside_down_face:.

vkatsuba (approved):
Looks awesome! Leaved few minor proposal of code changes. :upside_down_face:.

mworrell:
@vkatsuba I saw you had some open questions.  When you are ok, feel free to merge 👍

vkatsuba:
> @vkatsuba I saw you had some open questions. When you are ok, feel free to merge +1

Sure.

apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
williamthome: The regex also checks match for "tRuE", "FaLsE", or any caseless. z_convert:to_bool/1 does not make these checks. Anyway, I think the check is needed because if there is no match an error is displayed and the command exits.
Without the regex, the option "-A foo" it's a valid option and the command proceeds as #{app => false}.

vkatsuba (approved):

PR #2975, opened 2022-05-17
Umbrella proposal

williamthome:
### Description

Provides a multi-app project structure to the site, like [`rebar3 new umbrella`](https://rebar3.readme.io/docs/basic-usage) command.
The site dir will become, e.g.
```
foo
|-- apps
|   |-- foo
|   |   |-- priv
|   |   |   |-- dispatch
|   |   |   |-- templates
|   |   |   |-- ...
|   |   |   |-- zotonic_site.config
|   |   |   `-- zotonic_site.config.in
|   |   `-- src
|   |       |-- foo.app.src
|   |       `-- foo.erl
|   |-- zotonic_mod_foo
|   |   |-- priv
|   |   |   `-- dispatch
|   |   |-- src
|   |   |   |-- zotonic_mod_foo.app.src
|   |   |   `-- mod_foo.erl
|   |   `--rebar.config
|   `-- *more apps*
`-- .gitignore
```
For new projects, `-U` option was added to `addsite` command:
```
bin/zotonic addsite -U true <site>
```
Tested creating a new site with and without the new option. The site, database and observers continue working as expected.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
@@ -28,24 +28,26 @@
 
 -define(SKEL, "blog").
 -define(APP, false).
+-define(UMBRELLA, false).
 
 info() ->
     "Add a site and install database".
 
 usage() ->
     io:format("Usage: zotonic addsite [options] <site_name> ~n~n"),
-    io:format(" -s <skel>     Skeleton site (one of 'blog', 'empty', 'nodb'; default: ~s~n", [?SKEL]),
-    io:format(" -H <host>     Site's hostname (default: <site_name.test>) ~n"),
-    io:format(" -L            Create the site in the current directory and add a symlink to zotonic app_user~n"),
-    io:format(" -G <url>      Clone from this Git url, before copying skeleton~n"),
-    io:format(" -h <host>     Database host (default: ~s) ~n", [ z_config:get(dbhost) ]),
-    io:format(" -p <port>     Database port (default: ~p) ~n", [ z_config:get(dbport) ]),
-    io:format(" -u <user>     Database user (default: ~s) ~n", [ z_config:get(dbuser) ]),
-    io:format(" -P <pass>     Database password (default: ~s) ~n", [ z_config:get(dbpassword) ]),
-    io:format(" -d <name>     Database name (default: ~s) ~n", [ z_config:get(dbdatabase) ]),
-    io:format(" -n <schema>   Database schema (defaults to <site_name>) ~n"),
-    io:format(" -a <pass>     Admin password~n"),
-    io:format(" -A <app>      If true, initializes a site app and a root supervisor when the site starts (default: ~p)~n~n", [?APP]).
+    io:format(" -s <skel>        Skeleton site (one of 'blog', 'empty', 'nodb'; default: ~s~n", [?SKEL]),
+    io:format(" -H <host>        Site's hostname (default: <site_name.test>) ~n"),
+    io:format(" -L               Create the site in the current directory and add a symlink to zotonic app_user~n"),
+    io:format(" -G <url>         Clone from this Git url, before copying skeleton~n"),
+    io:format(" -h <host>        Database host (default: ~s) ~n", [ z_config:get(dbhost) ]),
+    io:format(" -p <port>        Database port (default: ~p) ~n", [ z_config:get(dbport) ]),
+    io:format(" -u <user>        Database user (default: ~s) ~n", [ z_config:get(dbuser) ]),
+    io:format(" -P <pass>        Database password (default: ~s) ~n", [ z_config:get(dbpassword) ]),
+    io:format(" -d <name>        Database name (default: ~s) ~n", [ z_config:get(dbdatabase) ]),
+    io:format(" -n <schema>      Database schema (defaults to <site_name>) ~n"),
+    io:format(" -a <pass>        Admin password~n"),
+    io:format(" -A <app>         If true, initializes a site app and a root supervisor when the site starts (default: ~p)~n~n", [?APP]),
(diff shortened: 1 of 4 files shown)


apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
```diff
@@ -28,24 +28,26 @@
...
+    io:format(" -L               Create the site in the current directory and add a symlink to zotonic app_user~n"),
+    io:format(" -G <url>         Clone from this Git url, before copying skeleton~n"),
+    io:format(" -h <host>        Database host (default: ~s) ~n", [ z_config:get(dbhost) ]),
+    io:format(" -p <port>        Database port (default: ~p) ~n", [ z_config:get(dbport) ]),
+    io:format(" -u <user>        Database user (default: ~s) ~n", [ z_config:get(dbuser) ]),
+    io:format(" -P <pass>        Database password (default: ~s) ~n", [ z_config:get(dbpassword) ]),
+    io:format(" -d <name>        Database name (default: ~s) ~n", [ z_config:get(dbdatabase) ]),
+    io:format(" -n <schema>      Database schema (defaults to <site_name>) ~n"),
+    io:format(" -a <pass>        Admin password~n"),
+    io:format(" -A <app>         If true, initializes a site app and a root supervisor when the site starts (default: ~p)~n~n", [?APP]),
```
vkatsuba, line 49:
```suggestion
    io:format(" -A <app>         If true, initializes a site app and a root supervisor when the site starts (default: ~p)~n", [?APP]),
```

apps/zotonic_launcher/src/command/zotonic_cmd_addsite.erl
```diff
@@ -193,14 +196,19 @@ parse_args([ "-n", Schema | Args ], Acc) ->
...
+    parse_bool_arg(app, App, Args, Acc);
+parse_args([ "-U", Umbrella | Args ], Acc) ->
+    parse_bool_arg(umbrella, Umbrella, Args, Acc);
 parse_args([ "-" ++ _ = Arg | _ ], _Acc) ->
     {error, Arg};
 parse_args(Rest, Acc) ->
     {ok, {Acc, Rest}}.
+
+parse_bool_arg(Key, Arg, Args, Acc) ->
+    case re:run(Arg, "^(true|false|1|0)$", [caseless, {capture, none}]) of
```
vkatsuba, line 208:
```erlang
re:run("ц", "^(true|false|1|0)$", [caseless, {capture, none}]). ** exception error: bad argument
     in function  re:run/3
        called as re:run([1094],
                         "^(true|false|1|0)$",
                         [caseless,{capture,none}])
        *** argument 1: not an iodata term
```
Here will be a crash with try to parse `unicode` string. Maybe it make sense to add `catch` here.

williamthome:
> If you make this new umbrella site, where is it created? As the bin/zotonic command is there, I assume you start with a Zotonic clone/checkout?

The site will be generated under the `apps_user/` as is today.
```
apps_user
|-- foo
|   |-- apps
|   |   |-- foo
|   |   |   |-- priv
|   |   |   |   |-- dispatch
|   |   |   |   |-- templates
|   |   |   |   |-- ...
|   |   |   |   |-- zotonic_site.config
|   |   |   |   `-- zotonic_site.config.in
|   |   |   `-- src
|   |   |       |-- foo.app.src
|   |   |       `-- foo.erl
|   |   |-- zotonic_mod_foo
|   |   |   |-- priv
|   |   |   |   `-- dispatch
|   |   |   |-- src
|   |   |   |   |-- zotonic_mod_foo.app.src
|   |   |   |   `-- mod_foo.erl
|   |   |   `--rebar.config
|   |   `-- *more apps*
|   `-- .gitignore
`-- *other sites nor modules*
```
Nothing more changes, just the site directory structure.
The `rebar3` will compile the umbrella to the `ebin/` folder.
To clarify:
```
apps_user
|-- foo
|   |-- priv
|   |-- src
|   `-- .gitignore
|-- zotonic_mod_foo
|   |-- priv
|   |-- src
|   `-- .gitignore
`-- *other sites nor modules*
```
compiles to 
```
ebin
|-- foo
|-- zotonic_mod_foo
`-- *other apps*
```
and
```
apps_user
|-- foo
|   |-- apps
|   |   |-- foo
|   |   |   |-- priv
|   |   |   `-- src
|   |   |-- zotonic_mod_foo
|   |   |   |-- priv
|   |   |   `-- src
|   |   `-- *more apps*
|   `-- .gitignore
`-- *other sites nor modules*
```
also compiles to
```
ebin
|-- foo
|-- zotonic_mod_foo
`-- *other apps*
```
This change adds the monorepo possibility for the site.

mworrell:
@mmzeeman I think this is also a good solution for developing umbrella apps (that use the Zotonic Hex packages) in the Zotonic umbrella   What do you think?

mmzeeman:
Does this work for rebar? Rebar can be a bit tricky with umbrella apps.

mworrell:
> Does this work for rebar? Rebar can be a bit tricky with umbrella apps.

The change does set an extra src path in the rebar.config to `apps_user/*/apps/*`, so I think it is safe.  In Zotonic we only scan the code lib directory for new apps, so Zotonic should be ok with this extra level as well.

williamthome:
Yes, Zotonic only scans the compiled files in the `_build/lib` dir and `rebar3` works very well with umbrella apps.

mmzeeman:
Rebar3 does not support umbrella apps as dependencies.

But this is a great idea, and more inline with the new example project. 👍🏻

williamthome:
You are right, umbrella apps are not allowed to be a dependency. However, a monorepo app should be a top-level app and not a dependency. If the dependency is needed, it must live outside the umbrella app.

mmzeeman:
That is why every app in zotonic apps directory get pushed separately to hex.

mmzeeman (approved):

mworrell:
👍 Thanks!

PR #2977, opened 2022-05-17
Fix compile warning

williamthome:
### Description

Running `make` I receive this warning:
> ===> Compiling zotonic_filehandler
apps/zotonic_filehandler/src/zotonic_filehandler_compile.erl:335:29: Warning: key 'result' will be overridden in expression

This PR just remove the overridden arg.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Ah, the `result` should be `error` here, and the second `result => Reason` should be removed.

Idea is that the `reason` reflects _why_ something happened, and the `result` what the result of the operation was.

mworrell:
Thanks!

PR #2984, opened 2022-05-18
Adds a function to open site as safe in Chrome for development

williamthome:
### Description

I'm facing issues getting the Cotonic to work in localhost over HTTPS.
In the browser console, I receive these messages:
```
Could not start serviceWorker due to a SecurityError.
See https://cotonic.org/#model.serviceWorker for more information.
```
The [Cotonic docs](https://cotonic.org/#model.serviceWorker) say to me this:
> A Note on Security
The serviceWorker needs https to work correctly. The Chrome browser does not accept self-signed certificates for running serviceWorkers. Either run from localhost, use a real certificate, or follow the instructions [on this page by Dean Hume](https://deanhume.com/testing-service-workers-locally-with-self-signed-certificates/). <br>
Safari and Firefox accept self-signed certificates for running the serviceWorker.

But I don't want to generate a self-signed certificate, then the solution I chose was to run Chrome by the command line.

This PR implements `mod_development:chrome/1` and `mod_development:chrome/2`.
e.g. `mod_development:chrome(foo, ["--incognito"])` opens the Chrome in the `foo` site URL and in incognito mode.

I decided to open this PR because I think this may be helpful to someone else and can facilitate the dev.
One GIF to the explain rescue:
zotonic_chrome

Side note:
* Firefox doesn't support these options by command line;
* Don't know about Opera;
* os:cmd/1 freezes the shell.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_launcher/src/command/zotonic_cmd_chrome.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_cmd_chrome.erl
@@ -42,7 +42,7 @@ usage(Browser) ->
     io:format("USAGE: zotonic " ++ Browser ++ " [options] [switches] <site_name> ~n"),
     io:format("~n"),
     io:format("Options: ~n"),
-    io:format("  -s <secure>   Opens site as secure and ignore certificate errors ~n"),
+    io:format("  -s <secure>            Opens site as secure and ignore certificate errors ~n"),
     io:format("~n"),
     io:format("Switches: ~n"),
     io:format("  --incognito            Launches " ++ Browser ++ " directly in Incognito private browsing mode ~n"),
--- a/apps/zotonic_launcher/src/command/zotonic_cmd_chrome.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_cmd_chrome.erl
@@ -1,0 +1,70 @@
+%% @author William Fank Thomé <williamthome@hotmail.com>
+%% @copyright 2022 Marc Worrell
+%% @doc Chrome CLI command
+
+%% Copyright 2022 Marc Worrell
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+
+-module(zotonic_cmd_chrome).
+-author("William Fank Thomé <williamthome@hotmail.com>").
+
+%% API
+-export([
+    info/0,
+    usage/0,
+    run/1
+]).
+
+info() ->
+    "Opens Chrome in the site URL with secure certificate flags".
+
+usage() ->
+    io:format("USAGE: zotonic chrome [options] <site_name> ~n"),
+    io:format("Options: ~n"),
+    io:format("  See https://peter.sh/experiments/chromium-command-line-switches/ ~n"),
+    io:format("Note: ~n"),
+    io:format("  Requires 'zotonic start' ~n~n").
+
+run(Args) ->
+    case zotonic_command:net_start() of
+        ok ->
+            run_parse_args(Args);
+        {error, _} = Error ->
+            zotonic_command:format_error(Error)
+    end.
+
+run_parse_args(Args) ->
+    case parse_args(Args, #{args => []}) of
+        {ok, #{site := Site, args := ParsedArgs}} ->
+            SiteName = list_to_atom(Site),
+            Res = zotonic_command:rpc(
+                mod_development, exec_browser, [ chrome, SiteName, ParsedArgs ]
+            ),
+            io:format("~p~n", [ Res ]);
+        {error, _} = ZError ->
+            zotonic_command:format_error(ZError)
+    end.
+
+parse_args(["--" ++ _], _Acc) ->
+    {error, "The last entry should be the site name"};
+parse_args([Site], Acc) ->
+    {ok, Acc#{site => Site}};
+parse_args(["--" ++ _ = Arg | Rest], #{args := Args} = AccIn) ->
+    AccOut = AccIn#{args => [Arg | Args]},
+    parse_args(Rest, AccOut);
+parse_args([], _Acc) ->
+    usage(),
+    halt(1);
+parse_args([Arg | _Args], _Acc) ->
+    {error, "All Chrome args starts with '--'. '" ++ Arg ++ "' is invalid"}.
--- a/apps/zotonic_launcher/src/command/zotonic_cmd_chrome.erl
+++ b/apps/zotonic_launcher/src/command/zotonic_cmd_chrome.erl
@@ -22,34 +22,46 @@
 %% API
 -export([
     info/0,
+    info/1,
     usage/0,
-    run/1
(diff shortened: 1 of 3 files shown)


mworrell:
Maybe you can use erlexec?

I was wondering if we should add a `bin/zotonic dev chrome sitename` command?  Where the `chrome sitename` part is an argument to the `dev` (mod_development) module command.

I agree that typing `thisisunsafe` in Chrome is a bit of a nuisance.  I am mostly using Safari and the self-signed certs. They are tied to the `.test` domains I am using, so I consider them to be safe enough.

mworrell:
BTW on macOS Google Chrome is always at the path:

```
/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome
```

mworrell:
Found a better macOS command top open Chrome, example:

```
open -a "Google Chrome" --args --ignore-certificate-errors http://www.example.com
```

apps/zotonic_mod_development/src/mod_development.erl
```diff
@@ -96,6 +98,48 @@ observe_request_context(#request_context{ phase = refresh }, Context, _Context)
...
+%%   --multi-profiles       Enable multiple profiles in Chrome
+%% e.g.
+%% ``` mod_development:chrome(foo, ["--incognito", "--start-maximized"]). '''
+-spec chrome(SiteOrContext :: atom() | z:context()) -> {ok, port()} | {error, term()}.
+chrome(SiteOrContext) ->
+    chrome(SiteOrContext, []).
+
+-spec chrome(SiteOrContext :: atom() | z:context(), ExtraArgs :: [string()]) -> {ok, port()} | {error, term()}.
+chrome(#context{} = Context, ExtraArgs) ->
+    case os:find_executable("google-chrome") of
```
vkatsuba, line 116: I suppose need also add `os:find_executable("chromium-browser")` for Linux.

williamthome:
> Maybe you can use erlexec?
> 
> I was wondering if we should add a `bin/zotonic dev chrome sitename` command? Where the `chrome sitename` part is an argument to the `dev` (mod_development) module command.
> 
> I agree that typing `thisisunsafe` in Chrome is a bit of a nuisance. I am mostly using Safari and the self-signed certs. They are tied to the `.test` domains I am using, so I consider them to be safe enough.

Totally agreed. I was thinking about add some functionality to `bin/zotonic` script when was creating the PR, but I preferred to wait a response about this before.
I will do some changes tonight and also try to use `erlexec`.

apps/zotonic_mod_development/src/mod_development.erl
williamthome: I think we can try to pick the executable in this order `["chromium-browser", "google-chrome"]`.

williamthome:
@mworrell the PR command works on Mac?
If not, we can use `os:type().` to get the OS name and use another one.

mworrell:
On macOS we need to use `/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome`

williamthome:
Ok, so `os:find_executable/1` doesn't work for macOS.
According to [this post](https://apple.stackexchange.com/a/287686) for chromium should be `/Applications/Chromium.app/Contents/MacOS/Chromium`.

williamthome:
I will add support for both, Chromium and Chrome.
If you guys have more suggestions, they are welcome.

williamthome:
Could someone confirm what `os:type().` returns on mac?

williamthome:
Changelog:
* Add support for Chromium browser and macOS. Chrome and Chromium are working on Ubuntu 21.10. I don't know about mac, it needs to be tested;
* Add `chrome` and `chromium` to zotonic script as `bin/zotonic <browser> [options] [switches] <site_name>`, e.g. `bin/zotonic chrome -s --incognito foo` starts chrome with secure option, incognito mode and in the foo site URL.

mworrell:
Ah, then just don't use erlexec, it needs to be started and a NIF needs to be loaded to communicate with the processes.

You might have seen that in Zotonic start/stop/debug we generate a command line that is the executed in the bin/zotonic script, so there we also work around this kind of problems.

BTW `os:type()` on macOS returns:

```
1> os:type().
{unix,darwin}
```

Alternative is to do an RPC to the zotonic beam, and let that one execute the command. But then the browser is started using the operating-system-user of the beam, which might not always be the best solution.

apps/zotonic_launcher/src/command/zotonic_cmd_chrome.erl
```diff
@@ -0,0 +1,100 @@
+%% @author William Fank Thomé <williamthome@hotmail.com>
+%% @copyright 2022 Marc Worrell
+%% @doc Chrome CLI command
+
+%% Copyright 2022 Marc Worrell
```
mworrell, line 5: Don't forget to set your copyright :-)

apps/zotonic_mod_development/src/mod_development.erl
```diff
@@ -250,3 +336,27 @@ observe_admin_menu(#admin_menu{}, Acc, Context) ->
...
+        "--ignore-certificate-errors",
+        "--unsafely-treat-insecure-origin-as-secure=" ++ SiteUrl
+        | ExtraArgs
+    ]),
+    do_exec_chrome(Executable, SiteUrl, Args, Options);
+exec_chrome(Executable, SiteUrl, ExtraArgs, Options) ->
+    do_exec_chrome(Executable, SiteUrl, ExtraArgs, Options).
+
+do_exec_chrome(Executable, SiteUrl, Args, _Options) ->
+    Command = io_lib:format("~s ~s ~s", [Executable, Args, SiteUrl]),
```
mworrell, line 352: We need to escape the path and siteUrl here as they might contain spaces and/or other characters.

Options to use are `z_filelib:os_escape(Executable)`, and `z_filelib:os_filename(SiteUrl)`

apps/zotonic_mod_development/src/mod_development.erl
```diff
@@ -250,3 +336,27 @@ observe_admin_menu(#admin_menu{}, Acc, Context) ->
 
      |Acc].
 
+exec_chrome(Executable, SiteUrl, ExtraArgs, #{secure := true} = Options) ->
+    TmpPath = filename:join([<<"/">>, <<"tmp">>, <<"foo">>]),
```
mworrell, line 340: To fetch the tmp dir you can use `z_tempfile:temppath()` which also checks the `TMPDIR` and `TEMPDIR` env variables.

Then you would get something like:

```erlang
TmpPath = filename:join([ z_tempfile:temppath(), "zotonic-chrome" ]),
```

williamthome:
> Ah, then just don't use erlexec, it needs to be started and a NIF needs to be loaded to communicate with the processes.
> 
> You might have seen that in Zotonic start/stop/debug we generate a command line that is the executed in the bin/zotonic script, so there we also work around this kind of problems.
> 
> BTW `os:type()` on macOS returns:
> 
> ```
> 1> os:type().
> {unix,darwin}
> ```
> 
> Alternative is to do an RPC to the zotonic beam, and let that one execute the command. But then the browser is started using the operating-system-user of the beam, which might not always be the best solution.

So it's ok to stay as is?

mworrell:
> So it's ok to stay as is?

Yes 👍

mworrell:
Just let me know when you are ready!

williamthome:
Ready.
Just for mac I couldn't say this works.

mworrell:
Reading the code it will work 👍 

Thanks!  Merging this nice feature!

mworrell:
Just tried:

```shell
bin/zotonic chrome -s learningstone
```

And it worked like advertised.

williamthome:
Nice! o/

PR #2987, opened 2022-05-23
Search view

vkatsuba:
### Description

Quick search view implementation

### DIFF
--- a/apps/zotonic_mod_search/priv/lib/js/zotonic-search-view.js
+++ b/apps/zotonic_mod_search/priv/lib/js/zotonic-search-view.js
@@ -1,0 +1,62 @@
+jQuery( document ).ready(function() {
+    var surl, input, $form, $result;
+
+    jQuery( '[name="qs"]' ).on('input', function() {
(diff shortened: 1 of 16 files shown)


apps/zotonic_mod_search/priv/lib/js/zotonic-search-view.js
```diff
@@ -0,0 +1,62 @@
+jQuery( document ).ready(function() {
+    var surl, input, $form, $result;
+
+    jQuery( '[name="qs"]' ).on('input', function() {
+        surl = window.location.protocol + '//' + window.location.host + '/search-view?qs=';
+        $form = jQuery( this ).closest( 'form' );
+        input = this;
+
+        jQuery.ajax({
```
mworrell, line 9: Maybe we could also use the MQTT API?

I first thought of the HTTP API, but on re-reading the code we could also use:

```
cotonic.broker.call("bridge/orgin/model/template/get/search_view.tpl", { qs: this.value }).then(...)
```

PR #2992, opened 2022-05-30
core: more readable errors from the template compiler.

mworrell:
### Description

Better error messages on syntax errors.
Upgrade template_compiler for structured error logging.

Screenshot 2022-05-30 at 18 00 16

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_core/src/support/z_logger_formatter.erl
```diff
@@ -497,7 +497,7 @@ pretty_stack(Stack, Config) ->
             [
                 "\n     ",
                 maybe_color(colored_mfa, Config),
-                io_lib:format("~tp:~tp/~p", [ M, F, Arity ]),
+                io_lib:format("~ts:~ts/~p", [ bin(M), bin(F), Arity ]),
```
vkatsuba, line 500: Maybe can be use `z_convert:to_binary/1` instead of custom `bin/1`?
```suggestion
                io_lib:format("~ts:~ts/~p", [ z_convert:to_binary(M), z_convert:to_binary(F), Arity ]),
```

rebar.lock
```diff
@@ -56,7 +56,7 @@
  {<<"sidejob">>,{pkg,<<"sidejob">>,<<"2.1.0">>},0},
  {<<"ssl_verify_fun">>,{pkg,<<"ssl_verify_fun">>,<<"1.1.6">>},0},
  {<<"syslog">>,{pkg,<<"syslog">>,<<"1.1.0">>},0},
- {<<"template_compiler">>,{pkg,<<"template_compiler">>,<<"2.1.1">>},0},
+ {<<"template_compiler">>,{pkg,<<"template_compiler">>,<<"2.3.0">>},0},
```
vkatsuba, line 59: Maybe also make sense update https://github.com/zotonic/zotonic/blob/master/apps/zotonic_core/rebar.config#L48

mworrell: This is the lock file, in the .config file it says something like `~> 2.1`

apps/zotonic_core/src/support/z_logger_formatter.erl
mworrell: Maybe, for lists (string) I would like to handle them as unicode strings, where `z_convert` at the moment handles lists as utf8 `iodata`.

mworrell: The `z_convert:to_binary` is something I would like to change in a next big release, so that it assumes Unicode data if it is a list.

vkatsuba (approved):

PR #3000, opened 2022-06-03
Adds middleware notifier

williamthome:
### Description

Gives control over how requests are to be processed by a middleware notification.
Handling the request in a module:
```erlang
observe_middleware(#middleware{ on = welformed }, Context, _Context) ->
    io:format("on welformed middleware~n"),
    Context;
observe_middleware(#middleware{ }, Context, _Context) ->
    Context.
```
So custom headers and other stuff can be made during the request.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/include/zotonic_notifications.hrl
+++ b/apps/zotonic_core/include/zotonic_notifications.hrl
@@ -1192,7 +1192,7 @@
 %% Type: foldl
 %% Return: modified context
 -record(middleware, {
-    on :: executed | welformed | handled
+    on :: request | welformed | handled
 }).
 
 
--- a/apps/zotonic_core/include/zotonic_notifications.hrl
+++ b/apps/zotonic_core/include/zotonic_notifications.hrl
@@ -1188,6 +1188,14 @@
 }).
 
 
+%% @doc Delegates the request processing.
+%% Type: foldl
+%% Return: modified context
+-record(middleware, {
+    on :: executed | welformed | handled
+}).
+
+
 % Simple mod_development notifications:
 % development_reload - Reload all template, modules etc
 % development_make - Perform a 'make' on Zotonic, reload all new beam files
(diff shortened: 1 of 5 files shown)


apps/zotonic_core/include/zotonic_notifications.hrl
```diff
@@ -1188,6 +1188,14 @@
 }).
 
 
+%% @doc Delegates the request processing.
+%% Type: foldl
+%% Return: modified context
+-record(middleware, {
+    on :: executed | welformed | handled
```
mworrell, line 1195: Maybe `executed` should be called `init` or `start` ?
That might be clearer.

mmzeeman (approved):
Nice idea. That allows for some interesting site level customizations. Maybe a little change like @mworrell suggested, because it happens before cowmachine executes the request.

vkatsuba:
Wow PR with number 3000! Zotonic rocks 🤘

apps/zotonic_core/include/zotonic_notifications.hrl
williamthome: I'm in doubt about what could be better, but I think `start` would be best. What do you think?

williamthome: Or maybe `request`?

mworrell: `request` sounds good to me.  @mmzeeman @vkatsuba  ok?

vkatsuba: Sounds good.

williamthome: Changed `executed` to `request`.

vkatsuba (approved):

mworrell (approved):

mworrell:
Thanks again!

PR #3001, opened 2022-06-03
Retain MQTT client id

williamthome:
### Description

Adds the possibility to subscribe into `model/sessionStorage/post/mqtt-origin-client-id` topic.
The only changed thing is the `{ retain: true }` option.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
+++ b/apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
@@ -107,7 +107,12 @@ function zotonic_startup() {
                     cotonic.broker.publish(
                         "model/sessionStorage/post/mqtt-origin-client-id",
                         msg.payload.client_id,
-                        { retain: true }
+                    );
+                    // Retain the client-id
+                    cotonic.broker.publish(
+                        "$bridge/origin/client-id",
+                        msg.payload.client_id,
+                        { retain: true },
                     );
                 }
             }, { wid: "zotonicjs" });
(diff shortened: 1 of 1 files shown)


mmzeeman (commented):
What is the intended effect of this change? The sessionStorage model also stores the client id.

Ah.. being able to subscribe to the the id is nice. Maybe we should make this more explicit by publishing it retained to the `$bridge\origin\client-id` instead of doing it indirectly via the sessionStorage. What do you think @mworrell

williamthome:
The motivation for this is to get notified immediately after a custom worker connects.
Without the retain option I could not receive the client id because the topic `model/sessionStorage/post/mqtt-origin-client-id` fires before a custom worker can handle the event.

mworrell:
Retained in `$bridge/origin/client-id` sounds good to me, it is the client-id of that bridge.

williamthome:
@mmzeeman @mworrell could check if I understand the proposed changes?

mmzeeman:
What I wanted to prevent was that you need to know the location of the client-id of the bridge and get it from some session storage topic. It was a bit strange for me to also use retained messages to the session storage model post calls.

With this change the client-id is now stored in two locations. One as retained message, and second in the session storage. That is probably not a good idea.

Is it possible to just store it just as a retained message? And maybe move this before the code which publishes the client id? https://github.com/zotonic/zotonic/blob/79475730eceebcfd44c8ce56b7ba563a139c1baa/apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js#L121 

It is probably a bit tricky with respect to race conditions during reconnects.

What do you think @mworrell?

mworrell:
Maybe we could add a small broker function to do a retained value lookup?

Then we can remove the session storage of client-id.

PR #3029, opened 2022-06-23
mod_seo_sitemap: update sitemap on pivot, remove prio 0 resources from urlsets (0.x)

mworrell:
### Description

Fix an issue where resources with pro 0.0 were still included in the sitemap.

Also update the sitemap on pivot, this is better as _on update_ will give more frequent updates of the sitemap (after every edit).

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


robvandenbogaard:
Good to have it on pivot instead. I realised that if priority means "boost", that priority 0 entries are actually ok to have, but it is good to now have a simple way to exclude entries.

PR #3052, opened 2022-07-14
core: show more descriptive error on Ajax errors.

mworrell:
### Description

Also suppress errors if the page is unloading.

On FireFox in-flight Ajax requests are canceled with an error if the page is unloading (navigating away).
This results in "FAIL: error" messages.

By suppressing the messages if the page is unloading we can prevent those errors from appearing.

Also "FAIL: error" is not very helpful, so only show the error text if it is not "error" and replace the "FAIL" with something more descriptive about _what_ is failing.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_base/lib/js/apps/zotonic-1.0.js
+++ b/modules/mod_base/lib/js/apps/zotonic-1.0.js
@@ -991,8 +991,14 @@ function z_ajax(options, data)
         error: function(xmlHttpRequest, textStatus, errorThrown)
         {
             z_stop_spinner();
-            $.misc.error("FAIL: " + textStatus);
             z_unmask_error(options.trigger_id);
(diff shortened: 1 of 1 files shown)


robvandenbogaard (approved):

mmzeeman (commented):
Where is $.misc defined?

mworrell:
> Where is $.misc defined?

In `z.widgetmanager.js`

mmzeeman:
That looks like it deserves some modernization.. I think all browsers define console.log nowadays. That looks like it comes from the pre IE 11 dark ages.

mworrell:
True, it came from there.  It is the 0.x so I am ok with leaving this as-is.

For the master we might want to update this.

PR #3058, opened 2022-07-20
Return not_found instead of no_session. Fixes #3057

### LINKED ISSUE #3057: Filter temporary_resource always returns "no_session"
## Steps to reproduc

Create a template which uses `temporary_resource`. You will see it always returns `no_session` in the zotonic console.

## Current behavior

It is not possible to store the id of the temporary resource.

## Expected behavior

The id is saved.

mmzeeman:
### Description

Fix #3057

The filter returned no_session, when there was no server_storage session. This session was not started yet, because it will only be created when a value is stored.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman:
Made a mess of this. Closing PR and branch.

PR #3059, opened 2022-07-20
Return not_found instead of no_session. Fixes #3057

### LINKED ISSUE #3057: Filter temporary_resource always returns "no_session"

mmzeeman:
### Description

Fix #3057

When the user has no server_storage_session, nothing could have been stored, so return not_found instead of no_session. The storage session will be created when an id is stored.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman:
It is probably handy to include issue #3060 in this PR as well.

mmzeeman:
Fixes #3057 en #3060

PR #3067, opened 2022-07-22
core: on rsc update, as sudo, do not allow to add edges.

mworrell:
### Description

Edges insertions on rsc-update can change subjects and also add ACL rights (like membership of user groups). Using admin rights for such updates (like when signing up) can have unwanted changes to normally not user-editable resources or predicates.

This is a proposal to not allow inserting edges during a sudo. This change makes it safe(r) to use sudo on insertion of resources using non-checked user-supplied forms.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mmzeeman (approved):
Good call.. It might be possible to slip in edge changes this way.

PR #3074, opened 2022-07-27
Check for empty search string

cdfa:
So it does not get propagated as filter

### Description

Text search query parameters with an empty binary string are now ignored, so all results are returned instead of none. This is now in line with the behavior of normal empty strings, when the parameter is missing and the 1.0 version.

### Checklist

- [X] documentation updated
- [ ] tests added
- [X] no BC breaks


mworrell:
Thanks!

PR #3089, opened 2022-08-22
core: add option to exclude categories from the SEO indexing (0.x)

mworrell:
### Description

This adds an option to exclude all pages in a category from indexing by a search engine.

For this the property `is_seo_noindex_cat` is set 

Also update the Docs GH Action to keep working with a newer Ubuntu version as GH is removing 18.04 support.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_admin/templates/_admin_edit_content_seo.tpl
+++ b/modules/mod_admin/templates/_admin_edit_content_seo.tpl
@@ -100,7 +100,7 @@
                                 {% if id.seo_noindex or is_seo_noindex_cat %}checked="checked"{% endif %}
                                 {% if is_seo_noindex_cat %}disabled{% endif %}
                             />
-                            {_ Ask google to not index this page _}
+                            {_ Ask Google to not index this page _}
                         </label>
 
                         {% if is_seo_noindex_cat %}
@@ -127,7 +127,7 @@
                                 {% if not is_editable %}disabled{% endif %}
                                 {% if id.is_seo_noindex_cat %}checked="checked"{% endif %}
                             />
-                            {_ Ask google to not index any page of this category _} ({{ id.title }})
+                            {_ Ask Google to not index any page of the category: _} {{ id.title }}
                         </label>
                     </div>
                 </div>
--- a/modules/mod_admin/templates/_admin_edit_content_seo.tpl
+++ b/modules/mod_admin/templates/_admin_edit_content_seo.tpl
@@ -96,9 +96,8 @@
                                 type="checkbox"
                                 name="seo_noindex"
                                 value="1"
-                                {% if not is_editable %}disabled{% endif %}
+                                {% if not is_editable or is_seo_noindex_cat %}disabled{% endif %}
                                 {% if id.seo_noindex or is_seo_noindex_cat %}checked="checked"{% endif %}
-                                {% if is_seo_noindex_cat %}disabled{% endif %}
                             />
                             {_ Ask Google to not index this page _}
                         </label>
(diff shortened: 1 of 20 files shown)


modules/mod_seo_sitemap/support/seo_sitemap.erl
```diff
@@ -83,7 +83,8 @@ categories(Context) ->
     Cats = z_db:q("select id from rsc where category_id = $1", [ CatCat ], Context),
     PIds = lists:filtermap(
         fun({Id}) ->
-            case z_acl:rsc_visible(Id, Context) of
+            IsSeoNoIndexCat = z_convert:to_bool(m_rsc:p_no_acl(Id, is_seo_noindex_cat, Context)),
+            case IsSeoNoIndexCat andalso z_acl:rsc_visible(Id, Context) of
```
robvandenbogaard, line 87: Shouldn't this condition be `(not IsSeoNoIndexCat) andalso visible`?

modules/mod_base/controllers/controller_file.erl
```diff
@@ -136,6 +139,18 @@ provide_content(ReqData,  {Info,Context} = State) ->
...
 
+is_noindex(#z_file_info{acls=Acls}, Context) ->
+    lists:any(
+        fun
+            (Id) when is_integer(Id) ->
+                CatId = m_rsc:p_no_acl(Id, category_id, Context),
+                z_convert:to_bool(m_rsc:p_no_acl(Id, seo_noindex, Context))
+                orelse z_convert:to_bool(m_rsc:p_no_acl(CatId, is_seo_noindex_cat, Context));
+            (_) ->
+                false
```
robvandenbogaard, line 150: Should non-resources be indexed?

modules/mod_seo_sitemap/models/m_seo_sitemap.erl
```diff
@@ -164,7 +168,7 @@ maybe_add_category_attrs(#{ category_id := CatId } = Map, Context) ->
                 undefined -> <<"weekly">>;
                 CF -> CF
             end,
-            Map1#{ changefreq := Freq };
+            Map1#{ changefreq => Freq };
```
robvandenbogaard, line 171: Good side-catch 😬

modules/mod_admin/templates/_admin_edit_content_seo.tpl
```diff
@@ -87,24 +87,54 @@
...
+                <div class="col-md-9 col-md-offset-3">
+                    <div class="checkbox">
+                        <label>
+                            <input id="seo_noindex"
+                                type="checkbox"
+                                name="seo_noindex"
+                                value="1"
+                                {% if not is_editable %}disabled{% endif %}
+                                {% if id.seo_noindex or is_seo_noindex_cat %}checked="checked"{% endif %}
+                                {% if is_seo_noindex_cat %}disabled{% endif %}
```
robvandenbogaard, line 101: Should this be merged with the condition at line 99? The duplication probably doesn't matter that much actually, in practice.

modules/mod_base/controllers/controller_file.erl
mworrell: I think so, yes.

modules/mod_seo_sitemap/models/m_seo_sitemap.erl
mworrell: Weird that this seemed to work, and compiled...

modules/mod_admin/templates/_admin_edit_content_seo.tpl
mworrell: Ah good one, indeed better to merge.

PR #3091, opened 2022-08-22
core: add option to exclude categories from the SEO indexing

mworrell:
### Description

This adds an option to exclude all pages in a category from indexing by a search engine.

For this the property `is_seo_noindex_cat` is set.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
See also https://github.com/zotonic/zotonic/pull/3089

PR #3096, opened 2022-08-23
Filter the redirect args of logon/logoff to ensure that it is a local redirect

mworrell:
### Description

This adds a check on all `p` and `page` arguments of various controllers to ensure that the redirect is a redirect to a local URL.

This does not break code that is redirecting within the site.

After review the code in master should also be checked and the `z_context:site_url/2` routine converted for master.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_translation/controllers/controller_language_set.erl
+++ b/modules/mod_translation/controllers/controller_language_set.erl
@@ -51,10 +51,10 @@ moved_temporarily(ReqData, Context0) ->
     Context = ?WM_REQ(ReqData, Context0),
     Context1 = mod_translation:set_user_language(z_context:get_q("code", Context), Context),
     Page = z_context:get_q("p", Context1),
-    Location = case z_utils:is_empty(Page) of
-                   true -> "/";
-                   false -> Page
-               end,
+    Location = case Page of
+        "/" ++ _ -> Page;
+        _ -> "/"
+    end,
     AbsUrl = z_context:abs_url(add_language(mod_translation:url_strip_language(Location), Context1), Context1),
     ?WM_REPLY({true, AbsUrl}, Context1).
 
--- a/modules/mod_translation/controllers/controller_language_set.erl
+++ b/modules/mod_translation/controllers/controller_language_set.erl
@@ -55,7 +55,8 @@ moved_temporarily(ReqData, Context0) ->
         "/" ++ _ -> Page;
         _ -> "/"
     end,
-    AbsUrl = z_context:abs_url(add_language(mod_translation:url_strip_language(Location), Context1), Context1),
+    Location1 = z_sanitize:uri(Location),
+    AbsUrl = z_context:abs_url(add_language(mod_translation:url_strip_language(Location1), Context1), Context1),
     ?WM_REPLY({true, AbsUrl}, Context1).
 
 moved_permanently(ReqData, Context) ->
--- a/src/support/z_context.erl
+++ b/src/support/z_context.erl
@@ -437,15 +437,15 @@ site_url(None, Context) when
     None =:= <<>> ->
     z_context:abs_url(<<"/">>, Context);
 site_url("#" ++ _ = Frag, _Context) ->
-    z_convert:to_binary(Frag);
+    z_sanitize:uri(z_convert:to_binary(Frag));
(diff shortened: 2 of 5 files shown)


modules/mod_translation/controllers/controller_language_set.erl
```diff
@@ -51,10 +51,10 @@ moved_temporarily(ReqData, Context0) ->
...
     Page = z_context:get_q("p", Context1),
-    Location = case z_utils:is_empty(Page) of
-                   true -> "/";
-                   false -> Page
-               end,
+    Location = case Page of
+        "/" ++ _ -> Page;
+        _ -> "/"
+    end,
     AbsUrl = z_context:abs_url(add_language(mod_translation:url_strip_language(Location), Context1), Context1),
```
robvandenbogaard, line 58: Shouldn't Location also be fed through a `z_context:site_url`? As to have it sanitized?

src/support/z_context.erl
```diff
@@ -400,7 +401,7 @@ prune_reqdata(ReqData) ->
 %% @spec abs_url(iolist(), Context) -> binary()
 abs_url(<<"http:", _/binary>> = Url, _Context) ->
     Url;
-abs_url(<<"https", _/binary>> = Url, _Context) ->
+abs_url(<<"https:", _/binary>> = Url, _Context) ->
     Url;
 abs_url(<<"//", _/binary>> = Url, Context) ->
     <<(site_protocol(Context))/binary, $:, Url/binary>>;
```
robvandenbogaard, line 407: I wonder if `abs_url` should sanitize. In its use elsewhere at least in some cases no sanitization is done, or am I overlooking something? E.g. line 443 `site_url("/" ++ _ = Path, Context) -> abs_url(Path, Context);`

src/support/z_context.erl
```diff
@@ -428,6 +429,41 @@ abs_url(Url, Context) ->
...
+            abs_url(Url1, Context);
+        Url1 ->
+            Path1 = case mochiweb_util:urlsplit(Url1) of
+                {_Scheme, _Netloc, Path, "", ""} ->
+                    Path;
+                {_Scheme, _Netloc, Path, Query, ""} ->
+                    Path ++ "?" ++ Query;
+                {_Scheme, _Netloc, Path, "", Frag} ->
+                    Path ++ "#" ++ Frag;
+                {_Scheme, _Netloc, Path, Query, Frag} ->
```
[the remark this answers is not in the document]
robvandenbogaard: What is your uptake on this, Marc?

src/support/z_context.erl
```diff
@@ -428,6 +429,41 @@ abs_url(Url, Context) ->
...
+                    Path ++ "?" ++ Query ++ "#" ++ Frag
+            end,
+            abs_url(Path1, Context)
+    end.
+
```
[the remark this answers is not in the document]
mworrell: Script URLs need to start with something like `javascript:`  (and variations on that because browsers tend to be lax about the syntax here).  So if we add the domain name in front `https://example.com/` then all should be ok (after the sanitize, you have to catch newlines in the URL, which otherwise can add extra headers to the HTTP response).

src/support/z_context.erl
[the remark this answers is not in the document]
mworrell: As this is a GET operation it won't be too bad, no data can be harmed.

In the master we have better checks on the domain name (different code there), so there we are more strict and replace the URL with `/` if the hostname does not match.

modules/mod_translation/controllers/controller_language_set.erl
mworrell: Might be better, indeed.  Or at least pass Location through the URI sanitizer.

src/support/z_context.erl
mworrell: I think the input should be sanitized by the caller, we use this for different formatted URLs (hex encoded, HTML escaped, XML escaped).

modules/mod_translation/controllers/controller_language_set.erl
mworrell: Added the sanitize.

PR #3098, opened 2022-08-25
core: add timer on pw checks against timing attacks for finding valid usernames

mworrell:
### Description

There is a timing difference between checking a password for valid usernames and invalid usernames.

The timing difference can be used to find valid usernames.

This change adds a timer around the pw check to make all pw checks about 300msec.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_identity.erl
+++ b/apps/zotonic_core/src/models/m_identity.erl
@@ -111,6 +111,12 @@
 
 -define(IDN_CACHE_TIME, 3600*12).
 
+%% Default duration and random variance interval for password checks.
+%% This prevents a timing difference between checks for existing and
+%% non existing accounts.
+-define(DEFAULT_PW_CHECK_DURATION, 280).
+-define(DEFAULT_PW_CHECK_VARIANCE, 40).
+
 
 %% @doc Fetch the value for the key from a model source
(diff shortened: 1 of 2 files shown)


mworrell:
Discussion is if this should have a base of 300 msec.

We can also measure the time needed for a typical password hash check, and then use that as a base for the timer.

On my dev machine a password hash takes around 350 msec:

```
(zotonic@PoToi)6> timer:tc(fun() -> erlpass:hash(<<"jdhjsadhidhihihhfs">>) end).
{351085,
 <<"$2a$12$2vty8V2M.g3iP1.4CtWFVutCOrlyNytqcOyxRklcCbxl//7oGiQ.2">>}
```

mmzeeman:
Performance varies indeed. My M1 MacBook takes 230 msec.

```erlang 
(zotonic@Tukkal)3> timer:tc(fun() -> erlpass:hash(<<"jdhjsadhidhihihhfs">>) end).
{232047,
 <<"$2a$12$us0N8kuE58n7GHFLrnD2au02rsrTlh3wuvrMlfwwq.7goK3MRTnhu">>}

```

mmzeeman (approved):
Looks like a good default pw-check duration. Our dedicated server is about 5 years old.

mworrell:
My dev machine is 4 years old, 2,2 GHz 6-Core Intel Core i7.

PR #3122, opened 2022-09-01
mod_auth2fa: add option to force setting 2FA before log on.

mworrell:
### Description

This adds an extra level of force to setting the 2FA code.

If enabled then the user MUST set their 2FA code before they can proceed with the log on.

After this has been merged, a master (1.x) version should be added.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


modules/mod_auth2fa/templates/_auth2fa_html_body.tpl
```diff
@@ -1,5 +1,7 @@
-{% with m.auth2fa.user_mode as mode %}
-    {% if mode == 2 or (mode == 1 and not m.auth2fa.is_totp_requested) %}
-        {% wire postback={request_2fa} delegate=`mod_auth2fa` %}
-    {% endif %}
-{% endwith %}
+{% if not m.auth2fa.is_totp_enabled %}
```
robvandenbogaard, line 1: Why should this only happen if totp is not enabled?

modules/mod_authentication/templates/_logon_login_admin_form_fields.tpl
```diff
@@ -1,24 +1,29 @@
...
+    </div>
 
-<div class="form-group">
-    <label for="password" class="control-label">{_ Password _}</label>
-    <input class="form-control" type="password" id="password" name="password" value="" autocomplete="off" />
-    {% validate id="password"
-        type={presence}
-        only_on_submit
-    %}
+    <div class="form-group username-password">
```
robvandenbogaard, line 11: The class username-password is also set on the parent, is this duplication intentional?

modules/mod_authentication/templates/_logon_reset_form_fields.tpl
```diff
@@ -20,7 +20,7 @@
 %}
 <div class="form-group">
     <label class="control-label" for="password_reset1">{_ New password _}</label>
-    <input class="form-control" type="password" id="password_reset1" name="password_reset1" value="" autocomplete="new-password" placeholder="{_ New password _}" />
+    <input class="form-control {% if not is_password_change %}do_autofocus{% endif %}" type="password" id="password_reset1" name="password_reset1" value="" autocomplete="new-password" placeholder="{_ New password _}" />
```
robvandenbogaard, line 23: Shouldn't it autofocus if is_password_change is true, instead the other way round?

modules/mod_auth2fa/templates/_auth2fa_html_body.tpl
mworrell: Not enabled for this user.  If the user as TOTP enabled then we don't need to ask.

modules/mod_auth2fa/mod_auth2fa.erl
```diff
@@ -121,5 +121,27 @@ observe_auth_postcheck(#auth_postcheck{ id = UserId, query_args = QueryArgs }, C
                     end
             end;
         false ->
-            undefined
+            % Could also have a POST of the new passcode secret to be set.
+            % In that case the passcode can be set for the user and 'undefined'
+            % returned.
```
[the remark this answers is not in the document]
mworrell: Yes (there is a check a bit more up in the code).  And this only gets called _after_ the username/password check.

modules/mod_authentication/templates/_logon_login_admin_form_fields.tpl
mworrell: Yes, can be removed.

modules/mod_authentication/templates/_logon_reset_form_fields.tpl
mworrell: If password change is true then there is another field to enter the current password, that field gets then the focus.

PR #3146, opened 2022-10-13
Fix -include{,_lib} attributes to unbreak doc chunk generation

erszcz:
This addresses [the issue mentioned on Slack](https://erlanger.slack.com/archives/C0SEMDD36/p1665592350066809), i.e. doc chunk generation with `rebar3 ex_doc` not working as expected. I'm not including `rebar3_ex_doc` in `rebar.config`, but the patch I tested with is just:

```
diff --git a/rebar.config b/rebar.config
index efd2bb3af..d117d883b 100644
--- a/rebar.config
+++ b/rebar.config
@@ -33,3 +33,5 @@
       % no_return
   ]}
 ]}.
+
+{project_plugins, [rebar3_ex_doc]}.
```

### DIFF
--- a/apps/zotonic_core/src/support/z_memo.erl
+++ b/apps/zotonic_core/src/support/z_memo.erl
@@ -34,7 +34,7 @@
 	delete/1
 ]).
 
--include_lib("include/zotonic.hrl").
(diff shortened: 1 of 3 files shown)


mworrell:
This is great, thank you so much!

PR #3160, opened 2022-10-31
mod_ssl_letsencrypt: ensure that connections to LE are closed

mworrell:
### Description

This fixes an issue where the connections to LE are left open after a certificate is fetched.

The use of shotgun is replaced with z_fetch (using httpc atm) and the code is simplified by removing some options.

The issue:

https://erlangforums.com/t/letsencrypt-activation-logs-warnings/1983

/cc @karlsson 

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Looks good to me.

PR #3190, opened 2022-11-22
mod_authentication: fix a problem with pw reset for 2FA enabled accounts

mworrell:
### Description

This fixes a problem where 2FA enabled accounts could not reset their password.

After the new password was entered the view changed to logon, making it impossible to reset the password.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
@mmzeeman This undos a change that you made in 7a93c1eb417caedf7325b0cd6be867a46c6110d4 we might need an additional change to fix your initial problem.

mmzeeman:
Can't remember why I also set `logon_view`.

mmzeeman (approved):

PR #3194, opened 2022-11-28
mod_wires: Remove 2 char limit on typeselect search

### LINKED ISSUE #3192: Typeselect action does not reset results when going from 3 to 2 characters
## Steps to reproduce

Setup the type select or use the media select in the admin

## Current behavior

Enter at least 3 characters in the input box, wait for results to appear, and remove a character. The previously found results are still visible.

## Expected behavior

No results should be available when less than 3 characters should be reset. When the user previously had 3 characters in the input field

## Thoughts

Why is the query only run after at least 3 characters are entered. This is behavior is probably easier to get right when the typeselect is always run when something is typed in the input box.

mmzeeman:
### Description

Fix #3192 3192

Removes the two character limit on type select searches. As a result, previously found resources are removed when there is no input.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
+++ b/apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
@@ -893,13 +893,9 @@ function z_typeselect(ElementId, postbackInfo)
 
     z_input_updater = setTimeout(function()
     {
-        var obj = $('#'+ElementId);
-
-        if(obj.val().length >= 2)
-        {
-            obj.addClass('loading');
-            z_queue_postback(ElementId, postbackInfo);
-        }
+        const obj = $('#'+ElementId);
+        obj.addClass('loading');
+        z_queue_postback(ElementId, postbackInfo);
     }, 400);
 }
 
--- a/doc/ref/actions/action_typeselect.rst
+++ b/doc/ref/actions/action_typeselect.rst
@@ -9,14 +9,14 @@ Performs a search for the typed text whilst typing in an input field.  Shows pos
 Example::
 
    <form method="get" action="/search">
-     <input type="text" id="person" name="person" value="" />
+     <input type="search" id="person" name="person" value="" />
      <ul id="suggestions"></ul>
      <input type="hidden" id="person_id" value="" />
-     {% wire id="person" type="keyup"
+     {% wire id="person" type="input"
              action={typeselect cat="person" 
-                               target="suggestions"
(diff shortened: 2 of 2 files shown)


mworrell (approved):

apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
```diff
@@ -893,13 +893,9 @@ function z_typeselect(ElementId, postbackInfo)
 
     z_input_updater = setTimeout(function()
     {
-        var obj = $('#'+ElementId);
-
-        if(obj.val().length >= 2)
```
mworrell, line 898: Maybe make it >= 1 ?

Then we don't post on 0 chars.

mworrell (approved):

apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
mmzeeman: But then it won't reset the search results when you go from 1 to 0, which is what users expect.

mworrell: Should we return the empty result for the empty query?  Or just anything that matches?

mmzeeman: The example code in the documentation is also interesting. A commented out wire statement.

Screen Shot 2022-11-28 at 12 33 03

mmzeeman: Side issue.. the color scheme for example code on zotonic.com makes it look like wire statements are comments.

Screen Shot 2022-11-28 at 12 38 03

mmzeeman: Ah.. `%`

mmzeeman: The typeaction uses the autocomplete query. This returns 0 result without querying the database. 

I've updated the documentation to correctly use an input type="search" input box. This makes sure a "Search" button appears on the keyboard of an iOS device. On Safari an Chrome the search box can also be cleared by typing <kbd>Esc</kbd> or clicking on a small cross appearing at the right of the input box. For the clickable cross the type of the wire must be changed from `keyup` to `input`.

PR #3196, opened 2022-11-28
[Search] Refactor search view

DorienD:
### Description

Search view JS now listens to a `data-attribute` so it's a choice to use the search view

- Cleanup css a bit

### DIFF
--- a/apps/zotonic_mod_admin/priv/templates/_admin_menu.tpl
+++ b/apps/zotonic_mod_admin/priv/templates/_admin_menu.tpl
@@ -82,7 +82,7 @@
                 <form class="navbar-right navbar-form form-inline" action="{% block search_target %}{% url admin_overview_rsc %}{% endblock %}" method="get">
                     <input type="hidden" name="qcat" value="{{ q.qcat|escape }}" />
                     <input type="hidden" name="qquery_id" value="{{ q.qquery_id|escape }}" />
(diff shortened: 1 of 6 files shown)


apps/zotonic_mod_admin/priv/templates/_admin_menu.tpl
```diff
@@ -82,7 +82,7 @@
                 <form class="navbar-right navbar-form form-inline" action="{% block search_target %}{% url admin_overview_rsc %}{% endblock %}" method="get">
                     <input type="hidden" name="qcat" value="{{ q.qcat|escape }}" />
                     <input type="hidden" name="qquery_id" value="{{ q.qquery_id|escape }}" />
-                    <input autocomplete="off" class="search-query form-control" type="text" name="qs" value="{{q.qs|escape}}" placeholder="{% block search_placeholder %}{_ Search... _}{% endblock %}" />
+                    <input autocomplete="off" class="search-query form-control" type="text" name="qs" value="{{q.qs|escape}}" placeholder="{% block search_placeholder %}{_ Search... _}{% endblock %}" data-search-view="true" />
```
mmzeeman, line 85: When a `type="search"` is used, the return button on mobile devices changes to "Search". On desktop devices the input box gets an extra reset button and it is possible to clear out any typed input with the escape key.

mmzeeman (approved):

apps/zotonic_mod_admin/priv/templates/_admin_menu.tpl
DorienD: Good point! I will add this.

mworrell (approved):

PR #3200, opened 2022-12-03
mod_base: Fix an issue where controller_page could enter in a redirect loop.

mworrell:
### Description

This could happen if a resource has a page path with a space (encoded as %20). The reconstitution of the page path did not account for non url-safe characters.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):

mworrell:
I think the we also could check de re-concatenated path with the decoded request path. This might be more robust.

PR #3211, opened 2022-12-14
Tune memory usage on 0.x

mworrell:
### Description

Slightly better memory usage by using forced gc

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/src/support/z_module_indexer.erl
+++ b/src/support/z_module_indexer.erl
@@ -56,6 +56,7 @@
 -include("zotonic.hrl").
 
 -define(TIMEOUT, infinity).
+-define(GC_TIMEOUT, 1000).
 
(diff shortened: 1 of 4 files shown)


robvandenbogaard:
I am wondering if timeouts are triggered by the garbage collection taking a lot of time; in that case invoking garbage_collect would not help because it is already running?

However, if it is memory allocation taking a lot of time, because of swap use for instance, then it is indeed sensible to do a garbage collect.

Remaining point of concern would be dropping the timed out tasks?

mworrell:
The module index is quite big.  We might want to check if all paths are binaries and if there is data that doesn't need to be used.

PR #3217, opened 2022-12-20
mod_search_query: fix a problem with filtering facets

mworrell:
### Description

The wrong table was linked.

Added extra logging to notify if a filter term could not be added.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

PR #3220, opened 2022-12-21
mod_search: fix a problem with joining search_facet table

mworrell:
### Description

Fix #3217

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_search/src/support/search_query.erl
+++ b/apps/zotonic_mod_search/src/support/search_query.erl
@@ -1401,7 +1401,7 @@ map_filter_column(<<"pivot.", P/binary>>, #search_sql_term{ join_inner = Join } 
 map_filter_column(<<"facet.", P/binary>>, #search_sql_term{ join_inner = Join } = Q) ->
     Q1 = Q#search_sql_term{
         join_inner = Join#{
-            <<"search_facet">> => {<<"facet">>, <<"facet.id = rsc.id">>}
+            <<"facet">> => {<<"search_facet">>, <<"facet.id = rsc.id">>}
         }
     },
     Field = sql_safe(P),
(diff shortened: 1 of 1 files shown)


karlsson:
Sorry, this does not work either:
```
ERROR  site=skolan  dispatch=page <0.2412.0> [z_db:q/4:645] ▸ text="z_db error in query" result=error reason="{error,error,<<\"42P01\">>,undefined_table,<<\"invalid reference to FROM-clause entry for table \\\"search_facet\\\"\">>,[{file,<<\"parse_relation.c\">>},{hint,<<\"Perhaps you meant to reference the table alias \\\"facet\\\".\">>},{line,<<\"3574\">>},{position,<<\"290\">>},{routine,<<\"errorMissingRTE\">>},{severity,<<\"ERROR\">>}]}" query="select rsc.id from rsc rsc join search_facet facet on facet.id = rsc.id where  (rsc.content_group_id is null or rsc.content_group_id = any($2::int[])) and rsc.is_published = true and rsc.publication_start <= now() and rsc.publication_end >= now() and rsc.pivot_category_nr = 100003000 and search_facet.f_su_status = $1  order by rsc.pivot_title ASC, rsc.created DESC offset $3 limit $4" in=zotonic_core args="[<<\"Aktiv\">>,[316,317],0,200]"
```
I noticed it is the same as in add_order function which is for sorting (I guess). and this works....

karlsson:
When I added search_facet as column in all places I checked the map_filter_column clause for pivot handling just before. This uses the same table name everywhere. If I do the same for the facet it workes too. Although it is a bit annoying that you cannot use alias as for the add_order...
```
map_filter_column(<<"facet.", P/binary>>, #search_sql_term{ join_inner = Join } = Q) ->
    T = <<"search_facet">>,
    Q1 = Q#search_sql_term{
        join_inner = Join#{
            T => {T, <<T/binary, ".id = rsc.id">>}
        }
    },
    Field = sql_safe(P),
    {T, <<"f_", Field/binary>>, Q1};
```

mworrell:
Ah, I see what goes wrong here, confusion between the alias and the table name.
Needs a bit better fix.

karlsson:
I checked out the branch and it works fine now.
One small compiler warning
zotonic_mod_search/src/support/search_query.erl:1381:15: Warning: variable 'Tab' is unused
% 1381| create_filter(Tab, Alias, Col, Operator, Value, Q) ->
%     |               ^

karlsson:
Thanks @mworrell . It works fine from master too now.
I will not report anything more until after Christmas :smile:

mworrell:
Thanks for reporting and figuring this out!

PR #3222, opened 2022-12-21
mod_admin: add admin notes; mailinglist opt out

mworrell:
### Description

This adds a notes field to the admin that is only visible for admin users that can edit the resource.

Also add an optional mailinglist opt-out flag.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Good idea... 👍

PR #3223, opened 2022-12-21
mod_mailinglist: allow multiple mailinglists subscribe forms on one page

### LINKED ISSUE #3221: Mod_mailinglist: multiple signup forms on 1 page
## `Zotonic` version
Master

## Steps to reproduce
Put multiple mailinglist subscribe scomps on 1 page

## Current behavior
In the `scomp_mailinglist_mailinglist_subscribe.erl:65` we use a unique id to update to the thank you or error text. But it's not unique enough on 1 page.

mworrell:
### Description

Make the element-id's unique for every form

Fix #3221

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@DorienD Can you check if this is ok for your problem?

mworrell:
@DorienD I merged it so that we can proceed with the rc release - if any problems then let me know.

Issue #2884, opened 2022-02-14
Investigate use of modern image formats in {% image %} tag


mmzeeman:
When you use chrome's lighthouse tool they always suggest to serve `webp` or `avid` image files.

Images are usually served with the `{% image %}` tag and we use imagemagick to crop, rotate and manipulate images via the {% image %} tag. It will probably be a nice addition to also serve images as `webp`.

There are a couple of options.

1) Use the `picture` tag and provide `webp`, `jpg` alternatives.
2) Make use of content negotiation, and serve webp variants when the client supports it. It may be a bit strange to return webp when the extension is jpg or png. Then it is a variant of the original resource, just like is intended with content negotiation.

More information:

https://www.smashingmagazine.com/2019/10/speed-up-your-website-webp/

seannaswell:
When architecting a solution, you might considering also adding `avif` as an option, as it is the likely successor of `webp`: [https://avif.io/blog/comparisons/avif-vs-webp/](https://avif.io/blog/comparisons/avif-vs-webp/)

It's not available everywhere yet, but there are discussions that Apple is in the process of adding support now.

Apple Support: [https://www.reddit.com/r/AV1/comments/m8ij2s/safari_adding_avif_support_soon/](https://www.reddit.com/r/AV1/comments/m8ij2s/safari_adding_avif_support_soon/)
Current Support: [https://caniuse.com/?search=avif](https://caniuse.com/?search=avif)

mworrell:
We could generate a picture tag with all these options. And then generate the needed images on demand.  Just wondering if this matrix of formats and sizes isn't a bit large?

mworrell:
I like the 10bit depth and HDR of AVIF, and the larger sizes.  We are handling images of 20Kx10K pixels, which apparently don't fit in WEBP with their max 16Kx16K pixels.

Maybe, as a stop-gap while we figure this out, we can add progressive rendering to all our JPEG preview images. This could be done by adding `-interlace plane` to the convert command used to make the JPEGs.

seannaswell:
Would adding progressive rendering eliminate the need for [blurhash](https://github.com/zotonic/zotonic/issues/2385)?

mworrell:
Not completely, blurhash can be used to:

 * Show a placeholder for lazy loaded images
 * Show color around the image that matches the image color

The problem with blurhash is that we need JS to show it.  A shame that `lowsrc` is not supported anymore.

mworrell:
[...]

PR #3226, opened 2022-12-23
core: Support wepb in image tag

mmzeeman:
### Description

See #2884 

This PR allows the image tag to produce webp formatted images. See: https://developers.google.com/speed/webp

Modern image formats are often recommended by Lighthouse. This makes
it possible to easily enable it.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_core/src/support/z_media_preview.erl
```diff
@@ -387,13 +387,19 @@ out_mime(Mime, Options, Context) ->
 %% @doc Return the preferred mime type of the image generated by resizing an image of a certain type and size.
 -spec out_mime( InMime :: binary(), list() ) -> { OutMime :: binary(), Extension :: string()}.
 out_mime(Mime, Options) ->
-    out_mime1(get_lossless_value(Options), Mime).
+    Mime1 = case proplists:get_value(webp, Options, false) of
```
mworrell, line 390: I wonder if we should make this a "format" option to enable more formats in the future?

mworrell (approved):

apps/zotonic_core/src/support/z_media_preview.erl
mmzeeman: Ah.. yes... that would be more future proof indeed.

mworrell: We can map an extension to mime format, using the `[Mime|_] = mimetypes:ext_to_mimes(<<"webp">>)` or just use a list of supported formats of which we know it works, with a fallback to webp or jpeg.

I think the list of known formats might be best.

mmzeeman: Right now the "lossless" option decides if the extension will be png or jpg depending on the setting. Right now I added webp to this in a backward compatible way. Modern image formats have the possibility to choose between lossy and lossless compression. We might want to change this configuration setting for `Z 1.1`.

PR #3234, opened 2023-01-03
mod_admin_modules: show version of modules in modules overview

mworrell:
### Description

Add `z_module_manager:mod_info/1` to obtain information about a module.

Show version of modules in admin/modules list.

After this has been merged we can port it back to the 0.x

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_module_manager.erl
+++ b/apps/zotonic_core/src/support/z_module_manager.erl
@@ -670,7 +670,7 @@ module_exists(M) ->
         schema := integer() | undefined,
         title := binary() | undefined,
         description := binary() | undefined,
-        app_dir := filename:filename_all()
+        app_dir := filename:filename_all() | undefined
     }.
 mod_info(Module) ->
     App = module_to_app(Module),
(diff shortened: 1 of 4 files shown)


apps/zotonic_core/src/support/z_module_manager.erl
```diff
@@ -634,21 +660,138 @@ module_exists(M) ->
...
 
+%% @doc Get the description of a module.
+-spec mod_description(Module) -> Desc when
+    Module :: atom() | binary() | string(),
+    Desc :: binary() | undefined.
+mod_description(Module) ->
+    Mod = module_to_mod(Module),
+    try
+        Desc = proplists:get_value(mod_description, Mod:module_info(attributes), <<>>),
+        unicode:characters_to_binary(Desc)
```
robvandenbogaard, line 768: Should author etc also use `unicode:characters_to_binary`?

robvandenbogaard (approved):

apps/zotonic_core/src/support/z_module_manager.erl
mworrell: Good point.  With something extra, as the author is sometimes an atom in Erlang modules...

PR #3249, opened 2023-01-12
mod_admin_modules: add warning if not all modules are running.

### LINKED ISSUE #3045: Huge warning when required modules are inactive [Merge 0.x into master]
## `Zotonic` version
0.x and 1.x

## `OS` version
Irrelevant

## Description
* **Motivation**
We keep finding issues on websites that are caused by some modules being inactive. It happens so often that checking if all the required modules are on should be the first step in our debugging process. Checking this is pretty easy, but still we often forget about it, because this problem can cause such a great variety of issues (where we think something else is going on) and modules can go down at unexpected times (or for example: you still need to start a module that you just added).
In the end, we still spend a significant amount of time debugging issues caused by modules being inactive.

* **Proposal**
To make it a bit easier to notice, I would like a HUGE WARNING (okay, that's exaggerated, but large enough that you can't miss it) to appear in the admin dashboard when one of the required modules is inactive.
Additionally, I would like to have a warning about this in the menu bar, so you can see it from any page.

## Current behavior
Zotonic behaves incorrectly because of required modules being inactive without any warning.

## Expected behavior
Zotonic gives warnings in the admin interface when required modules are inactive.

mworrell:
### Description

Fix #3045

Show a warning on the admin dashboard and modules pages when modules are not running.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_admin/models/m_admin_status.erl
+++ b/modules/mod_admin/models/m_admin_status.erl
@@ -63,6 +63,22 @@ m_find_value(unused, #m{value=memory}, _Context) ->
 m_find_value(usage, #m{value=memory}, _Context) ->
     recon_alloc:memory(usage);
 
+% modules
+m_find_value(modules, #m{value=undefined} = M, Context) ->
+    M#m{value=modules};
+m_find_value(down, #m{value=modules}, Context) ->
+    Status = z_module_manager:get_modules_status(Context),
+    lists:flatten([
+        [ Module || {Module, _, _Pid, _Date} <- Specs ]
+        || {State, Specs} <- Status, State =/= running
+    ]);
+m_find_value(up, #m{value=modules}, Context) ->
+    Status = z_module_manager:get_modules_status(Context),
+    lists:flatten([
+        [ Module || {Module, _, _Pid, _Date} <- Specs ]
+        || {running, Specs} <- Status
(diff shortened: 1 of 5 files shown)


modules/mod_admin/models/m_admin_status.erl
```diff
@@ -63,6 +63,22 @@ m_find_value(unused, #m{value=memory}, _Context) ->
...
     recon_alloc:memory(usage);
 
+% modules
+m_find_value(modules, #m{value=undefined} = M, Context) ->
+    M#m{value=modules};
+m_find_value(down, #m{value=modules}, Context) ->
+    Status = z_module_manager:get_modules_status(Context),
+    lists:flatten([
+        [ Module || {Module, _, _Pid, _Date} <- Specs ]
+        || {State, Specs} <- Status, State =/= running
```
robvandenbogaard, line 73: Is it not possible to do this with `[Module || {State, {Module, _, _Pid, _Date}} <- Status, State =/= running]`?

mworrell: No, per State we have list, so we end up with a nested list comprehension.

robvandenbogaard (approved):

PR #3250, opened 2023-01-12
mod_video: fix a problem where preview generation would crash on audio-only videos.

mworrell:
### Description

The preview generation would crash because the dimensions of the video could not be fetched.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Yes, videos are just containers with streams. And this one missed the video stream. Which makes it a bit hard to generate an image from the stream...

PR #3253, opened 2023-01-13
Manage system dependencies with Nix

cdfa:
### Description

This PR adds support for installing Zotonic's system dependencies using [Nix](nixos.org/). I think it's great for ease of installation and reproducibility.
The configuration is quite simple: the dependencies are listed in `shell.nix` and when you enter the shell using `nix-shell` the relevant environment variables (like `$PATH`) get set, so the packages become available. It should override any other version installed on your system, but only when you're in the shell. Otherwise, it does not conflict with anything (you could have dependencies installed with Nix next to those installed with your standard package manager on the same machine).

The dependency versions are pinned by a specific version of the Nix package set defined in `nix/nixpkgs.nix`. We can override the package set with the `nixpkgs` argument in `shell.nix` and this is what we do in the over-arching Nix configuration at Driebit.

If you're interested in merging, I'll also add it to the installation instructions in the documentation.
I'm not sure anymore where the changes in the lock file came from. I think it happened when I was trying to get things running with Marc, but I'm not sure if it should stay.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [X] no BC breaks


rebar.lock
```diff
@@ -69,6 +69,7 @@
  {<<"unicode_util_compat">>,{pkg,<<"unicode_util_compat">>,<<"0.7.0">>},2},
  {<<"webdavfilez">>,{pkg,<<"webdavfilez">>,<<"1.2.1">>},0},
  {<<"yamerl">>,{pkg,<<"yamerl">>,<<"0.8.1">>},0},
+ {<<"zotonic_apps">>,{pkg,<<"zotonic_apps">>,<<"1.0.0-rc.10">>},0},
```
mworrell, line 72: As this is the master Zotonic repo, do we need this dependency?

cdfa: These get added on `make` on my machine, because I have a project in `apps_user` (atlas) that has `zotonic_apps` as dependency. We probably do not want site dependencies to end up in here. Do you know any way to fix that?

mworrell: I tend to update the zotonic master dependencies in a separate directory. This prevents that development apps are coming along.

cdfa:
Regarding the python dependencies for the docs: I'll add python and pip in a second.

If you want to prevent cases like a few days ago where the CI fails because python packages changed, we could also manage the python packages with Nix. We could choose to keep the standard `requirements.txt` and generate a nix expression from that with something like https://github.com/DavHau/mach-nix, or we could use the python packages from nixpkgs.
The first option would at least give you a warning when the package has changed on Pypi (instead of just downloading something else which I think you said happened), but the second option would have prevented the issues altogether, because nix packages always stay the same (until you update the package set). This option also makes it easier to upgrade, since it is a rolling release package set.
However, it would require you to use Nix for building the docs (the `requirements.txt` would be replaced by a .nix file) and there are some version discrepancies between the versions used now and those in nixpkgs: sphinx 1.5.3 vs 5.1.1 in nixpkgs and sphinx-rtd-theme 0.2.4 vs 1.0.0 in nixpkgs. So we might need to adapt to some breaking changes if we choose this route.

Or we can leave it as is and only have python and pip in nix 🤷.

robvandenbogaard (approved):

PR #3259, opened 2023-01-13
mod_cookie_consent: new modules for asking consent to place cookies

### LINKED ISSUE #3188: Adapt mod_cookie_consent for 1.x
## Description

Adapt the mod_cookie_consent and changes from 0.x https://github.com/zotonic/zotonic/pull/3184 to 1.x

mworrell:
### Description

Fix #3188

This module provides:

 - A dialog to ask consent to place cookies
 - Load 3rd-party content if cookies are accepted

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


DorienD:
I think you're missing the `_html_head.tpl` cause the css is not included yet.

mworrell:
Good one, and with http2 there is no reason anymore to force bundling all css, so we can add a single include.

Shall I split it again in a `_html_head.tpl` with an included `_html_head_cookie_consent.tpl`?  Then it will be possible to overrule the second one when needed.

DorienD:
Yes you are correct with http2 we don't need the bundeling anymore, bundling would be an anti pattern. And i would like to be able to overrule the second one.

mworrell:
I have added a promise so that the cookie_consent.js can be loaded before jQuery and zotonic-wires.js.  It is resolved at the bottom of the zotonic-wires.js.

apps/zotonic_mod_cookie_consent/priv/templates/_media_cookie_consent.tpl
```diff
@@ -0,0 +1,29 @@
...
+             title=_"Click to change cookie settings"
+    %}
+    <p>
+        {_ Please consent to cookies to display external content. _}
+        <a href="#cookie-consent" id="{{ #consent }}" class="cookie-consent-change">{_ Change cookie settings _}</a>
+    </p>
+
+    
+    {% javascript %}
+    alert("bliep");
```
DorienD, line 18: @mworrell i can't seem to let this alert show itself, do you have a hint? Maybe because this template is inserted by the mod_cookie_consent.erl?

mworrell: Indeed, the HTML is used, the actions are discarded.
I have added comments to make that clear.

DorienD:
I've added a focus trough javascript, i couldn't get the action to work in the cookie-consent-banner action. So now i've worked with a `settimeout` function which i do not prefer to use. @mworrell could you maybe check why this doesn't work:

```
{% wire name="cookie-consent-banner"
    action={
            insert_top 
            target=" body" 
            template="_cookie_consent.tpl"
            action={script script="document.getElementById('cookie-consent').focus()"}
    }
%}
```

mworrell:
That second action will not be executed.

You could add a "do_focus" class on the input field. Then the focus widget will handle it. (Have to include the js, I think z.focus.js - am now not at my machine)

DorienD:
Found a different solution :)

DorienD (approved):

PR #3264, opened 2023-01-18
mod_survey: in survey results editor show user and allow to link or create new person

mworrell:
### Description

In the survey results editor we now show which user submitted the survey (if any).

There is also an option to create a new `person` using the entered answers or link to an existing person based on the entered email address, name or address.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


DorienD (changes requested):
Creating a person works well. Could you check these things:

- Add a title based on the filled name
- Could we add the language based on the what the user used in its browser? When filling in the survey?

mworrell:
Good one, we actually don't know the language. Maybe we should store it as well with the result, then we can also use it to set the preferred language of the new person record.

PR #3272, opened 2023-01-25
Amazing changes for admin frontend edit

DorienD:
- Add an editor for different options in frontend edit
- Limit options in Tinymce
- Add a dialog for image editing

This PR will let the users simply edit their images in page in stead of losing context. The editor has been limited in options. 

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_mod_admin_frontend/priv/templates/_admin_frontend_editor.tpl
```diff
@@ -0,0 +1,62 @@
...
+        tab="find"
+        callback="window.zAdminLinkDone"
+        center=0
+        autoclose
+        tabs_enabled=["find", "upload"]
+    }
+%}
+
+{% with
+    "5.10.2",
```
mworrell, line 44: Is there a way we can get this version from the mod_editor_tinymce?

mworrell: I am giving it a try, via the editor_tinymce model

mworrell:
I have added a method to fetch the "real" tinymce version.

Also changed the tabs of the frontend zlink dialog to exclude "new", instead of including "find" and "upload".  Is that ok? Or did you really want to exclude downloading files and embeds from the frontend admin?

mworrell (approved):

DorienD:
Yes! I didn't know we had an exclude. I like it. I've fixed an error in the admin_editor.tpl and now everything is gucci.

mworrell:
And merged 🥳

PR #3293, opened 2023-02-16
mod_image_edit: add some tweaks for the modal version, remove rotate …

DorienD:
Reduced the space the settings take, added a max height for the image so portrait images stay in sight. Removed the rotate button for now, because this works buggy in the modal.


mworrell:
That is the 90 degrees rotate button?

I will have to check why this gives problems, as it is used in some sites.

DorienD:
Yes, the calculating of the 'new' width and height are weird because in the modal i don't want the image to be displayed in a smaller version than its original width/height. So the settings are instantly visible.

mworrell:
Ah, the 90 degree rotation is indeed applied before the width/height resizing.
It is an addition to the image rotation in the EXIF attributes.

Though the image shown is the full-size image without any changes in the aspect ratio.  So I think applying the rotation should be ok.

mworrell:
@DorienD Can you maybe rebase?  Apparently some of the css/tpl was also changed in master.

PR #3298, opened 2023-02-17
mod_admin: add basic block view templates

### LINKED ISSUE #3297: Re-add the basic block view templates
## `Zotonic` version

1.0.0-rc.15

## Description

In the admin we can add blocks, and we have edit templates for those blocks.
However we don't have default view templates for the blocks.

Proposal:

- Add block view templates
- Add them to mod_admin, as that module also defined the edit templates
- Also add a default _blocks.tpl to mod_base

mworrell:
### Description

Fix #3297

Add view templates for the block definitions in the admin.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@DorienD Agree with these basic templates?

PR #3304, opened 2023-02-22
Add defer argument to lib tag

williamthome:
### Description

This PR adds `defer` argument to the `lib` tag.
The `defer` first will download the script file and then wait for HTML parsing. In other words, the script will execute after `domInteractive` and before `domContentLoaded` events.

image

**NOTE**

The `async` argument was changed to always set also the defer, as specified by the [W3 documentation](https://www.w3.org/TR/2011/WD-html5-20110525/scripting-1.html#attr-script-async):

> The [defer](https://www.w3.org/TR/2011/WD-html5-20110525/scripting-1.html#attr-script-defer) attribute may be specified even if the [async](https://www.w3.org/TR/2011/WD-html5-20110525/scripting-1.html#attr-script-async) attribute is specified, to cause legacy Web browsers that only support [defer](https://www.w3.org/TR/2011/WD-html5-20110525/scripting-1.html#attr-script-defer) (and not [async](https://www.w3.org/TR/2011/WD-html5-20110525/scripting-1.html#attr-script-async)) to fall back to the [defer](https://www.w3.org/TR/2011/WD-html5-20110525/scripting-1.html#attr-script-defer) behavior instead of the synchronous blocking behavior that is the default.

### Checklist

- [X] documentation updated
- [ ] tests added
- [X] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_lib_include.erl
+++ b/apps/zotonic_core/src/support/z_lib_include.erl
@@ -38,6 +38,8 @@
                 | {minify, boolean()}
                 | async
                 | {async, boolean()}
+                | defer
+                | {defer, boolean()}
                 | absolute_url
                 | {absolute_url, boolean()}
                 | {media, binary()}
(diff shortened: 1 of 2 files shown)


mworrell:
Good addition, and with clear explanations, thank you!

mworrell (approved):

PR #3307, opened 2023-02-23
core: change the signature algorithm for payloads to hmac-sha256

mworrell:
### Description

The postback payloads and image URLs were signed using md5.
As that algorithm has known weaknesses this changes it to hmac-sha256. 

The encoding is also changed from hex to base64url.

Old image URLs will be invalid after this change. Unless the image URL only uses a mediaclass, in which case the URL is still valid as there isn't a signature in those URLs.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_utils.erl
+++ b/apps/zotonic_core/src/support/z_utils.erl
@@ -93,6 +93,10 @@
     ensure_existing_module/1
 ]).
 
+-define(HASH_ALGORITHM, sha256).
+-define(HASH_BYTES, 32).
+
+
 %% @doc Apply a list of functions to a startlist of arguments.
 %% All functions must return: ok | {ok, term()} | {error, term()}.
 %% Execution stops if a function returns an error tuple.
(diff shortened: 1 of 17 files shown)


apps/zotonic_mod_authentication/src/models/m_authentication.erl
```diff
@@ -389,7 +389,7 @@ decode_token(Token, Context) ->
         {Type, UserId, Timestamp} = decode_payload(Payload),
         SiteSecret = site_auth_key(Context),
         UserSecret = user_auth_key(UserId, Context),
-        HashCheck = z_utils:hmac(sha256, <<SiteSecret/binary, UserSecret/binary>>, Payload),
+        HashCheck = crypto:mac(hmac, sha256, <<SiteSecret/binary, UserSecret/binary>>, Payload),
```
robvandenbogaard, line 392: Should we use a macro for `sha256` here too? Or maybe the use of the macro in `z_utils` is not needed; I am wondering if it is clear enough to just type `sha256` everywhere and if we want to upgrade it, it makes for an easy search and replace (while improving readability between upgrades).

robvandenbogaard: On second thought, it might be nice to have a z_crypto module or similar, where constants like these are listed - it would offer a nice overview on which algorithms are used for the respective uses. (sha256 for pickling, sha for 2fa for instance)

apps/zotonic_mod_auth2fa/src/support/z_auth2fa_qrcode_demo.erl
```diff
@@ -110,7 +110,7 @@ totp(Key, Period) ->
 
 %% @doc RFC-4226 "HOTP: An HMAC-Based One-Time Password Algorithm" http://tools.ietf.org/html/rfc4226
 hotp(Key, Count) when is_binary(Key), is_integer(Count) ->
-	HS = z_utils:hmac(sha, Key, <<Count:64>>),
+	HS = crypto:mac(hmac, sha, Key, <<Count:64>>),
```
robvandenbogaard, line 113: Should these ones stay like this or also moved to `z_crypto`? I think I can also understand why it should stay here.

apps/zotonic_core/src/support/z_utils.erl
```diff
@@ -324,93 +323,91 @@ erase_process_dict() ->
...
-    catch
-        _M:_E ->
-            ?LOG_ERROR(#{
-                text => <<"Postback data invalid, could not depickle">>,
-                in => zotonic_core,
-                data => Data
-            }),
-            erlang:throw({checksum_invalid, Data})
-    end.
+    z_crypto:depickle(Data, Context).
```
robvandenbogaard, line 406: Nice to have backwards compatibility 👍

apps/zotonic_mod_auth2fa/src/support/z_auth2fa_qrcode_demo.erl
mworrell: This is the fixed algorithm, so better keep it here.

mmzeeman (approved):
👍 this makes protocol changes easier in the future.

robvandenbogaard (approved):

PR #3308, opened 2023-02-27
Make mod_cookie_consent selector slightly more general

cdfa:
### Description

This allows strict frontend frameworks like Elm to replace the `<script>` with some other element and still have it work. Ideally, we would use a `<template>` instead of a `<script>`, but that also causes problems in Elm at the moment: https://github.com/elm/virtual-dom/issues/185

### Checklist

- [ ] documentation updated
- [ ] tests added
- [X] no BC breaks


mworrell:
Master version in https://github.com/zotonic/zotonic/pull/3309

PR #3314, opened 2023-03-02
Add keyboard to dialog options

williamthome:
### Description

This PR adds the `keyboard` option to prevent dialog closing by pressing the escape key.
Reference: https://getbootstrap.com/docs/4.4/components/modal/#options

### Checklist

- [X] documentation updated
- [ ] tests added
- [X] no BC breaks


mworrell:
Sounds good to me, in due time we can also add more keyboard shortcuts to this flag.

PR #3321, opened 2023-03-13
mod_search: Fix mismatch between sql and facet float type

karlsson:
Changed type from "float" to "double precision" when creating
  search_facet table as this is what postgresql returns.

### Description

Fixes comparison error at search_facet table creation due to mismatch  between search_facet "native" type (float) and sql column type (double precision). 
Tested by adding .._float type to facet.tpl template and successfully generating table with double precision column.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_search/src/support/search_facet.erl
+++ b/apps/zotonic_mod_search/src/support/search_facet.erl
@@ -965,7 +965,7 @@ facet_to_column(#facet_def{
 
 col_type(text) -> <<"character varying">>;
 col_type(integer) -> <<"integer">>;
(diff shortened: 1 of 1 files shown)


mworrell:
Thank you!

karlsson:
Thanks for merging @mworrell !
I have updated my [site](https://skolan.creado.se/sv/page/340/skolor?qfacet.su_status=Aktiv&qgyr_weighted=none&q_weighted=25&qsu_title=), and though I am just passing html around the turnaround time is really impressive when I change the facet filter.

PR #3322, opened 2023-03-14
core: on rsc update error for expected version, return error (0.x)

mworrell:
### Description

Do not throw an exception, as this expectation should be programmatically handled.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Ook prima.

mworrell:
In the master branch we don't throw exceptions anyway.
And in the case of failed expectations it is better to have return values, so it is possible to handle the failure in a programmatic way.

PR #3326, opened 2023-03-16
Change the language setting by changing the z.lang cookie setting

### LINKED ISSUE #2347: I can not change the language
set the language, but still everything is in English

mmzeeman:
### Description

Fix #2347

When you don't use the url rewrite method to change the language, zotonic relies on the `z.lang` cookie. This cookie was not changed when a language switch is made. 

I updated the `z_reload` function from `zotonic-wired.js` to also change the language cookie when the language switch happens.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
+++ b/apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
@@ -624,67 +624,81 @@ function z_reload(args)
         parts;
 
     if (page.length > 0 && page.val() !== "" && page.val() !== '#reload') {
-        window.location.href = window.location.protocol
+        window.location.href = `${ window.location.protocol }//${ window.location.host }${ page.val() }`;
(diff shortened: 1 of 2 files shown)


apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
```diff
@@ -624,67 +624,81 @@ function z_reload(args)
...
+        window.location.href = `${ window.location.protocol }//${ window.location.host }${ page.val() }`;
+        return;
+    } 
+
+    if (typeof args === "undefined") {
+        window.location.reload(true);
+        return;
+    }
+
+    newLanguage = args.z_language;
```
mworrell, line 636: Maybe `const newLanguage = ...` ?

mworrell (approved):

PR #3335, opened 2023-03-18
Added a otp and zotonic version number message to make all

### LINKED ISSUE #3329: Make doesn't complete on Mac OS zsh
## `Zotonic` version
[0.75.0](https://github.com/zotonic/zotonic/releases/tag/0.75.0)

## `OS` / `Browser` version
Mac OS 13.2.1

## Steps to reproduce
git clone https://github.com/zotonic/zotonic/releases/tag/0.75.0
cd zotonic
make

## Current behavior
```
bin/zotonic completion
~/.bash_profile does not exist.
The completion for 'bin/zotonic' was not created.
```

## Expected behavior
Latest versions of Mac OS use zsh, not bash. The script should ideally sense the shell and act accordingly. Failing that there should be some documentation about what to do if one is not using bash shell.

## Config
N/A

mmzeeman:
### Description

Fix #3329

Instead of running the command to install bash completions, which can fail when you are not a bash user and can be mistaken to be a compilation failure. Show a reassuring message containing the version number when the build is done.

```
$ make
/rebar3  compile
===> Verifying dependencies...
...
...
===> Compiling zotonic_mod_admin_config
===> Compiling zotonic_mod_custom_redirect
===> Compiling zotonic_mod_ssl_letsencrypt
Zotonic 1.0.0-rc.15 [Erlang/OTP 26.0-rc1] was successfully compiled
```

**Note** The bash completions are not installed automatically now.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_admin/src/models/m_admin_status.erl
+++ b/apps/zotonic_mod_admin/src/models/m_admin_status.erl
@@ -30,6 +30,7 @@
     m_get/3,
 
     otp_version/0,
+    zotonic_version/0,
     database_version/1,
     tcp_connection_count/0,
     group_sockets/0,
@@ -39,8 +40,11 @@
 ]).
 
 -spec m_get( list(), zotonic_model:opt_msg(), z:context()) -> zotonic_model:return().
-m_get([ <<"zotonic_version">> | Rest ], _Msg, _Context) ->
-    {ok, {z_convert:to_binary(?ZOTONIC_VERSION), Rest}};
+m_get([ <<"zotonic_version">> | Rest ], _Msg, Context) ->
+    case z_acl:is_admin(Context) of
+        true -> {ok, {zotonic_version(), Rest}};
+        false -> {error, eacces}
+    end;
 m_get(Path, Msg, Context) ->
     case z_acl:is_admin(Context) of
         true -> m_get_1(Path, Msg, Context);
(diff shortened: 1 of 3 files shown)


apps/zotonic_mod_admin/src/models/m_admin_status.erl
```diff
@@ -39,8 +40,11 @@
 ]).
 
 -spec m_get( list(), zotonic_model:opt_msg(), z:context()) -> zotonic_model:return().
-m_get([ <<"zotonic_version">> | Rest ], _Msg, _Context) ->
-    {ok, {z_convert:to_binary(?ZOTONIC_VERSION), Rest}};
+m_get([ <<"zotonic_version">> | Rest ], _Msg, Context) ->
+    case z_acl:is_admin(Context) of
```
mworrell, line 44: Maybe also let admin users see this? So `orelse z_acl:is_allowed(use, mod_admin, Context)`

mworrell:
Is the bash completion still installed (if bash is available). Or is it now just a command that you can run yourself?

mmzeeman:
I removed it because it fails when you are not using bash, it returns a confusing message when that happens, which looks like the compilation failed.

apps/zotonic_mod_admin/src/models/m_admin_status.erl
mmzeeman: It did not have a check yet. It looked like it needed one, because this is accessible via the external api, so I added the same check as is done with the db and otp version checks. We could address that in a different issue/pr I think?

Issue #3305, opened 2023-02-22
<select> with attribute multiple does not return an array in validation and submit


williamthome:
For `<select>` HTML elements, LiveValidation only returns the first option selected even if the attribute multiple is set (`<select multiple></select>`). See in `livevalidation,js`, line 385:
```javascript
case LiveValidation.SELECT:
            if (this.element.selectedIndex >= 0) return this.element.options[this.element.selectedIndex].value;
            else return "";
```

I think it needs to be:
```javascript
case LiveValidation.SELECT:
            if (this.element.selectedIndex >= 0) return $(this.element).val(); // jQuery returns an array when multiple is set
            else return this.element.getAttribute("multiple") ? [] : "";
```

Also, same problem with the submit.
In `zotonic-wired.js`, line 1854, the check is weird for me:
```javascript
var one = (t == 'select-one');
```
I think it needs to check for falsy value for the `el.getAttribute("multiple")`.

**Offtopic but related**

- Validation and Submit do not always produce/check the same value (checkbox input is an example);
- Checkbox values are strange to me, it returns array or undefined. I think it should be an array only if more elements with the same name exist in the form and should return an empty array, not undefined.

I can do a PR to continue this discussion with some code, but I first want to hear what you think about this @mworrell.

mworrell:
Good one.

How should we use validation on a multi-value select element?

For the post it is simpler, just multiple times the key/value combination. At least I think that is the correct behavior, bit like a multi-checkbox.

I agree with the array/undefined.  That is weird and very unexpected.

This whole LiveValidation might need an update, but that is for another time...

PR #3336, opened 2023-03-18
Add input types to livevalidation

williamthome:
### Description

Livevalidation it's an old lib and not all `input` types existed back then.
This PR adds the following types:
- color
- date
- datetime-local
- month
- week
- time
- range
- search

> **Note**
> The change is from line 285 to 300, the rest are just trimmed whitespaces.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [X] no BC breaks

### DIFF
--- a/apps/zotonic_mod_wires/priv/lib/js/modules/livevalidation-1.3.js
+++ b/apps/zotonic_mod_wires/priv/lib/js/modules/livevalidation-1.3.js
@@ -282,6 +282,22 @@ LiveValidation.prototype = {
             return LiveValidation.TEXT;
         if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'HIDDEN')
             return LiveValidation.TEXT;
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'COLOR')
+            return LiveValidation.TEXT;
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'DATE')
+            return LiveValidation.TEXT;
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'DATETIME-LOCAL')
+            return LiveValidation.TEXT;
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'MONTH')
+            return LiveValidation.TEXT;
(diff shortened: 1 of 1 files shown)


apps/zotonic_mod_wires/priv/lib/js/modules/livevalidation-1.3.js
```diff
@@ -282,6 +282,22 @@ LiveValidation.prototype = {
...
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'MONTH')
+            return LiveValidation.TEXT;
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'WEEK')
+            return LiveValidation.TEXT;
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'TIME')
+            return LiveValidation.TEXT;
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'RANGE')
+            return LiveValidation.TEXT;
+        if (nodeName == 'INPUT' && this.element.type.toUpperCase() == 'SEARCH')
+            return LiveValidation.TEXT;
```
vkatsuba, line 300: I suppose that this can be simplify like:
1 - Define a constant eg.: 
```js
const x = ['COLOR', 'DATE', 'DATETIME-LOCAL', ''WEEK', ...']
```
2 - Use `includes()` method for detect if item are exist: eg:
```js
        if (nodeName == 'INPUT' && x.includes(this.element.type))
          ....
```

williamthome: I agree with you, @vkatsuba. This code can be improved, but IMHO, this lib needs to be updated, reviewed, or changed by another one, like [Zod](https://zod.dev/). The majority of the code is still the same as the [original one](https://github.com/alechill/livevalidation) (the last commit date is Feb 13, 2010). I just kept the code semantic.
See [this comment](https://github.com/zotonic/zotonic/issues/3305#issuecomment-1443368354) from @mworrell.

vkatsuba: Yep. It was just a proposal.

mmzeeman:
Thanks @williamthome, and indeed, the validation code must be revised to include "modern" input types.

PR #3342, opened 2023-03-21
Add standard manifest json

### LINKED ISSUE #3339: mod_base does not have a manifest.json template/dispatch
## `Zotonic` version

1.0.0-rc.15

## Description
* **Motivation**

Cotonic likes to have a service worker installed. We usually do that from the template of the site. It would be nice though to have a base manifest file, which can be overridden, to keep things uniform, and serve as an example.

* **Proposal**

Add a manifest template file to `mod_base`

## Current behavior

No manifest template provided.

## Expected behavior

A manifest file/template is provided, so a service worker is started by default.

## Config

What kind of things must be configured?

mmzeeman:
### Description

Fix #3339 

- Added a dispatch rule `manifest_json` to have a (sort of) standardized way to add the manifest to a site.
- Added a maskable icon (which can be overridden)
- Added the manifest to the base template and the base template of the admin.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman:
Oh, I also removed the IE6-9 shim html comment hack from the base template.

mmzeeman:
@mworrell Do you know what I must do? It is looking for an rst file for the newly added, but sphinx does not tell what file it is looking for.

PR #3346, opened 2023-03-22
[mod_zotonic_admin_frontend] remove duplicate wire

DorienD:


apps/zotonic_mod_admin_frontend/priv/templates/_admin_frontend_edit.tpl
```diff
@@ -20,10 +20,9 @@
 	{% with id.is_a|default:(m.category[cat].is_a) as cats %}
 	{% wire id="rscform"
 			type="submit"
-			postback={rscform view_location=view_location cat=cat id=id}
+			postback={rscform view_location=m.acl.user.id cat=cat id=id}
```
mworrell, line 23: Shouldn't that be:  `view_location|default:id` ?   You are editing the id.

Some templates pass special view locations where the edited content can be seen, so we need to keep this functionality.

mworrell (changes requested):

apps/zotonic_mod_admin_frontend/priv/templates/_admin_frontend_edit.tpl
DorienD: Oof! you are super correct. Changed it.

mworrell:
Handled in different merge request.

PR #3347, opened 2023-03-23
New style for the _debug.tpl

williamthome:
### Description

This PR implements a new style for the `_debug.tpl` used by the `debug` tag in templates.
The motivation is that the current style it's a bit difficult to read and understand, especially if a big object is printed.

This is how it looks today:
z-old-debug-style

And that's what this PR implements:
z-new-debug-style

The only JS code is from [PrismJS](https://github.com/PrismJS/prism), the syntax highlighter.
PrismJS license is MIT and its lib is only loaded if the `m.site.environment` ~~is `development`~~ isn't `production` (ran into a doubt here, what's the best, is development or isn't production, or neither?).

The only thing that looks weird to me is the top right buttons. It can look better using icons, but I have just used HTML symbols.
The buttons are:
- &lowbar;: Collapse all nodes
- &square;: Expand all nodes
- &times;: Removes the debug element

There is also a resizer in the bottom right corner to resize the element horizontally.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Looking really nice.

I was thinking about other uses for prism too. :-)

williamthome:
Thank you!

PR #3348, opened 2023-03-24
Allow pick debug vars in debug scomp

williamthome:
### Description

This PR allows picking template variables to debug.
Optionally, variable names can be provided to be debugged, if none, all will be, like nowadays. 
For example, picking foo and bar:
```django
{% debug foo bar %}
```
image

Picking none:
```django
{% debug %}
```
image

I was thinking about using the `print` tag, but it is built-in `template_compiler` and will break compatibility if changed to use the runtime adapter instead of the current internal function.
Another possibility was to create a new scomp, but in the end, it will be the same as the debug scomp.

The motivation for this is that I miss the option to only debug what I need to debug.

I also updated the docs, but I was having some trouble generating it in my machine to visualize how it looks and if the format and images are ok.
_If there is something wrong in the docs, please let me know, will be happy to fix or improve it. English is not my best._

### Checklist

- [X] documentation updated
- [ ] tests added
- [X] no BC breaks


mworrell:
Great addition!

PR #3359, opened 2023-03-29
Fix deny rules for specific collaboration groups (0.x)

cdfa:
### Description

The documentation gives [an example](https://zotonic.com/docs/1752/mod_acl_user_groups) for deny rules where permissions on a specific collaboration group are denied with a deny ACL rule, but this does not actually work.
The reason is that the function for checking if there was an ACL rule that grants permissions for all collaboration groups did not check if one was excluded by a deny rule.

Deny rules for collaboration groups still work a bit differently from deny rules on specific categories, though.
Deny rules on specific categories work for all child user groups, while deny rules on specific collaboration groups only work if the rule that grants access is in the same user group.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [X] no BC breaks (unless someone depends on faulty behavior)

### DIFF
--- a/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
+++ b/modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
@@ -585,7 +585,7 @@ can_rsc_1(Id, Action, CGId, CatId, UGs, Context) ->
 can_rsc_for_collab(Id, Action, CGId, CatId, UGs, Context) ->
     m_rsc:is_a(CGId, acl_collaboration_group, Context)
     andalso (
-        can_rsc_for_all_collab(Id, Action, CatId, UGs, Context)
+        can_rsc_for_collab_ugs(Id, Action, CGId, CatId, UGs, Context)
         orelse can_rsc_collab_group_content(Action, CGId, UGs, Context)
     ).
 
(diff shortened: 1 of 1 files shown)


modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
```diff
@@ -639,12 +654,13 @@ is_collab_group_member_action_allowed(_CGId, _Action, _Context) ->
 % the action on all collaboration groups.
 can_rsc_for_collab_ugs(Id, Action, CollabId, CatId, UGs, Context) ->
     AllCollabId = m_rsc:rid(acl_collaboration_group, Context),
+    ?DEBUG(Id),
```
mworrell, line 657: Can go away?

modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
```diff
@@ -587,8 +587,23 @@ can_rsc_for_collab(Id, Action, CGId, CatId, UGs, Context) ->
...
 
+has_unrestricted_collab_content_user_groups(Context) ->
+    AllUGs = sets:from_list(user_groups_all(Context)),
+    UnrestrictedUGs = sets:from_list(
+        lists:map(fun (UG) -> m_rsc:rid(UG, Context) end,
+        unrestricted_collab_content_user_groups())
+    ),
+    not sets:is_disjoint(AllUGs, UnrestrictedUGs).
+
+unrestricted_collab_content_user_groups() ->
```
mworrell, line 601: Maybe add a TODO here?

I could also imagine that we could make those user groups or users "unrestricted" if they have permission to delete or insert random collaboration groups.

So in this case we could check if the user has permission to insert a (not insert-own, but really insert) a collaboration group into the Content-Group of the Collaboration-Group we are checking.

modules/mod_acl_user_groups/support/acl_user_groups_checks.erl
```diff
@@ -639,12 +654,13 @@ is_collab_group_member_action_allowed(_CGId, _Action, _Context) ->
 % the action on all collaboration groups.
 can_rsc_for_collab_ugs(Id, Action, CollabId, CatId, UGs, Context) ->
     AllCollabId = m_rsc:rid(acl_collaboration_group, Context),
+    ?DEBUG(Id),
     lists:any(
         fun (GId) ->
-            case {
+            case ?DEBUG( {
```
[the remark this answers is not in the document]
cdfa: Woops, didn't check my last commit

PR #3370, opened 2023-04-05
Fix automatic recompiling in nested src directory

cdfa:
### Description

I had zotonic cloned under `cdfa/src/zotonic-1.x/zotonic`, which broke automatic recompilation, because regex always gives the longest match (`cdfa`, `zotonic-1.x/zotonic/apps/zotonic_filehandler/src/zotonic_filehandler_mappers`.

I tried to come up with a solution that just starts looking for `src` from the end of the path, but in the end I think recursively calling the test function is the most compact. 

### Checklist

- [ ] documentation updated
- [ ] tests added
- [X] no BC breaks


mworrell:
Aha, a `rsc` in a `src` directory.  That might indeed confuse some patterns.

PR #3373, opened 2023-04-06
core: Trigger a poll when  task_done is received

### LINKED ISSUE #3372: Faster purging of due tasks on the task queue.
## `Zotonic` version
Zotonic 1.0.0-rc.15 [Erlang/OTP 26.0-rc1]

## `OS` version
MacOS

## Description
* **Motivation**

When you enter multiple due tasks in the task queue, they should be handled fairly quickly. 

* **Proposal**

After receiving `task_done` let the queue check if there are other due tasks. And start them as quickly as possible.

## Current behavior

The due tasks stay in the queue, waiting for the next poll. Usually about 20 seconds.

## Expected behavior

The tasks are handled more quickly.

mmzeeman:
### Description

Fix #3372

When the task queue process receives a task queue, it triggers another poll. By doing this other due tasks are run more quickly.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_pivot_rsc.erl
+++ b/apps/zotonic_core/src/support/z_pivot_rsc.erl
@@ -402,6 +402,7 @@ handle_call({task_done, _TaskPid, TaskId}, _From, #state{ task_id = TaskId } = S
         task_id = undefined,
         task_pid = undefined
     },
+    self() ! poll, % Request another poll, there could be more due tasks on the queue.
     {reply, ok, State1};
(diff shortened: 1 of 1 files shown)


mworrell:
@mmzeeman I reworked it a bit to poll if you insert an immediate task (`undefined` date) and also cancel the last poll timer.

mmzeeman:
Ah, good one. Usually I trigger a poll after doing that. That is no longer needed. 👍

PR #3387, opened 2023-04-21
core: always set the is_protected flag for the admin user

mworrell:
### Description

Add an extra protection against deletion of the admin user.

The delete routine already refuses to delete `1`, setting the `is_protected` flag protects against any other accidental deletion because of triggers or erroneous queries.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (commented):
Good idea... 😀

PR #3395, opened 2023-05-01
Inserting and deleting edges from the edge model

mmzeeman:
### Description

Adds the possibility of adding edges from the edge model api. This makes it very handy to
use in live templates. 

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell (approved):

mmzeeman:
I saw your approach when I was finished with this PR. My thoughts were to follow the erlang api more closely, so I opted for the `post/insert` topic. The elements in the path next to it are optional with a question mark, then it `subject`, `predicate` or `object` is either taken from a data attribute or the context.

mmzeeman:
Then it can be used like this (in a live template or teleview):

```django
<div class="d-flex align-items-center bg-light border-0 rounded-pill py-2 px-3 mt-2"
     data-onclick-topic="bridge/origin/model/edge/post/insert/{{ id }}/?/{{ m.acl_profile.effective }}" >
    <button data-edge-predicate="is_interested"
            class="btn btn-warning rounded-pill d-flex align-items-center me-3">
        <i class="icon-circle-question-regular me-2"></i>{_ I Might Go _}
    </button>
    <button data-edge-predicate="is_going"
            class="btn btn-success rounded-pill d-flex align-items-center">
        <i class="icon-circle-check-regular me-2"></i>{_ I'm Going _}
    </button>
</div>
```

mmzeeman:
The `o`, `s` approach also looks logical btw... Maybe we should merge the two approaches.

mworrell:
I am ok with both.  Using the `o` and `s` make it a bit more similar with the get and the rsc model.

PR #3407, opened 2023-05-09
core/seo: redirect pages to the current language version

mworrell:
### Description

These are SEO optimizations that should help Googlebot make more sense of our multi-lingual resources.

Short description: if a page without language is requested then redirect to the path which includes the selected language code.

For example this will redirect `/` to `/en/` (assuming `en` is the default language and there is no `z.lang` cookie set that forces another language).

The algorithm used is:

```
IF and only if we have negotiated to show a HTML page.
AND there is NO language set in request URL
AND we have multiple languages enabled
AND the config to set the language in the URL is set
THEN
	IF there is a language cookie THEN
		Take its value as the selected language
	ELSE IF the force-default-language config is set THEN
		Take the config value as the selected language
	ELSE IF there is a accept-language header THEN
		Negotiate the language for the selected language
		Set the Vary response header to ‘accept-language’
	ELSE
		Take the default language as the selected language
	ENDIF
	Determine the new URL with the selected language
	Set the Location response header to this URL
	Set the response code to 303 (See Other)
ENDIF
```

Also:

- if force-default language is set then set the x-default hreflang variant in the page headers to the default language otherwise use the language-less url for the x-default tags
- always add a hreflang for the current default language, irrespective if that language is defined for the resource
- set the noindex for a page if the current language is not in the resource's languages AND the current language is not the default language AND the current page is not a collection (or query)
- generate hreflang tags for all languages if the current page doesn't have a resource id, or the resource is a collection (or query).

See also the Google documentation:

https://developers.google.com/search/docs/specialty/international/localized-versions

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Amazing what google does with the web these days. Doing content negotiation is too subtle I guess. What I read here: [managing-multi-regional-sites](https://developers.google.com/search/docs/specialty/international/managing-multi-regional-sites) is that when you have a multi regional site you can even better go with a generic tld. That also explains why our .org site gets a much higher in the google results than the .nl site.  (Recently switched)

mworrell:
Adding some extra logic, regarding the language selections.

From my notes:

```
Two different things:

- UI language, as selected by the user
- Content language

Language in Context should be a list of languages, in preference order.
If a new language is selected then it is moved to the front of the preference order.
The default preference order is defined by dragging & dropping the selected languages in the correct order. This is done on the admin/languages page.  The top language is the default language, the preference-order is stored as a list with the default language config.

On a request we:

1. Fetch the accept-language header, extract a preference order from that list, add base languages, append fallback languages from the language definitions, filter on available languages.
2. Append the list of preferred languages, remove duplicates.
3. Store this preference list in the z.lang cookie and the context language.

On display of a trans record:
- Select first language matching the preference order
- If no match, then try ‘en’
- If no match then take first language

On display of a resource with the page controller:
- If the page is a collection or the home page then:
    - Select the z_language as the z_content_language
- Else
    - Select best matching language from the ‘language’ property
- Use this language as the z_content_language, pass as the language variable

In the SEO templates:
- If no z_content_language then:
    - Show all config language a hreflang links
    - Use the z_language for SEO links
- Else
    - Show hreflang links for the resource languages
    - Use the z_content_language for the SEO links (as it is the language the main content is displayed in.)
- End

As the other controllers don’t pass the z_content_language those templates will use z_language as the assumed content language.
```

/cc @mmzeeman

PR #3414, opened 2023-05-11
core: z_install_update checking index before renaming

heiheshang:
Tried to migrate from version 0.34.0 to master , couldn't start the site due to missing fki_comment_persistent_id index. Added a check for the existence of an index

### Description

When trying to rename the fki_comment_persistent_id index in the comment table, added an index check.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Thanks!!!

PR #3420, opened 2023-05-18
WIP: Use installed or downloaded rebar3 version

mworrell:
### Description

Make it possible to use the locally installed rebar3 version.

 - [ ] Check if we want to update the test matrix like in https://github.com/epgsql/epgsql/blob/devel/.github/workflows/ci.yml

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Closing for now

PR #3443, opened 2023-06-12
mod_search: add 'hasanysubject' search term

mworrell:
### Description

Add `hasanysubject` search term, which is the same as `hasanyobject` but then for subjects.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


DorienD (approved):
Works like a charm

PR #3444, opened 2023-06-12
Clear ratelimit history for username after successful logon. 

### LINKED ISSUE #3442: mod_ratelimit limits logging on after 5 correct logons from MQTT client
## `Zotonic` version
Zotonic 1.0.0-rc.15 [Erlang/OTP 26.0]

## `OS` / `Browser` version
MacOS / Safari

## Steps to reproduce

Setup an mqtt client to connect via 1883 or 8883 with username password.

Screenshot 2023-06-12 at 09 04 39

Connect and disconnect a couple of times. After that you get the warning message the logon rate limit is enabled for this user.

Screenshot 2023-06-12 at 09 21 03

mworrell:
### Description

This fixes an issue where repeated, successful logons could trigger a rate limit restriction.

This is fixed by deleting the ratelimit history for the username (and device) after a successful logon.

Also let MQTT use the `z_auth:logon/2` instead of `z_acl:logon/2` so that the correct events are triggered and the user is checked for being published.

Fixes #3442

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_mqtt_sessions_runtime.erl
+++ b/apps/zotonic_core/src/support/z_mqtt_sessions_runtime.erl
@@ -227,17 +227,42 @@ connect(#{ type := connect, username := U, password := P, properties := Props },
                         undefined -> Context1;
                         Sid when Sid =/= <<>> -> z_context:set_session_id(Sid, Context1)
                     end,
-                    Context3 = z_acl:logon(UserId, Context2),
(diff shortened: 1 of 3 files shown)


mmzeeman (approved):
Works perfectly. Limits logons when needed, and does not trigger when not needed.

PR #3449, opened 2023-06-16
MQTT - Login format check

heiheshang:
### Description

this is for those clients who use the connection as user@domain

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mmzeeman:
Thanks @heiheshang,

In zotonic usernames are allowed to be email addresses. What happens when you try to login with `hostname:user@example.com` as username? 

This was possible with the previous implementation, but I think it will cause a crash with this implementation.

heiheshang:
this will be a problem, we need to make changes.

mmzeeman (commented):
Thanks @heiheshang. It looks like erlpop was accidentally added to the rebar.lock file.

heiheshang:
> 

not entirely by chance, I use it in the module, it was included in the general commit, probably the lock file should be included in the exceptions?

mworrell:
Is this still needed?

If so, then this is a simpler split, but note that this prohibits a logon using only an email address as username. We might want to put the split on `@` behind a configuration.

```erlang
split_vhost_username(Username) ->
    split_vhost_username(Username, Username).

split_vhost_username(<<>>, Username) ->
    % Username without designation
    [Username];
split_vhost_username(<<":", _/binary>>, Username) ->
    % "example.com:localuser"  (Zotonic 1.x format)
    binary:split(Username, <<":">>).
split_vhost_username(<<"@", _/binary>>, Username) ->
    % "localuser@examplesite"  (Zotonic 0.x format)
    [Username, VHost] = binary:split(Username, <<"@">>),
    [VHost, Username];
split_vhost_username(<<_/utf8, R/binary>>, Username) ->
    split_vhost_username(R, Username).
```

heiheshang:
> 

yes i use this in my project. In parallel, I did it in https://github.com/heiheshang/zotonic/blob/8722f41892d95f84a0c0490a9c52a357bc937f81/apps/zotonic_core/src/support/z_mqtt_sessions_runtime.erl#L462

mworrell:
See #3486 for a slightly cleaned up version, with an extra config key to prevent problem with people using email addresses as username and a 1.x logon format.

mworrell:
@heiheshang Can you check #3486 - if that work for you?  You will need to set a zotonic.config variable `{mqtt_username_format_0x, true}`

mworrell:
For now closing, as I think #3486 will resolve this issue.

PR #3458, opened 2023-06-27
core: add security.txt support

mworrell:
### Description

Add the `/.well-known/security.txt` file to the core.

See also https://securitytxt.org

The values are looked up using the following configuration order:

1. Site configs:

 * `site.security_email`
 * `site.security_url`
 * `site.security_policy_url`
 * `site.security_hiring_url`
 * `site.security_expires`

2. Special unique page names, to optionally add information to the site:

 * `page_security_contact`
 * `page_security_policy`
 * `page_security_hiring`

3. Global configs:

 * `security_email`
 * `security_url`
 * `security_policy_url`
 * `security_hiring_urll`

4. Email property of the admin, used for security_email (if visible by anonymous).

If the `site.security_expires` config is not set then the default expires is a month in the future.

A site specific `security.txt` can be defined by adding the template `security.txt.tpl` to the site.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_base/templates/security.txt.tpl
+++ b/modules/mod_base/templates/security.txt.tpl
@@ -1,0 +1,24 @@
+{% with m.site.security as security %}
+# Our security address
+{% for _, c in security.contact %}
+Contact: {{ c }}
+{% empty %}
+# Please contact the site administrator as no contact details are configured for security issues specifically.
+{% endfor %}
+
+{% if security.policy as url %}
+# Our security policy
+Policy: {{ url }}
+{% endif %}
+
+{% if security.hiring as url %}
+# We are hiring!
+Hiring: {{ url }}
+{% endif %}
+
+# The official location of this file
+Canonical: {% url security_txt use_absolute_url %}
+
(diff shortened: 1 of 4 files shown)


modules/mod_base/templates/security.txt.tpl
```diff
@@ -0,0 +1,24 @@
+{% with m.site.security as security %}
+# Our security address
+{% for k, c in security.contact %}
+Contact: {% if k == 'email' %}mailto:{% endif %}{{ c }}
+{% empty %}
+# Please contact the site administrator as no contact details are configured for security issues specifically.
+{% endfor %}
```
robvandenbogaard, line 7: Should it be possible to have multiple contact addresses configured?
Is that why it has the for..in..?

mworrell: Yes, the model supports an email and url.

robvandenbogaard (approved):

PR #3459, opened 2023-06-28
mod_authentication: prevent admin to be able to log in using an username_pw identity

mworrell:
### Description

This prevents the admin user to log in using an username_pw identity.
Only the configured admin password and username 'admin' can be used to logon the admin using an username/pw combo.

Also hide the UI to set an username_pw for the admin.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Good call..

Maybe a comment indicating how it is possible to set/update the admin at the location where `{error, admin_password_cannot_be_set}` is returned.

PR #3502, opened 2023-08-14
mod_zotonic_wires: Add focus when opening modal, close on esc when modal is open

DorienD:
### Description

This will add a focus inside the modal once it's been opened. It will use the close button since this is in most cases the first clickable element.

This wil also let users close the modal with the escape key.


apps/zotonic_mod_base/priv/lib/js/modules/z.dialog.js
```diff
@@ -166,8 +166,19 @@
             if (typeof($.widgetManager) != 'undefined') {
                 $dialog.widgetManager();
             }
+
             z_editor_add($dialog);
 
+            setTimeout(function() {
+                $('a.close').focus();
```
mworrell, line 173: Should we check if there is an input element with focus?

DorienD: Good one! Yes

DorienD: Fixed!

mworrell (approved):

PR #3532, opened 2023-09-12
mod_mailinglist: only export rsc emails if published

mworrell:
### Description

Only export the email addresses of subscribers that are published.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Closed, this filtering is already done in another routine.

PR #3546, opened 2023-09-25
mod_backup: periodic cleanup of revisions

mworrell:
### Description

Add a simple periodic cleanup of the revision log.

Revisions are deleted after 18 months.
This can be changed using the config `mod_backup.revision_retention_period`

For the cleanup of person data (GDPR) we will need a shorter period.
This will be done in a separate pull request.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_backup/src/models/m_backup_revision.erl
+++ b/apps/zotonic_mod_backup/src/models/m_backup_revision.erl
@@ -30,12 +30,19 @@
     get_revision/2,
     list_revisions/2,
     list_revisions_assoc/2,
+
+    periodic_cleanup/1,
+
     install/1
 ]).
 
 -include_lib("zotonic_core/include/zotonic.hrl").
 
 -define(BACKUP_TYPE_PROPS, $P).
+
+% Number of months we keep revisions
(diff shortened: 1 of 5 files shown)


robvandenbogaard:
Maybe good to have an indication in the mod_backup admin page for the currently set (or default) retention period.

apps/zotonic_mod_backup/src/models/m_backup_revision.erl
```diff
@@ -30,13 +30,20 @@
...
+
     install/1
 ]).
 
 -include_lib("zotonic_core/include/zotonic.hrl").
 
 -define(BACKUP_TYPE_PROPS, $P).
 
+% Number of months we keep revisions
+-define(BACKUP_REVISION_RETENTION_PERIOD, 18).
```
robvandenbogaard, line 44: Maybe too long for the name, but if we add the unit, it is clear from seeing the config value what is to be expected of the value ;)
`BACKUP_REVISION_RETENTION_(IN_)MONTHS`

robvandenbogaard (changes requested):
Config name and display on admin page.

robvandenbogaard (approved):

PR #3553, opened 2023-09-29
mod_development: add a dependency graph of all templates

mworrell:
### Description

Show a dependency graph of the templates in the admin/development.

Todo:

 - [x] includes
 - [x] coloring of node per module
 - [x] extends
 - [x] overrules
 - [x] include all
 - [x] catinclude

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


robvandenbogaard:
Nice to have dot/viz onboard, and I can imagine that this will help with some of the more complex template dependency issues.

mworrell:
@robvandenbogaard yes, what we now could add is:

 - Extra includes by
   - parsing block-includes
   - expressions like `v|default:"x.tpl"` 
- Filtering the graph on certain modules or templates, idea I have is to start with a set of templates from the selected modules (or template names) and then determine all includes/extended templates "connected" to that set.

DorienD:
with your third point, do you mean replacing the textual trace when you look at the source code? Because i quite like that :)

I would also love the filtering options, some suggestions: 

1. Clicking on a template in the graph and only see the connections to that specific template. 
2. Choose between modules to show the graph, so we can see the "site" template connections
3. Are the colors now used to show if a template is an include, overrules or extends? Or are the colors random? 

Love this btw!

mworrell:
With the third I mean the option that we had in the 0.x, where you could see a real-time trace of all the templates being compiled and/or included. This was in a separate mod_development page.

I want to keep the textual trace - that is really handy.

This would be some feature where we would on runtime all inclusions and extends being done. And as such building a graph of all templates actually being used.  This graph you could then leave open on a page, and have an option to clear it. In this way you can open a couple of pages and then see which templates are included/extended for rendering the pages you look at.

This could even be done on a live system, but then we need to restrict to the templates being rendered for the current user/session.  Which could also be an option.

Your questions:

1. Would be nice - for that we need some code to check the node clicked and then see if we can highlight the connections (should be doable with some good labeling of edges).
2. Yes, maybe a multi-select box.  Though I would prefer to "start" with those modules and then also show what is included from other modules.
3. The colors are selected by hashing the module name, so they are stable.

DorienD:
I would love the runtime of a page with a graph of the included templates on that page! That would be super helpfull. 

Would there also be a way of showing that an included template does not exist in the graph? You've made a textual overview of it but maybe it would be cool to also see it in the graph? Sort of red faded or something?

PR #3577, opened 2023-10-13
mod_acl_user_groups: add option to target the exact category in ACL rules

mworrell:
### Description

In some scenarios you want to allow an user to have access to a certain category but not too its sub-categories.

Previously a _deny_ rule was needed for those sub-categories.
Now there is an option "Exact category", which can be selected when adding ACL rules.

> Screenshot 2023-10-13 at 16 22 18

Also add the option `author_is_owner` to consider a page's authors as the owner of that page. This ensures that ACL rules marked as "manage own" are applicable to authors.

> Screenshot 2023-10-13 at 16 19 10

Also allow the module manager to continue in case of module downgrades.  Let the module be responsible for any backwards-compatibility problems.  (This was needed due to the upgrade of the schema-version in mod_acl_user_groups, and subsequent testing with different sites/versions.)

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


robvandenbogaard:
Sorry for the late review; good functionality, had some remarks about naming (author_is_owner 👍 vs is_category_exact ~ while I see it's functionality, usage readability is impacted by forcing prefix notation), but that is more a general thing; so, great that it was merged! ;)

PR #3581, opened 2023-10-26
mod_survey: Add get_stored to return a list of survey ids a user has answered

mmzeeman:
### Description

Adds a model get call to retrieve the survey ids user has (partially) answered.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_survey/src/models/m_survey.erl
+++ b/apps/zotonic_mod_survey/src/models/m_survey.erl
@@ -51,6 +51,7 @@
     delete_result/4,
     delete_results/2,
     get_questions/2,
+    get_stored/2,
 
     rsc_merge/3,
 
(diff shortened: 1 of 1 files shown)


apps/zotonic_mod_survey/src/models/m_survey.erl
```diff
@@ -205,6 +206,10 @@ m_get([ <<"result_columns">>, SurveyId, Format | Rest ], _Msg, Context) ->
         {error, _} = Error ->
             Error
     end;
+m_get([ <<"get_stored">>, UserId | Rest ], _Msg, Context) ->
+    {ok, {get_stored(m_rsc:rid(UserId, Context), Context), Rest}};
```
mworrell, line 210: We will need some access control here.

Either it is the user themselves (if there is an user) or someone with the correct rights on either the module or the returned surveys.

mworrell (changes requested):

mmzeeman:
Indeed.. I'll remove it from this model and close this PR.

We have different ACL for dealing with the filled out answers here. The ACL for accessing the survey answers is hard-coded right now. It can only be done when you are allowed to edit the survey, or when it is your own survey. We want members to be able to view the answers.

Issue #3590, opened 2023-11-08
Prevent secrets from being displayed


CyBeRoni:
## `Zotonic` version
0.x and 1.0

## `OS` version
Any

## Description
* **Motivation**
Prevent secrets like mod_filestore secret key and such from being extracted after being set, such that people who _do_ have access to Zotonic's admin UI but should _not_ have access to such credentials cannot get them. 

I am aware that the credentials would still be in the SQL backup as made by mod_backup and could still be accessed from the shell, but I can prevent access to these independently.

* **Proposal**
For mod_filestore: the API Secret field at /admin/filestore should not display its current value, rather it should only be able to be set by inputting a *new* value.

For the /admin/config screen, any field containing secrets should be masked; either entirely or by only displaying the first (and possibly last) few characters to allow an administrator to verify the value is what they expect it to be.

Perhaps a boolean value could be added for the config fields indicating if a field should be masked or not. (Of course this field should not be changeable from the admin ui.) This field can be initialised to true if the field name contains something like 'key', 'secret' or '_password'.

mworrell:
A partial solution is added in #3593 

That doesn't add any security to leaking secrets to people with access to the admin. We might want to add something that the user's password must have been recently entered, and if not force a password entry before editing a config value.

mworrell:
#3593 is merged.

Proposal to do next is a generic password setter dialog for secrets. This should have:

 * A check for recent auth of the user
 * Only allow secrets to be fetched via API if the user has recently authenticated (otherwise show password dialog - with 2FA)
 * No secrets in the DOM of the client
 * Generic action in mod_admin_config - for easy re-use in other interfaces
 * Optional test topic for checking if the newly entered username/pw is working (for APIs etc) before it is saved
 * Option: never let the API fetch secret configs - only the first/last characters
 * Option: use the encrypted transport of the key server - this prevents listeners on topics to see the secrets

PR #3593, opened 2023-11-13
mod_admin_config: hide secrets from direct view

mworrell:
### Description

Issue #3590 

Not a complete solution yet to #3590, this pull request:

 * Only show first three chars in /admin/config for values considered a secret
 * The `is_secret` is for now set by checking the config key name for `secret`, `key` or `password`
 * Some exemptions to this name matching are hard coded (`s3key`, `password_min_length`)
 * The filestore password is shown as a password input, hiding the password from plain sight for "shoulder surfers".

This doesn't add any security to leaking secrets to people with access to the admin. We might want to add something that the user's password must have been recently entered, and if not force a password entry before editing a config value.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_config.erl
+++ b/apps/zotonic_core/src/models/m_config.erl
@@ -90,10 +90,18 @@ all(Context) ->
                 {M, z_utils:index_proplist(key, CMs)}
                 || {M, CMs} <- z_utils:group_proplists(module, Cs)
             ],
-            Indexed1 = [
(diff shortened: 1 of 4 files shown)


mworrell:
Maybe we should set the filestore password/username via a popup, separate from the rest of the settings.

mmzeeman (approved):
👍

PR #3596, opened 2023-11-14
Add a admin development page where you can see all dispatch rules of a site

mmzeeman:
### Description

A page was added where you can see all dispatch rules of a site. This makes it possible to get an overview of what urls are available on the site, and what part of the system is going to handle the request.

Screenshot 2023-11-14 at 16 25 44

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell (approved):

mworrell:
Nice extra page!  Just go ahead and merge when you are ready.

PR #3601, opened 2023-11-20
mod_admin_modules: add button to re-install a module's datamodel

mworrell:
### Description

Add a small button in the module overview to re-install a module's schema and data.

Screenshot 2023-11-20 at 16 40 43

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@mmzeeman @robvandenbogaard I merged this already, as then it could be part of a site update.

mmzeeman:
Yes, thanks... Yesterday I ran into a problem with this. Somehow the data model version number was increased, without doing the actual database upgrade. Probably because I was switching from feature branch to a new master branch.

PR #3608, opened 2023-11-22
mod_search: fix shorthand searches to use new arg format

mworrell:
### Description

The builtin searches used the old map format for their arguments.
This modifies those searches to use the new query term format.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


karlsson:
In mod_search:search/4 line 493 there is a typo "rsc.modfied" instead of "rsc.modified"

mworrell:
Thanks @karlsson, I have made an update. Really good to know that more people are checking the commits!

PR #3612, opened 2023-11-27
mod_authentication: add password confirm on sign in with connecting external account

mworrell:
### Description

This fixes a problem where signing in via an external account that doesn't verify email addresses either a new account could be created or, if (and only if) it was assumed that the email address was verified and 2FA was not enabled then an existing account could be connected without further verification.

A new version of Cotonic is needed, please do run `make update-colonic`

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):

mmzeeman:
Interesting... Our sign in / continue with apple module also delivered unchecked email addresses. The address was not added to the signed data.

PR #3640, opened 2023-12-14
WIP mod_admin: update layout of edit page

mworrell:
### Description

Change layout of edit page.

 - [ ] More accessible connections
 - [ ] Group module panels in less prominent place
 - [ ] Tabs for translation

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
@DorienD Please check, and if you are agreeing then you can merge 😄

DorienD:
When you have the module acl_user_groups the template where you've moved the delete and duplicate buttons gets replaced and thus there is no delete or duplication button anymore.

DorienD:
There is also a bug where the tinymce gets too wide for the screen on initializing it.

DorienD:
On the body content there are still language tabs, i guess those could be removed too right?

mworrell:
> There is also a bug where the tinymce gets too wide for the screen on initializing it.

That is probably not something new - to be picked up in a separate issue.

> On the body content there are still language tabs, i guess those could be removed too right?

Ah yes - we should have only a single language tab bar.

DorienD:
It's something new, the too wide initializing isn't happening on the master branch

mworrell:
Can't reproduce the tinyMCE problem on Safari.  What are you using?

Pushed fixes for the other issues.

DorienD:
On chrome.

DorienD:
Screenshot 2023-12-19 at 12 44 56
Screenshot 2023-12-19 at 12 45 01

mworrell:
Ah, on mobile. I think we need a minmax on the grid.

mworrell:
Can't reproduce, but added minmax to the grid to prevent a grid blowout. Can you check again?

DorienD:
Works!

mworrell:
And merged!

PR #3641, opened 2023-12-14
[core] Self is passed as a binary instead of an atom now

robvandenbogaard:
### Description

In `update_transaction_fun_insert(#rscupd{id = insert_rsc}../5` the special value `self` is expected as an atom, while in `update_transaction_fun_insert(#rscupd{id = Id}../5` it is matched as binary. Using binaries helps with handling of input via API, so to be consistent the match on `self` should be on `<<"self">>`.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

Regarding backward compatibility, I am not sure this won't break anything, though it did break in vivo without the change.


mworrell:
Maybe we should accept both?

robvandenbogaard:
> Maybe we should accept both?

Yes, best for backwards compatibility. `m_rsc_update.erl` seems to be the only file in Zotonic using `<<"self">>`, so if we fix these two cases we should be fine.

mworrell (approved):

PR #3650, opened 2024-01-03
mod_admin: remove some #3L usage in lessc files

mworrell:
### Description

Now it compiles without any warnings about inlined javascript.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


DorienD:
Much better!

PR #3660, opened 2024-01-09
fileuploader: Add start_topic option

williamthome:
### Description

This PR adds a `start_topic` option to the `fileuploader` worker.
The motivation is that I could not get the upload name using the `zotonic.fileuploader.worker.js`, so I could not get the status or cancel the upload.

I also fixed a bug in line 199 of the worker where the `response_topic` came from an inexistent `msg` object.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Thanks!

PR #3661, opened 2024-01-09
fileuploader: Add the ability to cancel uploads

williamthome:
### Description

This PR adds the ability to cancel uploads via fileupload WebWorker.
It depends on #3660, which returns the upload names.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Nice addition!

PR #3664, opened 2024-01-11
fileuploader: Allows to define the upload name

williamthome:
### Description

This PR allows to define the upload name.
I miss this ability because I need to track the file upload before the upload starts.
The multiple `z_ids:id()` is to not break the compatibility.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Good addition, thanks!

PR #3668, opened 2024-01-12
docs: Fix a security problem in jinja before 3.1.3

mworrell:
### Description

See https://github.com/zotonic/zotonic/security/dependabot/1

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Apparently there is a problem in Jinja 3.1

As we don't use key input, I just close this issue and leave the security problem unpatched as it doesn't affect our use.

PR #3680, opened 2024-02-06
mod_oembed: remove the oembed_client process (0.x)

mworrell:
### Description

Remove the oembed_client process. The process was not restarted when it crashes, as the dummy module catches exits. 

As regulation can also be done using a _jobs_ regulator (if needed) this middle man process is not needed. Now we just do a direct oembed lookup for every request in the request context/process.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/modules/mod_oembed/mod_oembed.erl
+++ b/modules/mod_oembed/mod_oembed.erl
@@ -25,7 +25,6 @@
 
 %% interface functions
 -export([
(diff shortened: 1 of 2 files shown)


EdwardAbboud (approved):
Looks good! ready for 0.82 update

PR #3682, opened 2024-02-07
WIP: Upgrade admin to bootstrap5

DorienD:
### Description

Update zotonic_admin from bootstrap 3 to bootstrap 5

This will include a glyphicon to fontawesome refactor.

Add `?bs=3` to the url to see which elements are dropped, i've added fallbacks for them but they should be refactored in the future.

### ToDo

- [ ] Check why `_action_dialog_new_rsc_tab_find_results_item.tpl` and `_action_dialog_connect_tab_find_results_item.tpl` are almost the same but still two templates. And possibly refactor
- [ ] Refactor modals
- [ ] Make sure `data-toggle=tab` `data-toggle=dropdown` `data-toggle=pill` are backwards compatible with the new `data-bs-toggle` version

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Closed, as we have a bs3-to-bs5 pull request here: #4322

PR #3688, opened 2024-02-22
[Mod_menu] Make remove from menu text nicer

DorienD:
### Description

The text was the same as when you delete a resource, with a very scary text THIS CANNOT BE UNDONE. So i made it a little friendlier, since it can be undone because the resource is not deleted.

### DIFF
--- a/apps/zotonic_mod_menu/priv/templates/_menu_edit_scripts.tpl
+++ b/apps/zotonic_mod_menu/priv/templates/_menu_edit_scripts.tpl
@@ -245,7 +245,7 @@ $('#{{ menu_id }}').on('click', '.dropdown-menu a', function(e) {
             z_dialog_confirm({
                 text: "{_ Are you sure you want to delete _}: "
                     + "<b>" + $menu_item.find(".menu-label").html() + "</b><br>"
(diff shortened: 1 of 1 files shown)


mworrell:
Maybe we should make it scary only if the referred thing has:

 - `is_dependent` set; and
 - has only a single incoming edge (ie. the menu resource we are editing)

mworrell:
Hmmm, that is then something that needs to be checked before we can show that confirm dialog.  We will need an API call for that.

mworrell:
Or should we add a property for the rsc model:  `is_dependent_with_single_referrer` ?   Kind of ugly....  though it describes the condition.

mworrell:
My addition is incomplete (and untested), as it doesn't check the items below the deleted items.

mworrell:
@mmzeeman What do you think of this solution.  I am still in doubt about that menu model, the method name does cover what it does, but is kind of ugly...

mmzeeman:
I can take a look, but I have to familiarize myself with the model first. I think we have never used it.

mworrell:
The model is completely new, later we can add other functions to the model.
I am just unsure about the naming of this particular method in the model.

PR #3692, opened 2024-02-26
User creation fails: Add parameters for user and group id in dockerfile.

ptrck0:
Add a user and group id environment variable. These can be used by the entrypoint script to create the user correctly. Without any parameter it will the root id's (0:0) because the directory opt/zotonic is created by the root user.


Dockerfile.dev
```diff
@@ -1,5 +1,8 @@
 FROM erlang:24-alpine
 
+ENV PUID=1000
+ENV PGID=1000
```
mworrell, line 4: Why default to 1000?  Shouldn't this be derived from the current user?

The `stat` construct is to fix access permission errors on the directories and files accessed by the container, when accessing and updating files owned by the user running the container. For this we want to use the same UID and GID in the container as the current user.

ptrck0: The problem is as following:
The directory opt/zotonic is not present on a clean install, so it is created by the default user the base image is using; the root user. 
The docker/docker-entrypoint.sh script then performs a ```stat``` on the opt/zotonic directory, and get 0:0, the root user's UID and GID.
It then tries to create the user zotonic with UID and GID 0, which fails of course.

To reproduce: delete the zotonic volume, and run ```./start-docker.sh```.

Fixing it in the docker-compose file works too, but then you will still not be able to build the image using something like ```docker build -f Dockerfile.dev -t zotonic/zotonic-dev:latest .```. I do not know if that is ok.

mworrell: I am wondering why the directory `opt/zotonic` is not present on clean install.
Isn't it mounted before the install script?

mworrell:
@ptrck0 I have an alternative setup here, using your method of building the docker image in-situ.  https://github.com/zotonic/zotonic/pull/3723

PR #3697, opened 2024-02-28
core: Improve smtp plaintext port determination

mworrell:
### Description

Check various configs to determine the port to use for plain text email fallback.

@seannaswell Can you have a look at this?  I think the logic is ok.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


seannaswell:
Thanks for taking the time to look at this... So, it appears to work.

I picked a random port (28), configured the mail server and Zotonic accordingly, and after the first TLS attempt failed, the second non-TLS attempt on port 28 succeeded.

Many thanks!

mworrell:
Thanks for  checking this!

Also add the `smtp_plaintext_fallback` config option to have a bit more control over this behavior.

PR #3703, opened 2024-03-06
mod_translation: store the language profile data in files

mworrell:
### Description

This fixes a problem where the generated Erlang file containing the profile data could not be compiled if the server didn't have enough memory.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


karlsson:
Compiles fine on 2GB ram now!

PR #3715, opened 2024-03-21
Adapt cloudinit for Ubuntu 22

mworrell:
### Description

Tested with Ubuntu 22

Changes:

 - Add EPMD listen ports to `/etc/environment`
 - Use `kerl` to install erlang
 - Add erlang select using `kerl` to `/etc/profile`
 - Add max inotify file watches config
 - Add `wkhtmltopdf` - which is used by some modules for PDF generation

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/cloud-init/config.d/ports.yml
+++ b/cloud-init/config.d/ports.yml
@@ -8,14 +8,14 @@ zotonic:
     # Outside ports
     port: 80
     ssl_port: 443
-    # HTTP port, iptables maps port 80 to 8000
-    listen_port: 8000
-    # HTTPS port, iptables maps port 443 to 8443
-    ssl_listen_port: 8443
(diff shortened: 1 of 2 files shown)


mmzeeman:
We could simplify the setup a bit by using `setcap` on the `beam` and `beam.smp` processes described here: https://zotonic.com/docs/1789/running-on-port-80-and-port-443. This allows non root beam processes to bind to privileged ports. 

The zotonic config must then also change so it will bind to the right ports.

Setting up firewall rules would not be needed anymore.

mworrell:
We have a `ports.yml` file in the installation so we can just adapt that one.

mmzeeman (approved):
Nice and simple..

PR #3725, opened 2024-03-28
Add information about the epmd daemon to getting-started

ptrck0:
The epmd daemon is not (always) running by default, causing zotonic not be able to open local ports.


mworrell:
This was an error in the IP address defaults for EPMD.

Fixed in #3727

PR #3729, opened 2024-04-04
mod_base: let controller_id accept an extension on the id for content negotation

mworrell:
### Description

This enables URLs like:

 * `https://example.com/id/123.json`
 * `https://example.com/id/123.html`

The extension is mapped to a mime type and replaces the `accept` header used for content negotiation.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_base/src/controllers/controller_id.erl
+++ b/apps/zotonic_mod_base/src/controllers/controller_id.erl
@@ -100,6 +100,8 @@ find_html([ {{<<"text">>, <<"html">>, _}, _} = Prov | _CTs ]) ->
 find_html([ _ | CTs ]) ->
     find_html(CTs).
 
+% Fetch id from the request. Check if the id has an extension, if so split it
+% and set the accept header.
 get_id(Context) ->
     case z_context:get(id, Context) of
         undefined ->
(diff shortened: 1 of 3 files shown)


apps/zotonic_mod_base/src/controllers/controller_id.erl
```diff
@@ -109,9 +111,42 @@ get_id(Context) ->
                 <<>> ->
                     {undefined, Context};
                 Id ->
-                    RscId = m_rsc:rid(Id, Context),
-                    {RscId, z_context:set(id, {ok, RscId}, Context)}
+                    case maybe_split_extension(Id) of
+                        {Root, Ext} ->
+                            RscId = m_rsc:rid(Root, Context),
+                            Context1 = z_cowmachine_middleware:set_accept_context(Ext, Context),
```
robvandenbogaard, line 117: The mapping of extension to mime-type may be one to many, e.g. .json for ld+json, but also for other json based serializations. Can we require a one to one mapping for disambiguation? What would be the common extension for ld+json? .ld.json?

mworrell: I have no idea what the extension for ld+json would be.  But a double extension wouldn't be ok.  So more like `.jsonld`

mworrell: My goodness https://www.file-extension.info/format/jsonld

mworrell: Aha it is `.jsonld` see https://json-ld.github.io/json-ld.org/spec/latest/json-ld/#iana-considerations

mworrell: Added the mapping to the media identify routines.

robvandenbogaard (approved):

PR #3740, opened 2024-04-26
core: pivot all-day dates assuming the timezone of the resource

mworrell:
### Description

This makes some changes to the way "all day" date ranges are pivoted.

The problem was:

 * Date is in CET and all day
 * Pivot is stored in UTC (without date conversion)
 * Search for date in CET doesn't return the resource, as it is shifted one (or two for CEST) hours.

Solution:

 * Pivot the "all day" date_start and date_end shifted from the resource timezone to UTC.

This pull request changes the behavior for pivoting (and searching) the "all day" date ranges.
Now the date for all-day date ranges is stored as if that date is in the timezone stored with the resource.
This timezone is in the `tz` property and will default to the timezone of the editor creating the resource.

If the `tz` property is not set, then the default Zotonic timezone is used.

Note that the way the date itself is stored in `date_start` and `date_end` is unaffected.

Also adds an option to the `date` and `date_range` filters to pass an integer as the timezone.
This is assumed to be a resource id, and the date is then displayed in the timezone of the resource's `tz` property.

**WARNING**

This introduces an **incompatibilty** where previously passing `1` as the timezone will use `UTC`. From now on pass `true` or `"UTC"` to get a time in UTC.   Passing `1` will use the timezone in the `tz` property of the resource with id 1.

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_pivot_rsc_job.erl
+++ b/apps/zotonic_core/src/support/z_pivot_rsc_job.erl
@@ -82,24 +82,24 @@ pivot_resource_update(Id, UpdateProps, RawProps, Context) ->
     end,
     ContextTz = z_context:set_tz(Tz, Context),
     Props1 = Props#{
-        <<"pivot_date_start">> => tz_all_day(IsAllDay, DateStart, ContextTz),
-        <<"pivot_date_end">> => tz_all_day(IsAllDay, DateEnd, ContextTz),
+        <<"pivot_date_start">> => tz_shift_all_day(IsAllDay, DateStart, ContextTz),
+        <<"pivot_date_end">> => tz_shift_all_day(IsAllDay, DateEnd, ContextTz),
         <<"pivot_date_start_month_day">> => month_day(DateStart),
         <<"pivot_date_end_month_day">> => month_day(DateEnd),
         <<"pivot_title">> => PivotTitle
     },
     z_notifier:foldr(#pivot_update{id=Id, raw_props=RawProps}, Props1, Context).
 
-tz_all_day(true, ?EPOCH_START = Date, _Context) ->
+tz_shift_all_day(true, ?EPOCH_START = Date, _Context) ->
     Date;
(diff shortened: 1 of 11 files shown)


robvandenbogaard:
I guess the timezone of "administrator" (resource id 1) would not be UTC by default? (probably Zotonic zone instead) Because if so, the effect of the incompatibility of passing 1 would be limited.

mmzeeman (approved):

apps/zotonic_core/src/support/z_pivot_rsc_job.erl
```diff
@@ -71,15 +71,36 @@ pivot_resource_update(Id, UpdateProps, RawProps, Context) ->
...
+    Date;
+tz_all_day(true, ?ST_JUTTEMIS = Date, _Context) ->
+    Date;
+tz_all_day(true, Date, Tz) when is_tuple(Date) ->
+    % All-day dates are stored without timezone conversion.
+    % For indexing we store them as-if they were entered in the
+    % timezone of the resource.
+    z_datetime:to_utc(Date, Tz);
+tz_all_day(_IsAllDay, Date, _Context) ->
+    Date.
```
robvandenbogaard, line 103: The current name for this function suggests it returns a tz; I think I'd rather call it something like `tz_shift_all_day`.

mworrell: Agree - will change

robvandenbogaard:
> @robvandenbogaard The `1` and the timezone might be problematic indeed. We could also log a warning and then just continue with whatever `1` has as a tz.
> 
> This is possible as this is a new feature and I don't expect it to be used immediately. Especially not for id 1.

Good to note there is a workaround in case `1` causes trouble - setting admin tz to utc.

robvandenbogaard (approved):

PR #3748, opened 2024-05-17
chore: allow postgres cloud db user like supabase with '.'

redgreat:
### Description
Modify the admin -> addsite page to accept the Postgres username 'dbuser' that is compatible with cloud database account names, such as 'supabase'. 

图片

### Checklist

- [✔️] documentation updated
- [❎] tests added
- [✔️] no BC breaks

### DIFF
--- a/apps/zotonic_mod_zotonic_site_management/priv/templates/addsite.tpl
+++ b/apps/zotonic_mod_zotonic_site_management/priv/templates/addsite.tpl
@@ -9,7 +9,7 @@
             <h3 class="panel-title">{_ Add or checkout a new site _}</h3>
         </div>
         <div class="panel-body">
-            {% wire id=#addsite 
+            {% wire id=#addsite
                     type="submit"
                     postback=`addsite`
                     delegate=`mod_zotonic_site_management`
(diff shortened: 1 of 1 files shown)


apps/zotonic_mod_zotonic_site_management/priv/templates/addsite.tpl
```diff
@@ -166,7 +166,7 @@
                             <div class="col-sm-7">
                                 <input class="form-control" type="text" id="{{ #dbuser }}" name="dbuser" value="" placeholder="{{ m.config.zotonic.dbuser.value|escape }}" />
                                 {% validate id=#dbuser name="dbuser"
-                                    type={format pattern="^[A-Za-z0-9_]*$" failure_message=_"Please use only a…z and 0…9 characters."}
+                                    type={format pattern="^[A-Za-z0-9_.]*$" failure_message=_"Please use only a…z and 0…9 and . characters."}
```
mworrell, line 169: Shouldn't that be `\\.` as otherwise it would accept any character, which the `.` does in regexps?

mworrell: Never mind - your solution is correct 👍

mworrell (approved):

PR #3753, opened 2024-05-24
Temporary fix fileuploader upload_count bug

### LINKED ISSUE #3751: Bug uploading a file via `fileuploader` worker
## `Zotonic` version
[1.0.0-rc.15](https://github.com/zotonic/zotonic/releases/tag/1.0.0-rc.15)

## `OS` / `Browser` version
Operating system: Ubuntu 22.04.1 LTS
Browser: Google Chrome Version 125.0.6422.76 (Official Build) (64-bit)
Erlang version: Erlang/OTP 26 [erts-14.1]

## Steps to reproduce
Uploading a file using `fileuploader` worker, but it does not always occur, so I don't have a reproducible repo or code. I would say that out of 10 times, one error occurs.

## Current behavior
Sometimes, it stays in a loop with 1 `upload_count` remaining:
image

The JS code is from [zotonic.fileuploader.worker.js](https://github.com/zotonic/zotonic/blob/0a61d8ce5abd41d4cc7cf230b60273c11c690e1f/apps/zotonic_mod_fileuploader/priv/lib/js/zotonic.fileuploader.worker.js#L282).

Maybe the check could be the `progress_msg` that contains the percentage, but there is a bug somewhere, and I don't know where.
The `upload_count` var is defined and updated only in the worker file.

## Expected behavior
`upload_count` be zero when the upload percentage is 100%.

williamthome:
### Description

Fix #3751

Sometimes, `upload_count` never goes to zero, causing an infinite loop. 
This PR is just a workaround for the issue.
I don't know the real cause of the problem, so I'm proposing a temporary fix.
Feel free to disagree with it :)

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_fileuploader/priv/lib/js/zotonic.fileuploader.worker.js
+++ b/apps/zotonic_mod_fileuploader/priv/lib/js/zotonic.fileuploader.worker.js
@@ -279,7 +279,12 @@ model.present = function(data) {
             if (r.failure_topic) {
                 self.publish(r.failure_topic, r.failure_msg);
             }
-        } else if (r.upload_count == 0 && r.wait_count == 0) {
+        // FIXME: Sometimes 'r.upload_count' never goes to zero.
+        //        It freezes in 1 and causes an infinite loop.
+        //        The 'r.progress_msg?.percentage === 100' is just a
+        //        workaround for the issue.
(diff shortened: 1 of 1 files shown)


mworrell:
See also #3754

williamthome:
Closing in favor of #3754

PR #3757, opened 2024-05-30
Use a bigger timeout for upgrade migration queries

pasqu4le:
### Description

This PR increases the timeout used by the version upgrade routines, as the standard timeout is often too small for the queries to succeed.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/db/z_db_pgsql.erl
+++ b/apps/zotonic_core/src/db/z_db_pgsql.erl
@@ -43,7 +43,8 @@
     squery/3,
     equery/4,
     execute_batch/4,
-    get_raw_connection/1
+    get_raw_connection/1,
+    get_raw_connection/2
 ]).
 
 %% Used by the z_install_update to access props columns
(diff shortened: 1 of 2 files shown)


apps/zotonic_core/src/db/z_db_pgsql.erl
```diff
@@ -259,9 +260,11 @@ is_tracing() ->
 %% @doc This function MUST NOT be used, but currently is required by the
 %% install / upgrade routines. Can only be called from inside a
 %% z_db:transaction/2.
-get_raw_connection(#context{dbc=Worker}) when Worker =/= undefined ->
-    gen_server:call(Worker, get_raw_connection).
+get_raw_connection(Context) ->
+    get_raw_connection(Context, undefined).
 
+get_raw_connection(#context{dbc=Worker}, Timeout) when Worker =/= undefined ->
+    gen_server:call(Worker, {get_raw_connection, Timeout}).
```
mworrell, line 267: If you increase the timeout to get a connection, then this `gen_server:call` should also get timeout that is just a bit more.  Otherwise this call will timeout (default 5000 msec) before the timeout of the raw-connection fetch.

mworrell:
I am a bit unsure about this change.

This is all about fetching a connection, not about performing the queries.
The queries are running with the connection, and have their own timeouts.

pasqu4le:
My bad, I misunderstood one timeout with another.

Closing this PR as there's more to look into 👌

PR #3773, opened 2024-06-10
mod_copyright: new module to administrate copyrights for content items

mworrell:
### Description

Adds a panel to administrate the rights for a resource.

Use the `rights` property and values from the Driebit `ginger` modules.

Decisions to take:

 - [ ] Use inline images for extra signage?
 - [ ] Which images for the Rights Statements?
 - [ ] A central panel for most of the settings, maybe also usable for timezone config etc.
 - [ ] URLs for CR and PD licenses (for the schema.org license property)
 - [ ] Use the author for attribution?  (Isn't necessarily the copyright holder...)

Also select a default schema.org mapping for the new properties.

https://schema.org/docs/search_results.html?q=license

Especially the "attribution" isn't something I can't find on schema.org

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_mod_copyright/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+This modules uses two sets of rights statements:
+
+ * https://creativecommons.org/licenses/#licenses-40
+ * https://rightsstatements.org/page/1.0/?language=en
+
+The properties manages are:
+
+ * `license`
+ * `copyright_attribution`
```
robvandenbogaard, line 15: I think the names of the properties were changed to `rights`, `rights_attribution` and, `rights_year`.

robvandenbogaard:
I also didn't find specific "attribution" predicates in neither Dublin Core nor Schema. I think they indeed use author or creator for it.

apps/zotonic_mod_copyright/README.md
mworrell: Indeed - fixing that.

PR #3781, opened 2024-06-19
Update frontend-pagebackgrounds.rst

### LINKED ISSUE #3774: Page Specific Backgrounds
## `Zotonic` version
git master

The Cookbook entry for [page specific backgrounds](https://zotonic.com/cookbook/1586/page-specific-dynamic-backgrounds) does not appear to work as documented...or perhaps something is wonky on my end.

When using the recommended code at the bottom of the page above, `id.background` is never true, even when present. The recommended code snippet is being used in the "base.tpl" template (copied from sample Zotonic site) in the `<body>` section. 

However, removing {% if id.background %} and specifically calling a particular resource does work, ie: `{% image_url m.rsc[600].medium.filename %}` This seems odd, as it implies that `{% id.anything %}` is being ignored in the base.tpl `<body>` section. 

Question: if following the cookbook guide referenced above, should `id.background` still be working, or has that changed since this cookbook entry was originally added?

Follow-up question: should `id.medium.filename` work if used in the `<body>` section of base.tpl, and a media is attached to the page being displayed?

Sorry, this query may sound confusing, but not sure how else to describe what I am seeing. Am I missing something obvious?

seannaswell:
Change id.background to id.o.background, reflecting updates to Zotonic since cookbook entry originally written.

### Description

Fix #3774 

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/doc/cookbook/frontend-pagebackgrounds.rst
+++ b/doc/cookbook/frontend-pagebackgrounds.rst
@@ -123,8 +123,8 @@ So, the code to go with our checklist is as follows:
 
 .. code-block:: django
 
-  <div{% if id.background %}{# we check to see if background exists #}
-      style="background: url( {% image_url id.background.medium.filename %}{# output web accessible URL to the image #} ) no-repeat"
+  <div{% if id.o.background %}{# we check to see if background exists #}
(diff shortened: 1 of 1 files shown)


mworrell:
Thanks!

PR #3802, opened 2024-06-28
Add qcountry to the search qargs options

DorienD:
### Description

Adds a qcountry option in the qargs to filter on country of a rsc.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


DorienD:
This was unnecessary, since `qpivot.country` already works

PR #3804, opened 2024-07-03
Encrypt Backups

mmzeeman:
### Description

Add the possibility to encrypt backups. This makes it safer to store privacy sensitive user content on external backup services. This PR also includes backing up configuration files. These are not downloadable via the admin interface for security reasons.

1) Add possibility to encrypt backups with a password.
2) Backup configuration and security files
3) CLI to decrypt backups with a password.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_mod_backup/src/models/m_backup.erl
```diff
@@ -30,6 +30,8 @@ m_get([ <<"admin_panel">> | Rest ], _Msg, Context) ->
     {ok, {m_config:get_value(mod_backup, admin_panel, Context), Rest}};
 m_get([ <<"daily_dump">> | Rest ], _Msg, Context) ->
     {ok, {m_config:get_value(mod_backup, daily_dump, Context), Rest}};
+m_get([ <<"encrypt_backups">> | Rest ], _Msg, Context) ->
+    {ok, {m_config:get_value(mod_backup, encrypt_backups, Context), Rest}};
```
mworrell, line 34: Isn't this a boolean?  Like the ones above also should be?

apps/zotonic_mod_backup/src/support/z_file_crypto.erl
```diff
@@ -0,0 +1,175 @@
+%% @author Maas-Maarten Zeeman <maas@channel.me>
+%% @copyright 2024 Maas-Maarten Zeeman
+%% @doc Decrypt and encrypt files.
+
```
mworrell, line 4: Add an `@end` here, as otherwise the APL text is added to the `@doc`.

mworrell: Maybe also rename the file to `backup_file_crypto.erl`  (or us it something that would be nice for other modules to use? then we could move it to stdlib or zotonic-support)

mworrell (approved):

apps/zotonic_mod_backup/src/models/m_backup.erl
mmzeeman: Yes

apps/zotonic_mod_backup/src/support/z_file_crypto.erl
mmzeeman: Going to rename this.

PR #3811, opened 2024-07-12
mod_logging/mod_development: add logging to the browser console

mworrell:
### Description

Add an option to log the erlang console log to the browser console.
All log events for the current site, or generic events without a `site` in the metadata are sent to the browser's console log.

This is turned on by default if the environment is `development` and the class `environment-development` is set on the html tag.

For non development sites it can be turned on by an admin on the `admin/development` page.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
😎

mmzeeman (approved):
😎

PR #3824, opened 2024-07-23
Mod_survey: Add a preview of the button

DorienD:
### Description

Adds a preview so users know which type of button they choose.


mworrell (approved):

DorienD:
Removed the include, this was not necessary anymore, since i used a little bit of JS for the change

Issue #3826, opened 2024-07-24
Allow for privacy friendly, cookie-less surfing of zotonic website


hce:
## `Zotonic` version
I'm using the latest github release, cca31849d3e86a5357f6b42aa261cc808cc3c0f5

## `OS` version
FreeBSD 14.1-RELEASE-p2

## Description
* **Motivation**

In some countries like Germany there are very strict laws in place with regards to cookie use, data protection etc. A simple solution to not having to deal with any of this might be to not use any cookies or other local storage on the website visiting user's browser. This makes sense especially for websites like blogs, whose purpose is to simply be read by users, no account required.

* **Proposal**

Add a setting to zotonic that allows to disable the use of all cookies, local and session storage as long as the zotonic website is browsed anonymously, i.e., as long as no-one is trying to log in.

## Current behavior

Currently a lot of cookies are set when visiting a zotonic website, both by the zotonic server as well as the javascript. Amongst those cookies there seem to be some that may be used for tracking purposes.

## Expected behavior

When the new setting disable_all_cookies is set to "true", no cookies, session or local storage would be set when visiting a zotonic powered site.

mworrell:
The cookies that we use are all functional cookies and as such in compliance with the GDPR, they are:

 - `z.lang` language selection. This is for keeping the language consistent and is needed when switching between pages with and without language code.
 - `cotonic-sid`, this is needed to keep communication between tabs working and for some functionalities (especially when server-side data need to be stored), and security of the `z.auth` token. Currently as cookie and in localStorage.
 - `z.auth` session related - this is the authentication cookie. It also contains a session id. This is needed for security and communication.
 - `z.tz` the timezone as found out from the configs, the user preference, and the browser. We need this to display dates and times in the correct format.

And temporary data in sessionStorage. Which doesn't matter for the GDPR.

What might be possible is to make the `z.lang` and `z.tz` session cookies for anonymous users. The same could be done for the `cotonic-sid` cookie. The `cotonic-sid`  localStorage might be more problematic, but maybe that could be replaced with only the cookie.

For analytics/social/etc. types of cookies we have `mod_cookie_consent`

/cc @mmzeeman

hce:
[...]

PR #3829, opened 2024-07-25
zotonic_launcher: use '@localhost' for default SNAME for shortnames listen on 127.0.0.1

mworrell:
### Description

Fixes an issue where the Erlang distribution would open a port listening on all IP addresses when using a SNAME.

Instead of the local hostname we will add 'localhost' as the hostname part.
This also fixes issues with machines that change names (like sometimes happens on macOS) and with Ubuntu where the hostname is per default mapped to 127.0.1.1

Issue #3826 

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
👍

PR #3835, opened 2024-08-08
Refactor smtp options.

### LINKED ISSUE #3834: Sending email from Zotonic with OTP-26 always falls back to sending mails unencrypted.
## `Zotonic` version
master

## `OS` / `Browser` version
Linux 

## Steps to reproduce
Send a test email form the admin console and observe the log.

## Current behavior
The connection to send mails is dropped and retried with the option `{tls, never}`.

## Expected behavior
TLS connections should be used when available.

## Config
Sending mails directly, without using a relay.

## Reason
The reason for this behavior is that since OTP-26 the default options for setting up a tls connection is now `{verify, verify_peer}` which needs one to supply a more options. When the connection is currently setup `gen_smtp` returns tls_failed, instead of reporting,
`{error, {options,incompatible, [{verify,verify_peer}, {cacerts,undefined}]}}]}`

## Solution
`gen_smtp` must be updated to either use `{verify, verify_none}` as default, or allow passing correct and safe `tls_options`. See also (https://github.com/gen-smtp/gen_smtp/pull/340)

@mworrell What do you think?

mmzeeman:
### Description

Fix #3834 

Set tls_options when they are not set (when no email relay is used)

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
I am a bit unsure about the `verify_none`, though the change would work correctly.

One of my ideas was that if we can't perform the verification (like for self-signed certs), then we should not pretend that the TLS connection is fine and instead just use a plain text connection.  As the plain text connection has the same security as a self-signed certificate (which is: none).

mmzeeman:
The `verify_none` was the default up to OTP 26. Since OTP-26 Zotonic will never use an encrypted smtp connection (without using a dedicated relay). I will have to check if it is possible to at least check the certificate against all the root certs.

Ideally one would check the name of the server too, but to be able to do that one needs extra information about the mx-record lookup from gen_smtp. It could be that `gen_smtp` should do this validation instead of zotonic, because we are not responsible for the mx-record lookup.

mworrell:
Hmmm, indeed gen_smtp knows which server we are connecting with. Would be better if then gen_smtp also does the verification.  Maybe we can fix that there, as we do have access to that project.

For the time being, let's merge your change, then at least we do send some encrypted email.

mworrell (approved):

PR #3836, opened 2024-08-16
Add os memory status

mmzeeman:
### Description

Add an alert that the system memory is running low and shows more system memory status.

Screenshot 2024-08-16 at 16 54 27

Screenshot 2024-08-16 at 16 36 59

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_launcher/src/zotonic_launcher.app.src
+++ b/apps/zotonic_launcher/src/zotonic_launcher.app.src
@@ -5,7 +5,7 @@
   {registered, []},
   {mod, {zotonic_launcher_app, []}},
   {applications, [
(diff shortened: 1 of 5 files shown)


robvandenbogaard:
Would it be unwieldy to monitor the memory used by Erlang/Zotonic in addition to this system wide overview? (Which is already nice!)

mmzeeman:
There already is an erlang memory allocation panel. The panel needs a better label and some extra explanation via an info button. You have to know a little bit about Erlang to understand what it means.

Screenshot 2024-08-21 at 10 33 39

mworrell (approved):

PR #3837, opened 2024-08-27
mod_fileuploader: pre-allocate uploaded file to prevent enospace during uploads

mworrell:
### Description

By pre-allocating all the space for the to-be-uploaded file we prevent a problem with running out of space during the file upload but instead can signal the disk space problem before the uploading starts.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


pasqu4le (approved):
🆒

PR #3851, opened 2024-09-05
Support the stay parameter in media upload dialog

pasqu4le:
### Description

This adds support for the `stay` parameter in media upload dialogs.

The reason for this MR is for retro-compatibility, as this parameter was available in the old Zotonic-0.x version of `action_admin_dialog_media_upload`.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell (approved):

mworrell:
Thanks!

PR #3857, opened 2024-09-16
Remove use of 'eval'  Fixes #2883

### LINKED ISSUE #2883: Remove eval, use injected script tag
Use a script tag with the correct nonce from #2870 

Replacing the eval with a script addition will allow to remove the `Content-Security-Policy: 'unsafe-eval'` header.

Also:

 - [ ] Check if a default CSP header can be added

----

Checklist:

 - [ ] Document
 - [ ] Tests
 - [ ] Backwards compatible?

mworrell:
### Description

 * Remove `eval` use from wired and widget manager
 * Change widget data to use JSON
 * Make `escape` filter serialize maps to JSON - making it easier to generate JSON in templates
 * Change all `onclick` attributes in javascript event listeners
 * Ensure that the CSP-nonce is passed back to the server for postback notifications (which might render style-tags)
 * Add a default CSP header (which is still much more relaxed than what I would like)

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mmzeeman (approved):
Nice PR to improve the security of zotonic sites.

CyBeRoni:
https://github.com/zotonic/zotonic/pull/3857/files#diff-a65c67508ab47f3327e445aa24960fcda3c55233f01a341224da458f5932745fR80

Here the nonce is provided to the document in the tag id, which can be accessed from javascript. This would allow an attacker to find out what the nonce is and add scripts with the correct nonce to the document.

apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
```diff
@@ -23,621 +23,695 @@ Based on nitrogen.js which is copyright 2008-2009 Rusty Klophaus
...
+  false,
+);
+
+function zotonic_eval(script) {
+  if (script) {
+    if (z_script_nonce) {
+      // nonce - assume CSP headers set
+      const s = document.createElement("script");
+      const head = document.getElementsByTagName("head")[0];
+      const prev = document.getElementById("zotonic-eval-" + z_script_nonce);
```
CyBeRoni, line 74: This provides the nonce value to javascript

apps/zotonic_mod_wires/priv/lib/js/apps/zotonic-wired.js
```diff
@@ -23,621 +23,695 @@ Based on nitrogen.js which is copyright 2008-2009 Rusty Klophaus
...
+      if (prev) {
+        prev.remove();
+      }
+
+      s.type = "text/javascript";
+      s.id = "zotonic-eval-" + z_script_nonce;
```
CyBeRoni, line 80: This provides the nonce value to javascript

mmzeeman: That doesn't really give new capabilities. Once an attacker has script access it can simply get the nonce by getting it from a script tag.

```javascript
document.getElementsByTagName("script")[0].nonce
```

mworrell: Exactly what @mmzeeman writes - if you can run JS (and the DOM), then you have access to the nonce by inspecting script/style tags.

So, the nonce is just a lock on the proverbial Javascript door, if you are inside then you have full access to everything (and all keys).

mworrell: Leaving this as "unresolved" so that the discussion is clear.

CyBeRoni: Accessed like that you should find an empty value returned. See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce#accessing_nonces_and_nonce_hiding

Putting the nonce in an id however subverts this protection. 

I misspoke slightly in that the nonce can be had from javascript, albeit not like that, (see above link, and this is true in any case if you are programmatically generating the script tag and adding the nonce to it) but putting in a script id allows *other* things to potentially access it which is why it, as a tag attribute, is hidden.

mworrell: Ah, that is what you meant.   I will change that and make the tag id a bit less "unique".

mworrell: @CyBeRoni replaced it with one of the unique ids generated by Zotonic.

pasqu4le (approved):

PR #3860, opened 2024-09-17
core: default csp for webworker and mediastream

mworrell:
### Description

Allow

* mediastream for media
* blob for webworker

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


pasqu4le (approved):
LGTM 👍

PR #3864, opened 2024-09-19
[zotonic_status] Module health status

robvandenbogaard:
### Description

Proof of concept for discussing whether it would be good to have an indication of site module status in the status overview.
Use case is any server with a lot of sites, to quickly be able to see whether any site modules happen to not have started or have stopped due to e.g. a maintenance restart.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
Ok with me to merge this.

Maybe the 0.x module manager could also have a small update to periodically try a module restart for all "non healthy" modules. Just a dumb restart - no extra supervision.

PR #3868, opened 2024-09-23
mod_search: ensure all modules and site is running before (re)creating the facet table

mworrell:
### Description

When creating the pivot facet table, the current site module _must_ be running.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_module_manager.erl
+++ b/apps/zotonic_core/src/support/z_module_manager.erl
@@ -66,7 +66,6 @@
     mod_author/1,
     mod_schema/1,
     reinstall/2,
-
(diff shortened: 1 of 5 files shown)


apps/zotonic_core/src/support/z_module_manager.erl
```diff
@@ -65,7 +65,9 @@
     mod_description/1,
     mod_author/1,
     mod_schema/1,
-    reinstall/2
+    reinstall/2,
+
```
vkatsuba, line 69:
```suggestion
```

PR #3880, opened 2024-10-01
Session overview page

pasqu4le:
### Description

This adds a session overview page where logged in users can:
1. see all their recent known logons (with user agent and IP address)
2. see all their currently active sessions
3. interrupt all active session

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


apps/zotonic_core/src/install/z_install.erl
```diff
@@ -389,6 +389,30 @@ model_pgsql() ->
...
+      CONSTRAINT id_log_pkey PRIMARY KEY (id),
+      CONSTRAINT pk_id_log_rsc_id FOREIGN KEY (rsc_id)
+        REFERENCES rsc (id)
+        ON UPDATE CASCADE ON DELETE CASCADE,
+      CONSTRAINT pk_id_log_identity_id FOREIGN KEY (identity_id)
+        REFERENCES identity (id)
+        ON UPDATE CASCADE ON DELETE CASCADE
+    )",
+
+    "CREATE INDEX IF NOT EXISTS fki_identity_log_rsc_id ON identity (rsc_id)",
```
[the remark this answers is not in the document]
mworrell: Also for the key below.

apps/zotonic_core/src/install/z_install.erl
```diff
@@ -389,6 +389,30 @@ model_pgsql() ->
...
+    (
+      id serial NOT NULL,
+      identity_id int NOT NULL,
+      rsc_id int NOT NULL,
+      created timestamp with time zone NOT NULL DEFAULT now(),
+      user_agent character varying (240),
+      ip_address character varying (40),
+
+      CONSTRAINT id_log_pkey PRIMARY KEY (id),
+      CONSTRAINT pk_id_log_rsc_id FOREIGN KEY (rsc_id)
```
mworrell, line 405: We usually name these foreign key constraints `fk_...` and not `pk_...` 

As `identity_id` has a foreign key constraint we also need an index on that columns to prevent table scans on deletion of an identity table row.

apps/zotonic_core/src/models/m_identity.erl
```diff
@@ -1477,27 +1477,68 @@ insert_unique(RscId, Type, Key, Props, Context) ->
...
+    z_db:transaction(fun(Ctx) -> set_visited_trans(UserId, Ctx) end, Context);
+set_visited(User, Context) ->
+    set_visited(m_rsc:rid(User, Context), Context).
+
+% Part of 'set_visited' to execute in one DB transaction
+set_visited_trans(UserId, Context) when is_integer(UserId) ->
+    case z_db:q1("select id from identity where rsc_id = $1 and type = 'username_pw'", [UserId], Context) of
+        undefined -> {error, enoent};
+        EntryId ->
+            UserAgent = z_convert:to_binary(m_req:get(user_agent, Context)),
```
mworrell, line 1492: Better to truncate the UA to max 240 chars using `z_string:truncatechars/2`

apps/zotonic_core/src/models/m_identity.erl
```diff
@@ -1477,27 +1477,68 @@ insert_unique(RscId, Type, Key, Props, Context) ->
...
+            IpAddress = z_convert:to_binary(m_req:get(peer, Context)),
```
mworrell, line 1493: Maybe use `inet_parse:ntoa/1` (and then iolist_to_binary).
Of course will need a check for the case where the peer is `undefined`.

apps/zotonic_mod_authentication/priv/dispatch/dispatch
```diff
@@ -19,6 +19,9 @@
  {logon_reminder,   ["logon", "reminder"],                controller_template, [ {template, "logon.tpl"}, seo_noindex, is_logon ]},
  {logon_confirm,    ["logon", "confirm"],                 controller_template, [ {template, "logon.tpl"}, seo_noindex, is_logon ]},
 
+ % Simple session overview
+ {sessions, ["sessions"], controller_template, [ {template, "sessions.tpl"}, seo_noindex, {acl, is_auth} ]},
```
mworrell, line 23: In the 0.x it was:

```erlang
 {sessions_user, ["logon", "sessions"], ...}]}
```

Maybe we can give it the same name and path on the 1.x?

apps/zotonic_mod_authentication/priv/templates/_dialog_session_close.tpl
```diff
@@ -0,0 +1,14 @@
+
+<p>
+	{_ Every sessions will be interrupted and logged out of your account. _}<br>
```
mworrell, line 3: Maybe: 

`Every session will be interrupted and you will be logged out.`

apps/zotonic_mod_authentication/priv/templates/sessions.tpl
```diff
@@ -0,0 +1,139 @@
...
+                        {% endfor %}
+                    </tbody>
+                </table>
+            </div>
+        </div>
+    </div>
+</div>
+
+{% javascript %}
+cotonic.broker.subscribe("bridge/origin/test/{{ m.acl.user.id }}/session_extended", function(msg) {
```
mworrell, line 112: Better use the topic:

``user/:user_id/session_extended``

As the `test` topic is public and then everybody can listen in.
The `user/:user_id/...` topic is restricted to the user or those that can _edit_ the user.

apps/zotonic_mod_authentication/src/mod_authentication.erl
```diff
@@ -95,6 +114,26 @@ observe_request_context(#request_context{ phase = init }, Context, _Context) ->
...
+    % Notify that a client session for a user is still active
+    % (it has been extended with a status/refresh ping)
+    case z_acl:user(Context) of
+        undefined ->
+            Context;
+        UserId ->
+            case z_context:client_id(Context) of
+                {ok, ClientId} ->
+                    z_mqtt:publish(
+                        [<<"test">>, z_convert:to_binary(UserId), <<"session_extended">>],
```
mworrell, line 128: See comment above for the topic name.

apps/zotonic_core/src/install/z_install.erl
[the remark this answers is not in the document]
pasqu4le: Well spotted, fixed in 12c52c56b165968cdafcc47727eac3aa06924584

apps/zotonic_core/src/install/z_install.erl
pasqu4le: Makes sense, fixed in 12c52c56b165968cdafcc47727eac3aa06924584

apps/zotonic_core/src/models/m_identity.erl
pasqu4le: Done in 6d385cf382a3afe72a2281c5f24ca3e16a8edc42

apps/zotonic_mod_authentication/priv/dispatch/dispatch
pasqu4le: 6d46918d36cf9d602b9b9efbe64c1e78781e78b8 👍

apps/zotonic_mod_authentication/priv/templates/_dialog_session_close.tpl
pasqu4le: Sounds good, done in c52ab1f2abfec75e16bcc7895ed8abcab04d010e

apps/zotonic_mod_authentication/priv/templates/sessions.tpl
pasqu4le: Forgot to change it from `test`, switched to `user` in c52ab1f2abfec75e16bcc7895ed8abcab04d010e

PR #3881, opened 2024-10-01
mod_backup: stricter periodic cleanup of revisions for users

pasqu4le:
### Description

This is a follow-up to #3546 covering the point:
> For the cleanup of person data (GDPR) we will need a shorter period.

Specifically, this adds (configurable) periods to:
1. delete revisions of active user's resources (defaults to 90 days);
2. delete all revisions of users that have been deleted for some time (defaults to 30 days);

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/doc/ref/modules/mod_backup.rst
+++ b/doc/ref/modules/mod_backup.rst
@@ -55,8 +55,15 @@ is saved to a revision log.
 Using the revision log a resource can be rolled back to an older revision
 or, when deleted, recovered.
 
-Revisions older than 18 months are daily pruned. This can be changed by
-setting the configuration ``mod_backup.revision_retention_months`` to another
-number of months.
+Revisions are pruned daily and deleted if:
+1. older than 18 months;
+   This can be changed by setting the configuration ``mod_backup.revision_retention_months``
+   to another number of months.
+2. of user resources and older than 90 days;
+   This can be changed by setting the configuration ``mod_backup.user_revision_retention_days``
+   to another number of days.
+3. of users resources and the users was deleted for at least 30 days;
+   This can be changed by setting the configuration ``mod_backup.user_deletion_retention_days``
+   to another number of days (maximum 30).
 
 Currently edges (connections) and medium files are not kept in the revision log.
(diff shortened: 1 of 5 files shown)


doc/ref/modules/mod_backup.rst
```diff
@@ -55,8 +55,16 @@ is saved to a revision log.
...
+
+1. older than 18 months;
+   This can be changed by setting the configuration ``mod_backup.revision_retention_months``
+   to another number of months.
+2. of user resources and older than 90 days;
+   This can be changed by setting the configuration ``mod_backup.user_revision_retention_days``
+   to another number of days.
+3. of users resources and the users was deleted for at least 30 days;
+   This can be changed by setting the configuration ``mod_backup.user_deletion_retention_days``
+   to another number of days (maximum 30).
```
mworrell, line 68: Why is there a maximum of 30 days?

mworrell: Ah I see in the original issue description.

mworrell:
I see that there is  max of 30 days for the retention of user-resource-revisions after deletion.

This is accordance with the GDPR which states that personal information should be removed within 30 days after a request from the person to do so.

As this is GDPR related we might want to make the test a bit less strict.  Shouldn't we delete all _personal_ information after such a request? Think of records in a CRM.

Assuming such information is contained in a `person` resource, we could say that all (stored backups of) resources that are either _person_ or _user_ resources should be purged after 30 days.

Then we could make the `is_user` flag something like "this is information subject to the GDPR."   Which might make it (semantically) possible for us to add "quicker" purge to more kinds of information.

Maybe even call it `is_short_retention` ?

I know it is a bit splitting hairs :smile: - just that I would like to make this function as clear as possible. And then we could even set the flag if some depiction of a person is deleted.  Or add a little checkbox in the delete-confirm to set the shorter retention period on deletion (for the resource and all its depending resources).

pasqu4le:
@mworrell I agree and as we discussed elsewhere, I've changed the name from `has_identity`/`is_user` to `is_personal_data` (see e4b00ded7424f5e7eff9bf8895f97c457bec97dc).

PR #3900, opened 2024-10-11
zotonic_status: Automatically restart site modules if needed

robvandenbogaard:
### Description

If site modules crash after a zotonic upgrade or server restart they will get restarted by the site manager. The check runs once every 10 minutes.

Logs a debug message on check and a warning on any unhealthy module it finds and is about to restart.

As suggested in https://github.com/zotonic/zotonic/pull/3864 (PR on site module health indicators).

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell (approved):

mworrell:
Good workaround for the problems encountered.

PR #3909, opened 2024-11-03
Pass cat correctly through to action dialog connect

DorienD:
### Description

In the `action_dialog_connect.tpl` the category that you could pass with `cat=cat` was not properly followed through. So i added it in the `with` of the template as the first option.

### DIFF
--- a/apps/zotonic_mod_admin/priv/templates/_admin_edit_content_page_connections_list.tpl
+++ b/apps/zotonic_mod_admin/priv/templates/_admin_edit_content_page_connections_list.tpl
@@ -17,6 +17,7 @@ Params:
 Config key:
 - mod_admin.edge_list_max_length -- max number of items shown per predicate
 #}
+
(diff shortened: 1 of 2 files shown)


apps/zotonic_mod_admin/priv/templates/_admin_edit_content_page_connections_list.tpl
```diff
@@ -56,6 +58,7 @@ Config key:
           tabs_enabled=tabs_enabled
           callback=callback
           nocatselect=nocatselect
+          cat=cat
```
mworrell, line 61: I think we should use:

```
category=category
```

And then add `category` to the list of Params at the top.

mworrell:
Probably better to use `category` directly.
Which might need some additional documentation.

mworrell:
Changed in c42fbfa15c92f9c84d1a0978370cb275bf36973a

PR #3927, opened 2024-11-08
Show username admin overview list

mmzeeman:
### Description

Show the username of a person resource on the admin overview page as an extra label next to the title. The labels will disappear when the viewport gets to narrow in order to prevent cluttering the interface on mobile.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_mod_admin/priv/templates/_admin_overview_list.tpl
```diff
@@ -121,18 +129,24 @@ qcat
...
                 {% elseif id.name %}
-                    <span class="label label-default pull-right">
+                    <span class="label label-default pull-right hidden-xs">
                         {{ id.name }}
                     </span>
                 {% endif %}
-                <span {% include "_language_attrs.tpl" %}>{{ id.title|default:id.short_title|striptags|default:_"<em>Untitled</em>" }}</span>
-                {% if id.is_protected %}
-                    &nbsp; <span class="fa fa-lock text-muted" title="{_ Protected, not deletable _}"></span>
+                {% if id | is_a:"person" %}
```
mworrell, line 146: Logically anything can be a user.  Should we just check on username?   That will give a query for every row (am not sure if it is cached).

mmzeeman: Yes, probably not. That is why I limited it to person resources for now. I also don't think the username panel will appear in the admin for non-person resources.

mworrell: Maybe each resource - at OpenResearch they have non-person resources with a username.

mworrell:
Merge whenever you want 👍

PR #3929, opened 2024-11-10
Set the cotonic-routing-id only when a cotonic_sid is set

mmzeeman:
### Description

Let the mqtt-session runtime used in zotonic return the `cotonic-routing-id`. Currently this routing id is always set by mqtt_sessions, but some client implementations (EasyMQTT) crashes when Zotonic returns the routing id as property attached to the `connack`

With this change the mqtt runtime in zotonic is responsible for returning the `cotonic-routing-id`. The `routing-id` is only passed when the properties found in the `connack` message contains a `cotonic_sid`.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell (approved):

mworrell:
Good to move this from the MQTT code to here.

PR #3936, opened 2024-11-13
mod_oauth2: add extra logging on oauth redirect

mworrell:
### Description

This gives extra feedback if there are problems with the access token exchange for Auth2 logon services.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


pasqu4le (approved):
👍

PR #3956, opened 2024-11-25
Show admin warnings only to admin users

pasqu4le:
### Description

Problem: SSL configuration warnings in the admin status are misreported when the user is not an admin.
When 'm.admin_status.is_ssl_application_configured' is rendered for a non-admin, the result is 'undefined' which is in turn negated to 'true' by a 'not', meaning that the warning always appears for these users.

Solution: Only show warnings in the admin panel to admin users.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell (approved):

mworrell:
LGTM 👍

PR #3975, opened 2024-12-06
mod_admin: in the UI, rename 'block' to 'page block'

mworrell:
### Description

This fixes an issue where there is confusion between template blocks and page blocks.

See discussion #3955 

The documentation needs to be adapted separately, for example when we cleanup the documentation (and move the reference docs into the erlang modules).

We settled on _page block_ and not on _page section_ to prevent miscommunication between developers (who see the term _blocks_ in their code) and end-users/project-managers (who would see _section_, which is then confusing).

In the documentation we need to consistently use the term _page block_ and _template block_ to disambiguate the meaning of the word _block_.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Goed om verwarring te voorkomen.

PR #3988, opened 2024-12-20
mod_wires: signal user if they are having difficulties with authentication

mworrell:
### Description

This fixes an issue with old Safari (14 and earlier) where the auth exchange with the server can not be started and thus no interactivity would be possible.

This is caused by not being able to load the Auth web worker.

We now signal this issue and show a growl message if after 10 seconds the started auth worker is still not registered.

Screenshot 2024-12-20 at 12 43 04

For now this message is only in English.

Also part of this merge:

 * Remove `*` from the `z_context:get_q_all_noz/1` query args  (now compatible with the map version)
 * Add `wss:` to the default CSP connect-src, for older Safari browsers

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_context.erl
+++ b/apps/zotonic_core/src/support/z_context.erl
@@ -808,8 +808,8 @@ get_q_all(Key, #context{ props = Props }) ->
     end.
 
 
-%% @doc Get all query/post args, filter the zotonic internal args. Leave the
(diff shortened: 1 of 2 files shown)


mmzeeman (approved):
Ok

EdwardAbboud (approved):
nice!

PR #3999, opened 2025-01-14
Use glyphicon for icons, instead of font-awesome

mworrell:
### Description

This fixes a problem where the icons are only visible if the mod_artwork module is enabled.
That module is not always enabled for every site.

The glyph icons are uglier than font-awesome, but do work.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@DorienD What do you think?  Is this a good one, or use font-awesome and add more dependencies on some modules?

mworrell:
Closed, as these templates are out of date.
Keeping this so that we can revisit the changes later.

PR #4008, opened 2025-01-21
Add nonce support functions and protect against resubmit of same form

mworrell:
### Description

If we redirect after a form submit, then some browsers allow 'back' and then show the submit page again.
That allows people to resubmit the form multiple times.

This change protects the form against re-submitting by adding a unique 'nonce' and checking if that 'nonce' has been submitted already.

The nonce registration is added in `z_nonce`, which uses ets tables with generational cleanup. Currently nonce values are remembered for about 600 seconds.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (commented):
That is quite an nice addition to have for a lot of security protocols. 

This implementation registeres every in-flight nonce in a table. Which can become full and overloaded. Possibly by someone reloading a page. 

A while ago I also made a nonce service which prevents having to create a big administration of in-flight nonce's. 

I did this like this... 

- A nonce is an encrypted sequence number (so you can check if it is something you created yourself)
- It is a sequence number so you can just keep track  of the highest (and lowest) possibly valid ones. 
- Every now and then you increase your lowest used nonce and dump all used nonces with a lower number than you registered as being used.
- When a nonce is used up (after crypto checking, and checking the lower and upper bound, and the registered nonces), then you register it.

mworrell:
The implementation right now registers when a nonce is used.  Not the in-flight nonce values, which are just randomly generated.

I like the idea of a timestamp and encryption to check if it is something we created ourselves within the last period.  We could just add a sha2 hash to the value, using some secret.

Do we have a secret that we can use for this?

mworrell:
@mmzeeman I have tuned the nonce generation, so that we can check if we have generated them and also have a max validity built into the nonce token.

Also added a method to register _any_ keys, so that nonces can be checked for replays on API calls, which might use externally generated nonce values.

Can you have a look?

apps/zotonic_core/src/support/z_nonce.erl
```diff
@@ -0,0 +1,291 @@
...
+%% seconds or until unregistered. The nonce must be generated with nonce/1 and
+%% will be checked if it was expired.
+register(<<>>) ->
+    {error, key};
+register(Nonce) ->
+    Oldest = nonce_oldest_generation(),
+    case split_nonce(Nonce) of
+        {Number, Random, Hash} when Number >= Oldest ->
+            CheckKey = <<(integer_to_binary(Number))/binary, "-", Random/binary>>,
+            CheckHash = crypto:mac(hmac, sha256, CheckKey, nonce_secret()),
```
mmzeeman, line 119: If you want to have a cryptographically secure nonce, then you have to make sure you can also check nonces after restarts and crashes. The ets tables will disappear after a crash, but the nonce_secret will survive the restart. This means that the nonce can be used multiple times. A solution for this is to create a new secret for every restart. It has as downside that all in flight nonces are invalidated.

mmzeeman (approved):
Must this nonce service also be cryptographically strong? Say, used for external logon providers like "Sign in with Whatever"?

mworrell:
The hash is a hmac with sha256, is that strong enough?
The random key is generated with strong-random-bytes.

apps/zotonic_core/src/support/z_nonce.erl
mworrell: I much prefer to not have the inflight nonces be invalidated.
As that would result all kinds of requests to fail after a restart.

We could add an API for a "strong" nonce - where we re-generate the key after each restart.

mmzeeman:
> The hash is a hmac with sha256, is that strong enough? The random key is generated with strong-random-bytes.

Yes

PR #4015, opened 2025-01-24
Docker: use debian base image

mworrell:
### Description

The `sass` command is not available on Alpine Linux.

Switch to de Debian image to be able to use `sass`.
This gives a better development environment.

Also fix an issue with the security directory not being placed in docker-data.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


EdwardAbboud (approved):
Nice! Looks good to me

PR #4025, opened 2025-01-28
mod_filestore: only delete files if service url matches

### LINKED ISSUE #3726: mod_filestore: prevent destructive actions if file location does not match configured setting
## `Zotonic` version
0.x and 1.0

## Description
**Motivation**

If you use mod_filestore and move data between different DTAP-environments of your websites, it is easily possible to have mod_filestore data that is set to the "wrong" URL because the filestore table saves the full URL for the uploaded file. For example, if I use AWS S3 and my bucket location is set to `https://example-prod.s3.eu-west-1.amazonaws.com/`, then a file uploaded to S3 by Zotonic will have a `location` field of something like `https://example-prod.s3.eu-west-1.amazonaws.com/archive/today/file.png`. 

If I then take my production environment database and import it to a test environment (with its own S3 bucket, `example-test`) then newly uploaded files will instead have something like `https://example-test.s3.eu-west-1.amazonaws.com/archive/today/testfile.png` as its `location` field. But the original upload will still have a reference to the `-prod` bucket.

So long as the authentication data used for the `test` environment has at least read access to the `prod` bucket, this will work and the site will show files as normal. 

However, if it also has write access, a potential issue occurs when files are to be deleted from the `test` environment. Because the `location` field for a file originally uploaded to the `prod` environment still references the `prod` bucket, deleting that file in the `test` environment will delete the file from the `prod` bucket.

**Proposal**

While I do think it is advantageous to allow cross-environment use of data (if nothing else to reduce my storage bill), the ability for destructive actions to happen is not wanted.

My proposal is that for destructive actions (which I think would only be deletes), the action only be taken if the URL of the file's `location` field matches the currently configured filestore URL. 

This would allow use of files between environments or even websites in a safe way if the database were copied between them, provided the S3 bucked URL is changed in the non-production environment . This reduces the amount of data that needs to be stored (and paid for), while preventing inadvertent destruction of data.

While it is possible to solve this issue on the side of AWS, either by using per-environment IAM info (and denying write access to other envs) or using bucket versioning, I think that solving this on the Zotonic side would be preferable for the following reasons:
 - It allows cross-env use of data, reducing data duplication and preventing data from becoming inaccessible without manual intervention
 - Using bucket versioning increases costs because deleted files are saved, and using lifecycle rules to limit that and permanently delete files after some period removes the benefit if there is no way to detect and undo the errant deletion before the lifecycle rule permanently deletes the file.
 - It should not be assumed that all available storage backends (even if S3-compatible) implement the requisite access controls and versioning options.
  
## Current behavior
File is deleted at `location` if authentication data provides the required access to delete files.

## Expected behavior
File should only be deleted if above requirements are fulfilled.

mworrell:
### Description

Fix #3726 

Before deleting a file from the filestore, check if the service url matches with the location to be deleted.
This fixes an issue where files on a service that is not configured could be deleted if the credentials were valid for that service as well.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/include/zotonic_file.hrl
+++ b/apps/zotonic_core/include/zotonic_file.hrl
@@ -1,8 +1,9 @@
 %% @author Marc Worrell <marc@worrell.nl>
-%% @copyright 2013-2022 Marc Worrell
+%% @copyright 2013-2025 Marc Worrell
 %% @doc File/media interface definitions. See also z_media_request.erl
+%% @end
 
-%% Copyright 2013-2022 Marc Worrell
+%% Copyright 2013-2025 Marc Worrell
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
@@ -27,11 +28,14 @@
 -record(filestore, {
         action = upload :: lookup | upload | delete,
         path :: file:filename_all(),
-        filename :: file:filename_all() | undefined,
+        filename :: file:filename_all()
+                  | {prefix, file:filename_all()}
+                  | undefined,
         mime :: binary() | string() | undefined
     }).
 
-%% @doc Synchronous download, upload or delete a file on the filestore service.
+%% @doc Synchronous (first) notification to download, upload or delete
+%% a file for a filestore service.
 %% Report the result to the handler_pid.
 %% Where Result is ok or {error, Reason}.
 -record(filestore_request, {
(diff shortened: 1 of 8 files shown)


CyBeRoni:
Can this be backported to 0.x?

mworrell:
@CyBeRoni probably, yes.

PR #4027, opened 2025-01-29
core: add `#content_security_header` notification to modify the CSP

mworrell:
### Description

Also, for Google trackers amend the CSP headers and update the GTM javascript to be aware of the nonce.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_context.erl
+++ b/apps/zotonic_core/src/support/z_context.erl
@@ -1363,22 +1363,25 @@ set_nocache_headers(Context = #context{cowreq=Req}) when is_map(Req) ->
 %%      'security_headers' notification.
 -spec set_security_headers( z:context() ) -> z:context().
 set_security_headers(Context) ->
+    CSP = #content_security_headers{
+        default_src    = [ <<"'self'">> ],
+        script_src     = [ <<"'self'">>, <<"'nonce-'">>, <<"https:">> ],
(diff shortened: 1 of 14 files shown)


apps/zotonic_core/src/support/z_context.erl
```diff
@@ -1412,6 +1415,47 @@ set_security_headers(Context) ->
...
+        fun
+            ({_Directive, []}) -> false;
+            ({Directive, L}) ->
+                L1 = lists:usort(L),
+                L2 = case lists:member(<<"'unsafe-inline'">>, L1) of
+                    true -> lists:delete(<<"'nonce-'">>, L1);
+                    false -> L1
+                end,
+                D = [ Directive | [ [ " ", V ] || V <- L2 ] ],
+                {true, D}
```
pasqu4le, line 1429: Curious: is this necessary or is it merely for convenience?

I suppose `unsafe-inline` makes the `nonce` unnecessary, but is there any downside in keeping both?

pasqu4le (approved):
Nice, this will come useful 👍

apps/zotonic_core/src/support/z_context.erl
mworrell: I remember finding out that if you have both that some browsers start to get confused and apply the stricter nonce, which is a bit weird, but maybe it is hidden somewhere in the specs.

mmzeeman (approved):

PR #4040, opened 2025-02-11
mod_oauth2: add option to manually manage OAuth2 tokens

mworrell:
### Description

This makes it possible to:

 * manually add/remove OAuth2 tokens for users; and
 * fetch tokens from Zotonic sites using Client Credentials.

TODO:

- [x] Generate manual token
- [x] Select user for manual token
- [x] Enter manual token for app
- [x] List tokens on server
- [x] List tokens on client
- [x] Select user on client side for token fetched with client-credentials
- [x] Use client-credentials with Zotonic sites
- [x] Select user on server side for client credentials
- [ ] ~Paging of server tokens~ **later**
- [ ] ~Paging of client tokens~ **later**

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_identity.erl
+++ b/apps/zotonic_core/src/models/m_identity.erl
@@ -89,6 +89,7 @@
     is_verified/2,
     verify_primary_email/2,
 
+    flush/2,
     delete/2,
     merge/3,
     is_reserved_name/1,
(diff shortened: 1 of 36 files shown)


mmzeeman (approved):
There were also some updates in the survey module. Are those ok too?

mworrell:
Yes, I added those too, by accident, but they work well and are very small.

PR #4042, opened 2025-02-12
Removed apns4erl as dep. It is not used by zotonic

mmzeeman:
### Description

The zotonic lock file contains a dep of a package which is not used by zotonic. It is a module I use for sending apple notifications.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman:
This dep was somehow "escaped" from my dev environment. 

Are there other deps which should not be there?

vkatsuba:
> This dep was somehow "escaped" from my dev environment.
> 
> Are there other deps which should not be there?

The `apns4erl` is use libs `gun`, `jsx`, `base64url` deps - all of them also used in Zotonic, except maybe `base64url` lib(https://github.com/dvv/base64url) - I can be wrong but Zotonic use own implementation with same name. Maybe it makes sense rebuild `rebar.config` from scratch.

mworrell:
Indeed, we have a base64url hanging around in the core. Maybe good to check where that one is used and if it is the same one as that package.

PR #4044, opened 2025-02-13
core: fix trans lookup for nested lists

mworrell:
### Description

Fix a problem with translating nested lists of non-character data.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


apps/zotonic_core/src/i18n/z_trans.erl
```diff
@@ -203,6 +211,10 @@ lookup_fallback(_Text, _Lang, _Context) ->
     undefined.
 
 
+is_string([H|T]) when is_integer(H), H >= 1 -> is_string(T);
+is_string([H|T]) when is_binary(H) -> is_string(T);
+is_string(_) -> false.
```
vkatsuba, line 216: This function always return `false` but expected a `true` or `false`.

mworrell: Good one :-)

PR #4045, opened 2025-02-18
auth: ACL support for switching users (and back)

mworrell:
### Description

This adds an ACL check for testing if a user is allowed to switch to another user.
Also adds `m.acl.sudo_user` to check the original user account the user used to log on.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_auth.erl
+++ b/apps/zotonic_core/src/support/z_auth.erl
@@ -89,27 +89,47 @@ logon(UserId, Context) ->
 
 %% @doc Allow an admin user to switch to another user account.
 -spec logon_switch( m_rsc:resource_id(), z:context() ) -> {ok, z:context()} | {error, eacces}.
-logon_switch(UserId, Context) ->
-    case m_rsc:exists(UserId, Context) andalso z_acl:is_admin(Context) of
+logon_switch(ToUserId, Context) ->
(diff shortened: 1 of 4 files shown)


apps/zotonic_core/src/support/z_auth.erl
```diff
@@ -89,28 +89,48 @@ logon(UserId, Context) ->
...
         false ->
             {error, eacces}
     end.
 
+is_allowed_switch_user(?ACL_ADMIN_USER_ID, Context) ->
+    % Hard coded protection against logging in as the admin user from any account that
+    % is not the admin user.
+    z_acl:user(Context) =:= ?ACL_ADMIN_USER_ID;
+is_allowed_switch_user(ToUserId, Context) ->
+    m_rsc:exists(ToUserId, Context)
```
mmzeeman, line 126: During normal logins we also check the user resource is published. Should this also be done here?

mmzeeman (commented):
Added a question regarding the "published" status of the user resource switching to.

apps/zotonic_core/src/support/z_auth.erl
mworrell: Good question. My first thought was *yes*, but then I thought that it is actually useful to be able to logon as an user that is not yet published to test their access/account.

PR #4046, opened 2025-02-20
core: Add qs to the log request

mmzeeman:
### Description

Adds the query string to the log request, so it is possible to include it in access log metrics.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Just to be sure, that is `qs` from the URL, and not `qs` from the POST?

mmzeeman:
Yes it is supposed to be the qs of the URI. See: https://ninenines.eu/docs/en/cowboy/2.13/manual/cowboy_req.qs/

Note: `prune_reqdata` in `z_context` always sets `qs` to `<<>>` even when it was not set.

mworrell (approved):

PR #4048, opened 2025-02-21
Add a model for safer access to query parameters

pasqu4le:
### Description

This adds a new model for safer access to query parameter based on [the XSS prevention docs](http://zotonic.com/cookbook/2325/security-templates-and-xss-prevention).

I've marked this PR as a draft because I'd like to mention it in said documentation page as well, but also because before doing that I'd like to gather opinions/reviews on its interface.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_qarg.erl
+++ b/apps/zotonic_core/src/models/m_qarg.erl
@@ -1,0 +1,68 @@
+%% @author Driebit <tech@driebit.nl>
+%% @copyright 2025 Driebit
+%% @doc Model for safely accessing query arguments.
+%% @end
+
+-module(m_qarg).
+
+-behaviour(zotonic_model).
+
+-export([
+    m_get/3
+]).
+
+-include_lib("zotonic.hrl").
+
+%% @doc This model provides a safer alternative to 'q' and 'm.req' for obtaining
+%% the request's query arguments from templates.
+%% It does so by implementing convenient syntax for the recommendations in the
+%% "Special care" of the XSS prevention documentation:
+%% https://zotonic.com/cookbook/2325/security-templates-and-xss-prevention
+%%
+%% NOTE: this model differs from 'q' in that it _only_ looks at the query
+%% arguments in the URL (and not at arguments from API calls or other sources).
+-spec m_get( list(), zotonic_model:opt_msg(), z:context() ) -> zotonic_model:return().
+
+% Template syntax: m.qarg.<arg_name>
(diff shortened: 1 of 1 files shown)


apps/zotonic_core/src/models/m_qarg.erl
```diff
@@ -0,0 +1,68 @@
...
+%
+% Get all the escaped values for the given key from the query arguments.
+m_get([ ArgName, <<"all">> | Rest ], _Msg, Context) when is_binary(ArgName) ->
+    Qs = req_qs(Context),
+    All = proplists:get_all_values(ArgName, Qs),
+    Result = lists:map(fun(Val) -> filter_escape:escape(Val, Context) end, All),
+    {ok, {Result, Rest}};
+
+% Template syntax: m.qarg.<arg_name>.id
+% Replacement for: m.rsc[q.<arg_name>].id
```
robvandenbogaard, line 50: I wonder if we should make it `m.qarg[arg_name].id`. Also, just for the record, as we discussed it previously and decided against, though it did pop up in my mind again: Imagining a new dev seeing the interface, I'd probably suspect to have to write something like `m.qarg[arg_name].resource` (or `rsc`) to "cast" it to a resource, and then be able to explicitly ask for an id or another property. So:
- `m.qarg[arg_name].resource.id` or `m.qarg[arg_name].resource.title` (or `.rsc.`)
- versus `m.qarg[arg_name].id` and `m.qarg[arg_name].id.title` (or the `qarg.arg_name.` equivalent)

apps/zotonic_core/src/models/m_qarg.erl
```diff
@@ -0,0 +1,68 @@
...
+%% arguments in the URL (and not at arguments from API calls or other sources).
+-spec m_get( list(), zotonic_model:opt_msg(), z:context() ) -> zotonic_model:return().
+
+% Template syntax: m.qarg.<arg_name>
+% Replacement for: q.<arg_name>|escape
+%
+% Get the escaped value for the given key from the query arguments.
+m_get([ ArgName ], _Msg, Context) when is_binary(ArgName) ->
+    Qs = req_qs(Context),
+    Result = case proplists:get_value(ArgName, Qs, undefined) of
```
mworrell, line 32: Wy not `z_context:get_q/2` ?

mworrell:
Discussed this with @pasqu4le and we are investigating if we could make modifications to the template compiler instead.

For now on hold.

PR #4049, opened 2025-02-26
Fix a problem where a crashed application does not restart Zotonic

mworrell:
### Description

This fixes a problem where if an essential application crashes the Erlang node keeps on running without that application. 

The solution is to make the applications permanent, which on application failure will crash the Erlang node. The node will then be restarted by `heart`.

Sites and modules are started as temporary apps.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/db/z_db_pgsql.erl
+++ b/apps/zotonic_core/src/db/z_db_pgsql.erl
@@ -109,7 +109,7 @@ test_connection(Args) ->
     end.
 
 ensure_all_started() ->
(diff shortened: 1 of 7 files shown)


mmzeeman (approved):
A while ago I made https://github.com/mmzeeman/zotonic_mod_disrupt which uses https://github.com/ankhers/havoc. This randomly kills processes. Maybe we could use it to test if zotonic stays alive.

PR #4063, opened 2025-04-03
mod_search: fix a problem where counting rows could result in a SQL error

mworrell:
### Description

If rows are counted, then the sort terms are removed.
If a SQL argument is only used in the sort term, then too many arguments are supplied and PostgreSQL will complain about the type of the over-supplied argument.

This is now fixed by not supplying extra arguments (the ranking behavior) when counting rows.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


pasqu4le (approved):
Works like a charm 👍

PR #4075, opened 2025-04-17
Exclude 'pivot_geocode_qhash' from rsc raw gets

pasqu4le:
### Description

Problem: `pivot_geocode_qhash` can be any binary, including one that is not valid UTF8, and attempting to export the resource/property (via 'controller_api', for example) causes an error in the encoders, which attempt to interpret it as a binary string.

Solution: As the for most other `pivot` properties, this can be excluded from the rsc get (it's still accessible as a DB column).

@mworrell this is the PR following up our discussion earlier.
It seem to be working fine, but I'm not sure if anything depends on this property being there, so let me know if you think a TODO for later is better instead.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mworrell:
This seems to be ok, as the rsc props for `pivot_fields`  are fetched from the db by the pivot routines, and are not depending on the raw rsc from `m_rsc`.

PR #4084, opened 2025-05-02
Fix vevent issue

DorienD:
This will fix the ics file so importing works again.


mworrell:
Funny, is that because the VEVENT comes at the same line as something else?

mworrell (approved):

PR #4095, opened 2025-05-16
db: Added missing convert for group_by queries containing an array_agg column

mmzeeman:
### Description

There was a missing decode value when the column value is a list. This can happen when you use `ARRAY_AGG` in a group by query. In that case the values in the list is not converted, making it possible to return datetime values which are decoded. (postgres date time values can have float seconds)

### Checklist

- [ ] documentation updated
- [x] tests added
- [x] no BC breaks


mmzeeman:
No longer needed because of #4096

PR #4096, opened 2025-05-18
 Move datetime eterm decode to epgsql codec

mmzeeman:
### Description

In the past postgres did not have null conversion and pluggable codecs. Last week I needed a custom aggregate query which contained datetime values and discovered these where not converted from the postgres datetime (with float seconds) to the datetime as is usual in Erlang. At first I made PR #4095, but realized later that epgsql now supports pluggable codecs. This PR started as a proof of concept to see if it was possible to remove all custom decoding and move it to a codec.

There is one change. The use of `binary_to_term/1` scares me. That conversion was previously tried on all binary values returned from the database. We have a column named `props` which is defined as a `bytea` in which we store erlang terms. This change only tries the conversion on `bytea` type cells returned from the database.

In the future we should consider introducing a new type `eterm` into the schema definition for this special use props field, and make sure only cells from type `eterm` are encoded and decoded, and are not accidentally user values.

### Checklist

- [ ] documentation updated
- [x] tests added
- [x] no BC breaks


apps/zotonic_core/src/db/z_db_pgsql.erl
```diff
@@ -782,6 +771,25 @@ connect_1(Args, RetryCt, MRef) ->
...
+    Opts3 = maybe_put_arg(dbuser, Args, username, Opts2),
+    maybe_put_arg(dbpassword, Args, password, Opts3).
+
+maybe_put_arg(ArgsKey, Args, Key, Map) ->
+    case get_arg(ArgsKey, Args) of
+        undefined ->
+            Map;
+        Value ->
+            maps:put(Key, Value, Map)
+    end.
```
vkatsuba, line 791: **Minor:** Maybe it can be rewritten like:
```suggestion
    maybe_put_args([{dbhost, host}, {dbport, port}, {dbuser, username}, {dbpassword, password}], Args, Opts).

maybe_put_args([], _, Map) ->
    Map;
maybe_put_args([{ArgsKey, Key} | T], Args, Map) ->
    case get_arg(ArgsKey, Args) of
        undefined ->
            maybe_put_args(T, Args, Map);
        Value ->
            maybe_put_args(T, Args, Map#{Key => Value})
    end.
```

apps/zotonic_core/src/db/z_db_pgsql_codec.erl
```diff
@@ -0,0 +1,66 @@
...
+    epgsql_codec_text:decode_text(Cell, bytea, State);
+decode_text(Cell, TypeName, State) ->
+    epgsql_codec_datetime:decode_text(Cell, TypeName, State).
+
+decode_value(<<?TERM_MAGIC_NUMBER, B/binary>>) ->
+    binary_to_term(B);
+decode_value({H,M,S}) when is_float(S) ->
+    {H,M,trunc(S)};
+decode_value({{Y,Mm,D},{H,M,S}}) when is_float(S) ->
+    {{Y,Mm,D},{H,M,trunc(S)}};
```
vkatsuba, line 64: **Minor:**
```suggestion
decode_value({H, M, S}) when is_float(S) ->
    {H, M, trunc(S)};
decode_value({{Y, Mm, D}, {H, M, S}}) when is_float(S) ->
    {{Y, Mm, D}, {H, M, trunc(S)}};
```

apps/zotonic_core/src/db/z_db_pgsql_codec.erl
```diff
@@ -0,0 +1,66 @@
...
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+
+-module(z_db_pgsql_codec).
+-behaviour(epgsql_codec).
+
+-export([init/2, names/0, encode/3, decode/3, decode_text/3]).
+
+-define(TERM_MAGIC_NUMBER, 16#01326A3A:1/big-unsigned-unit:32).
```
vkatsuba, line 31: **Minor:** maybe make sense to leave a comment about `TERM_MAGIC_NUMBER` like or similar:
```suggestion
%% 4-byte magic number used as a prefix to serialized Erlang terms.
%% Helps identify and distinguish term binaries in raw byte streams.
-define(TERM_MAGIC_NUMBER, 16#01326A3A:1/big-unsigned-unit:32).
```

mmzeeman: Good call.

vkatsuba (approved):

PR #4099, opened 2025-05-21
mod_authentication: add extra login options in step 2 of login

mworrell:
### Description

In the second step, if the username is known, show only the options for that username.
Hide the social logins and other extra options.

Also ensure that the error div is always present, so that the HTML-template-patching goes well if it is added.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_authentication/priv/templates/_logon_login_extra.tpl
+++ b/apps/zotonic_mod_authentication/priv/templates/_logon_login_extra.tpl
@@ -1,17 +1,12 @@
 {#
 Add other login options
 #}
-<ul class="z-logon-extra">
-    {% all include "_logon_extra.tpl" %}
-    <li class="text-muted z-logon-extra-separator"><span>{_ or _}</span></li>
-</ul>
-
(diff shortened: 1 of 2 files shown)


apps/zotonic_mod_authentication/priv/templates/_logon_login_extra.tpl
```diff
@@ -1,17 +1,12 @@
...
-#}
+{% if not q.options.is_username_checked %}
+    <ul class="z-logon-extra">
+        {% all include "_logon_extra.tpl" %}
+        <li class="text-muted z-logon-extra-separator"><span>{_ or _}</span></li>
+    </ul>
+{% else %}
+    <ul class="z-logon-extra">
+    </ul>
+{% endif %}
```
vkatsuba, line 12: It looks like that need add new line in the end of file
```suggestion
{% endif %}
```suggestion
{% endif %}
```
```

vkatsuba (approved):

PR #4101, opened 2025-05-26
mod_mailinglist: add confirm option to subscribe action, add text for in emails

mworrell:
### Description

This adds a configuration and action argument to request confirmation of a new subscription by email before subscribing.

Add an editorial text that is inserted at the bottom of the confirm, welcome and unsubscribe emails.

Refactor admin templates, move the email options and the new text to the main content widgets.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@DorienD I've pushed some fixes and textual changes.

DorienD:
I see the text now in the welcome mail, but i don't get a confirmation email. Is that standard or is there an interface that iI need to adjust?

mworrell:
I've pushed a change to SN for the confirm.

PR #4105, opened 2025-05-30
Fix some rough facet search edges.

### LINKED ISSUE #4104: Facet search has some rough edges.
During the implementation of a search feature on a site I ran into some rough edges of the new facet search functionality.

- [x] Syntax errors in `facet.tpl` is not clearly reported, but instead displays a list of attempts to update the `search_facet` table.
- [x] An `xxxx_ids` block definition results in a column of type `integer[]`. This column can be filled with double `id` entries. When this is the case the reported count will be wrong.
- [x] There is no way to query integer array columns because there is no operator for doing postgres `>@`, `<@` or `&&` operations. This makes it impossible to use these facet columns effectively in searches.
- [x] Internal function building queries don't check for array type database columns which will later result in database connection crashes (`z_db:to_column_value/4`)

mmzeeman:
### Description

Fix #4104

Fixes some rough edges in the facet search code.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/db/z_db.erl
+++ b/apps/zotonic_core/src/db/z_db.erl
@@ -1430,9 +1430,16 @@ column_exists(Table, Column, Context) when is_atom(Column) ->
 -spec to_column_value(table_name(), column_name(), term(), z:context()) -> {ok, term()} | {error, term()}.
 to_column_value(Table, Column, Value, Context) ->
     case column(Table, Column, Context) of
-        {ok, #column_def{ type = Type, length = Length }} ->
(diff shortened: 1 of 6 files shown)


mworrell:
Looks good 👍

mmzeeman:
@mworrell Also added a short mention of the new search filters which can be used to used to filter columns based on values in a facets `xxxx_ids` columns.

Would it be an idea to add `keyword_ids` as default block to `facets.tpl` now that it is possible to use it in queries?

PR #4107, opened 2025-06-03
core: after insert of dependent rsc, check if connected

mworrell:
### Description

If a dependent resource is inserted then check after 12 hours if it is connected.
If it is not connected, then delete it.

This prevents the creation of dependent resources that are inserted and then not used.
This could happen if an image is inserted in TinyMCE but then replaced with another (new) image before save.

Also added an extra logger formatter, to make an iolist that was seen on the console more readable.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_rsc_update.erl
+++ b/apps/zotonic_core/src/models/m_rsc_update.erl
@@ -657,7 +657,7 @@ update_result({ok, NewId, {OldProps, NewProps, OldCatList, IsCatInsert}}, #rscup
         },
         Context),
 
-    % If a newly inserted resource is dependent, then schedule a check
+    % If a new or updated resource becomes dependent, then schedule a check
     % if it is connected. If not then it should be deleted.
     case maps:get(<<"is_dependent">>, NewProps, false) of
         true ->


mworrell:
@pasqu4le Do you agree with this change?

apps/zotonic_core/src/support/z_edge_log_server.erl
```diff
@@ -45,6 +47,17 @@
 -record(state, {site :: atom()}).
 
 
+%% @doc Schedule a check if the resource is connected, if not then then the resource will be deleted.
+%% This is called after a resource is inserted. The check is done 12 hours after the resource has
+%% been inserted.
```
pasqu4le: Similarly to [my other comment](https://github.com/zotonic/zotonic/pull/4107#discussion_r2142354880):
```suggestion
%% @doc Schedule a check if the resource is connected in 12 hours, if not then the resource will be deleted.
%% This is called after a resource is inserted or updated.
```

apps/zotonic_core/src/support/z_edge_log_server.erl
```diff
@@ -224,11 +219,24 @@ do_edge_notify(<<"INSERT">>, SubjectId, PredName, ObjectId, EdgeId, Acc, Context
...
             Key = z_convert:to_binary(Id),
-            z_pivot_rsc:insert_task_after(3600, ?MODULE, delete_if_unconnected, Key, [Id], Context);
+            case m_edge:has_subjects(Id, Context) of
+                true -> ok;
+                false ->
+                    case z_pivot_rsc:insert_task_after(Delay, ?MODULE, delete_if_unconnected, Key, [Id], Context) of
+                        {ok, _} -> ok;
+                        {error, _} = Error -> Error
+                    end
+            end;
```
pasqu4le, line 238: Why the additional check on `m_edge:has_subjects` now?

AFAICT when the `delete_if_unconnected` happens this is already accounted for.

apps/zotonic_core/src/models/m_rsc_update.erl
```diff
@@ -657,6 +657,18 @@ update_result({ok, NewId, {OldProps, NewProps, OldCatList, IsCatInsert}}, #rscup
         },
         Context),
 
+    % If a newly inserted resource is dependent, then schedule a check
+    % if it is connected. If not then it should be deleted.
```
pasqu4le: IIUC this would also be triggered for updated resources, if so I'd change the comment:
```suggestion
    % If a new or updated resource becomes dependent, then schedule a check
    % if it is connected. If not then it should be deleted.
```

apps/zotonic_core/src/models/m_rsc_update.erl
```diff
@@ -657,6 +657,18 @@ update_result({ok, NewId, {OldProps, NewProps, OldCatList, IsCatInsert}}, #rscup
...
+    case maps:get(<<"is_dependent">>, NewProps, false) of
+        true ->
+            case maps:get(<<"is_dependent">>, OldProps, false) of
+                true -> ok;
+                false -> z_edge_log_server:maybe_schedule_dependent_check(NewId, Context)
+            end;
+        false ->
+            ok
+    end,
```
pasqu4le, line 670: Why not directly:
```suggestion
    z_edge_log_server:maybe_schedule_dependent_check(NewId, Context),
```
The check of the `is_dependent` value is already done in `maybe_schedule_dependent_check` and if this is called multiple times per resources (in the case of multiple updates), the task is keyed so it's not duplicated.

apps/zotonic_core/src/support/z_edge_log_server.erl
mworrell: Because the edge log server does not check - and by checking if before the `delete_if_unconnected` we can prevent inserting (and running) it as a task if the resource still has referrers.

apps/zotonic_core/src/models/m_rsc_update.erl
mworrell: It is an optimization, only run the check if the `is_dependent` has been changed.

pasqu4le (approved):
LGTM now 👍

PR #4119, opened 2025-06-13
Add defaultLangText as an initial option for users

EdwardAbboud:
### Description
Add variable with a empty default to allow a text placement instead of the empty option.

Prevent user confusion

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_admin/priv/templates/_edit_medium_language.tpl
+++ b/apps/zotonic_mod_admin/priv/templates/_edit_medium_language.tpl
@@ -11,7 +11,7 @@
            }}" title="{_ Help _}"></a>
     </label>
     <select id="{{ #medium_language }}" name="medium_language" class="form-control" style="width: auto">
-        <option value="">{{ defaultLangText|default:_"Select language" }}</option>
+        <option value="">{_ Language independent _}</option>
         {% for code, lang in m.translation.language_list_editable %}
             <option value="{{ code }}" {% if id.medium_language == code %}selected{% endif %}>
                 {{ lang.name }}


apps/zotonic_mod_admin/priv/templates/_edit_medium_language.tpl
```diff
@@ -1,21 +1,23 @@
...
+        <select id="{{ #medium_language }}" name="medium_language" class="form-control" style="width: auto">
+            <option>{{ defaultLangText }}</option>
+            {% for code, lang in m.translation.language_list_editable %}
+                <option value="{{ code }}" {% if id.medium_language == code %}selected{% endif %}>
+                    {{ lang.name }}
+                </option>
+            {% endfor %}
+        </select>
+    </div>
+{% endwith %}
```
pasqu4le: I have two suggestions:
1. I think that the `with` can be avoided, I'm mostly suggesting this to avoid polluting the git history with the indentation changes
2. I think it would be even better for it to have a default value of `Select language` (which is also translated in `.po` files)
```suggestion
<div class="form-group">
    <label class="control-label" for="{{ #medium_language }}">
        {_ Language of media content _}
        <a href="#" class="z-btn-help do_dialog"
           data-dialog="{{
                %{
                    title: _"Language of media content",
                    text: _"Some images, PDFs, audio and video are presented in a single language. Here you can select this language.",
                    level: 10
                }|escape
           }}" title="{_ Help _}"></a>
    </label>
    <select id="{{ #medium_language }}" name="medium_language" class="form-control" style="width: auto">
        <option>{{ defaultLangText|default:_"Select language" }}</option>
        {% for code, lang in m.translation.language_list_editable %}
            <option value="{{ code }}" {% if id.medium_language == code %}selected{% endif %}>
                {{ lang.name }}
            </option>
        {% endfor %}
    </select>
</div>
```

apps/zotonic_mod_admin/priv/templates/_edit_medium_language.tpl
```diff
@@ -11,7 +11,7 @@
            }}" title="{_ Help _}"></a>
     </label>
     <select id="{{ #medium_language }}" name="medium_language" class="form-control" style="width: auto">
-        <option></option>
+        <option>{{ defaultLangText|default:_"Select language" }}</option>
```
mworrell: This needs `value=""`

mworrell:
We have already a label with the explanation, isn't that enough?

apps/zotonic_mod_admin/priv/templates/_edit_medium_language.tpl
EdwardAbboud: Added!

EdwardAbboud:
> We have already a label with the explanation, isn't that enough? And where are you setting this variable?

@mworrell Label is good, but some users are getting confused with the empty dropdown and not actually using it without it having any default text. The variable is getting set on the call of the template (e.g. in the media edit/upload modal)

PR #4131, opened 2025-06-25
Add possibility to schedule regular jobs via mod_cron

mmzeeman:
### Description

Adds the possibility to schedule regular events via mod_cron. This makes it possible to run module specific tasks on regular time intervals. When the module stops, the tasks are automatically cancelled.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_cron/src/mod_cron.erl
+++ b/apps/zotonic_mod_cron/src/mod_cron.erl
@@ -54,7 +54,7 @@
 
 -record(state, {
           jobs = #{},
-          context :: z_context:context()
+          context :: z:context()
          }).
 
 -include_lib("zotonic_core/include/zotonic.hrl").
--- a/apps/zotonic_mod_cron/src/mod_cron.erl
+++ b/apps/zotonic_mod_cron/src/mod_cron.erl
@@ -24,6 +24,21 @@
 -mod_description("Periodic tasks and ticks for other modules.").
 -mod_provides([cron]).
 -mod_prio(500).
+-mod_cron_jobs([
+    {{daily, {every, { 1, sec}}}, {z_notifier, notify, [tick_1s]}},
+    {{daily, {every, { 1, min}}}, {z_notifier, notify, [tick_1m]}},
+    {{daily, {every, { 5, min}}}, {z_notifier, notify, [tick_5m]}},
+    {{daily, {every, {10, min}}}, {z_notifier, notify, [tick_10m]}},
+    {{daily, {every, {15, min}}}, {z_notifier, notify, [tick_15m]}},
+    {{daily, {every, {30, min}}}, {z_notifier, notify, [tick_30m]}},
+    {{daily, {every, { 1, hr}}},  {z_notifier, notify, [tick_1h]}},
+    {{daily, {every, { 2, hr}}},  {z_notifier, notify, [tick_2h]}},
+    {{daily, {every, { 3, hr}}},  {z_notifier, notify, [tick_3h]}},
+    {{daily, {every, { 4, hr}}},  {z_notifier, notify, [tick_4h]}},
+    {{daily, {every, { 6, hr}}},  {z_notifier, notify, [tick_6h]}},
+    {{daily, {every, {12, hr}}},  {z_notifier, notify, [tick_12h]}},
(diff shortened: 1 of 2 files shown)


apps/zotonic_mod_cron/src/mod_cron.erl
```diff
@@ -88,26 +96,67 @@ handle_info(_Info, State) ->
 code_change(_Vsn, State, _Extra) ->
     {ok, State}.
 
-terminate(_Reason, #state{ tick_job_refs = TickJobs }) ->
-    lists:foreach(fun(JobRef) -> _ = erlcron:cancel(JobRef) end, TickJobs),
+terminate(_Reason, #state{ context = Context, jobs = Jobs}) ->
+    z_notifier:detach(module_activate, self(), Context),
+    z_notifier:detach(module_deactivate, self(), Context),
+    maps:foreach(fun(_Module, ModuleJobs) -> _ = cancel_jobs(ModuleJobs) end, Jobs),
```
mworrell, line 102: Do the jobs also get killed if the module process crashes?

And what happens if a job crashes?

apps/zotonic_mod_cron/src/mod_cron.erl
```diff
@@ -37,48 +52,41 @@
...
+    {ok, State}.
 
 handle_call(_Msg, _From, State) ->
     {reply, {error, unknown_message}, State}.
 
+handle_cast({#module_activate{ module=Module }, _Context}, State) ->
+        State1 = add_jobs(Module, State),
+    {noreply, State1};
+handle_cast({#module_deactivate{ module=Module }, _Context}, State) ->
+    State1 = cancel_jobs(Module, State),
```
vkatsuba, line 88: the API `cancel_jobs/2` always return atom `ok`. I can be wrong but should be used real state at `{noreply, State}`. But specifically for this clause keep atom `ok` as state is satisfied.

apps/zotonic_mod_cron/src/mod_cron.erl
```diff
@@ -37,48 +52,41 @@
...
+    {noreply, State1};
```
vkatsuba: Minor:
```suggestion
handle_cast({#module_activate{ module = Module }, _Context}, State) ->
    State1 = add_jobs(Module, State),
    {noreply, State1};
handle_cast({#module_deactivate{ module = Module }, _Context}, State) ->
    State1 = cancel_jobs(Module, State),
    {noreply, State1};
```

doc/ref/modules/mod_cron.rst
```diff
@@ -40,4 +43,38 @@ Check something every hour::
...
+
+When you want to schedule regular jobs, when you want your module to perorm
+jobs at regular intervals it is possible to use `mod_cron` to schedule those
+jobs.
+
+For example, if you want to send out an automated weekly reminder to your
+users you can do this by defining a module specific task in your module.
+
+Example::
+
```
vkatsuba, line 63: Minor: For highlight code block with Erlang syntax can be user ` ```erlang` wrapper, e.g.:
```erlang
-mod_depends([cron]).
-mod_cron_jobs([
    {{weekly, mon, {10, 0, 0}}, {?MODULE, send_weekly_reminders, []}}
]).
```

apps/zotonic_mod_cron/src/mod_cron.erl
mmzeeman: - When the module crashes the module manager will do a module_deactivate notification. This will cancel all its jobs in mod_cron.
- Each job is a process supervised by erlcron. When it crashes a message will be logged and the job will be restarted.

I added extra error handling for when the module has made a wrong job clause. For example when a non existing function is configured. If that happens a message will be logged.

apps/zotonic_mod_cron/src/mod_cron.erl
mmzeeman: Thanks... how do you make these easy to merge suggestions?

apps/zotonic_mod_cron/src/mod_cron.erl
mworrell: And if `mod_cron` crashes/stops?  What does the cron app do?

mworrell (approved):

PR #4135, opened 2025-07-01
core: let controller_template use a 308 when redirecting to language-specific URLs

mworrell:
### Description

This fixes an issue where controller_template used a 303 and omitted cache/vary headers if it was redirecting from (say) `/` to `/en/`

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


pasqu4le (approved):

mworrell:
@pasqu4le Thanks for reviewing!

PR #4140, opened 2025-07-06
mod_translation: make it possible to preview a language

mworrell:
### Description

If the user is allowed to use mod_admin, then make it possible to select the _editable_ languages to preview the site in that language.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


DorienD:
I like this! Easy to understand for users and you immediately see the result of your content.

PR #4142, opened 2025-07-08
core: add dispatch model to fetch dispatch information for paths and urls

mworrell:
### Description

Request the dispatch model how paths and/or URLs are dispatched. Return the controller, controller options, dispatch rule and bindings.

Useful for Javascript that want to handle requests and use the server side dispatch rules.

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_dispatch.erl
+++ b/apps/zotonic_core/src/models/m_dispatch.erl
@@ -37,12 +37,12 @@ m_get([ <<"path">> ], #{ payload := #{ <<"path">> := Path } }, Context) ->
     dispatch_path(Path, Context);
 m_get([ <<"path">> | Path ], _Msg, Context) ->
     dispatch_path(Path, Context);
-m_get([ <<"url_for">>, Name | Rest ], #{ payload := Payload }, Context) when is_map(Payload) ->
+m_get([ <<"url_for">>, Name | Rest ], #{ payload := Payload }, Context) ->
     case url_for(Name, Payload, Context) of
         undefined -> {error, enoent};
         Url -> {ok, {Url, Rest}}
     end;
-m_get([ <<"abs_url_for">>, Name | Rest ], #{ payload := Payload }, Context) when is_map(Payload) ->
+m_get([ <<"abs_url_for">>, Name | Rest ], #{ payload := Payload }, Context) ->
     case url_for(Name, Payload, Context) of
         undefined -> {error, enoent};
         Url -> {ok, {z_context:abs_url(Url, Context), Rest}}
@@ -53,6 +53,8 @@ m_get(_, _Msg, _Context) ->
 url_for(Name, undefined, Context) ->
     z_dispatcher:url_for(Name, Context);
 url_for(Name, Args, Context) when is_map(Args) ->
+    url_for(Name, maps:to_list(Args), Context);
+url_for(Name, Args, Context) when is_list(Args) ->
     ArgsList = lists:map(
         fun
             ({K, V} = KV) when is_binary(K) ->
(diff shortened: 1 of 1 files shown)


doc/ref/notifications/notification/meta-dropbox_file.rst
```diff
@@ -4,7 +4,9 @@ dropbox_file
 ^^^^^^^^^^^^
 
 Handle a new file received in the 'files/dropbox' folder of a site. 
-Unhandled files are deleted after a hour. 
+Unhandled files are deleted after an hour. If the handler returns 'ok' then 
+the file is moved from the files/processing folder to files/handled. 
+folder.
```
pasqu4le, line 9: This seems an unrelated change, just checking, maybe it wasn't intentional?

apps/zotonic_core/src/models/m_dispatch.erl
```diff
@@ -0,0 +1,109 @@
...
+    dispatch_url/2,
+    dispatch_path/2
+]).
+
+m_get([ <<"url">> ], #{ payload := #{ <<"url">> := Url } }, Context) ->
+    dispatch_url(Url, Context);
+m_get([ <<"url">> ], _Msg, Context) ->
+    dispatch_url(z_context:get_q(<<"url">>, Context), Context);
+m_get([ <<"url">>, Url ], _Msg, Context) ->
+    dispatch_url(Url, Context);
```
pasqu4le, line 35: I'm confused about this one, how could we pass a URL to the model as a single path section? 🤔 

I suppose we could do something like for `dispatch_path`:
```suggestion
m_get([ <<"url">> | Url ], _Msg, Context) ->
    io:format("debug url: ~p~n", [Url]),
    dispatch_url(Url, Context);
```
and rebuild the URL in `dispatch_url`, but `z_sites_dispatcher:dispatch_url` requires the URL to be complete, including scheme and host, so I think it would be messy to use, e.g.
```
    curl -k 'https://example.test:8443/api/model/dispatch/get/url/https://example.test:8443/nl/page/1234/hello'
```

Unless I'm missing something, I'd suggest to remove this:
```suggestion
```

apps/zotonic_core/src/models/m_dispatch.erl
```diff
@@ -0,0 +1,109 @@
...
+m_get([ <<"path">> ], #{ payload := #{ <<"path">> := Path } }, Context) ->
+    dispatch_path(Path, Context);
+m_get([ <<"path">> ], _Msg, Context) ->
+    dispatch_path(z_context:get_q(<<"path">>, Context), Context);
+m_get([ <<"path">> | Path ], _Msg, Context) ->
+    dispatch_path(Path, Context);
+m_get(_, _Msg, _Context) ->
+    {error, enoent}.
```
pasqu4le, line 51: What do you think of adding another path here to instead access the result of `z_dispatcher:url_for`?

I think that would also be useful for external services/apps.

doc/ref/notifications/notification/meta-dropbox_file.rst
mworrell: Yes, it just tags along.

apps/zotonic_core/src/models/m_dispatch.erl
mworrell: You can pass it at once if you url encode it or use it as an argument in a template.  In that way it is a shorthand:

```django
m.dispatch.url:%{ url: 'http://...' }
```

versus

```django
m.dispatch.url['http://...']
```

apps/zotonic_core/src/models/m_dispatch.erl
mworrell: To generate URLs?  That is a good idea.

apps/zotonic_core/src/models/m_dispatch.erl
mworrell: Passing URLs in a URL is messy anyways, even when passing in the query (search) arguments.

mworrell:
@pasqu4le Added the `url_for` method to the model.

PR #4144, opened 2025-07-14
Adjust MQTT admin subscription permissions to $SYS

CyBeRoni:
Fix the wrong site name being checked for admin privileges when subscribing to `$SYS/site/<some site>` and allow the zotonic_site_status admin to subscribe to all `$SYS` topics

### Description

Fixes the logic for subscribing over MQTT to  `$SYS/site/<sitename>` for the zotonic admin user. Additionally, allows that admin user to subscribe to all `$SYS` topics.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@mmzeeman Do you agree with giving the status site subscribe permission on all `$SYS/#` topics?  As we already allowed the status site to subscribe on all `$SYS/site/#` topics I think it doesn't act any security issues.

mworrell (approved):

mmzeeman:
No, me neither... 👍

PR #4145, opened 2025-07-14
Fix MQTT auth as admin to zotonic status site

CyBeRoni:
### Description

When trying to use MQTT to authenticate as admin using the credentials from the zotonic_site_status site, authentication fails due to the IP allowlist. The allowlist checks for `peer` in the incoming message but only `peer_ip` is available in that case. Since `peer_ip` is also available when logging in using a web browser, this PR changes the logic to use that instead.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Thanks!

PR #4149, opened 2025-07-21
ACL: add support for visibility levels in rules

pasqu4le:
### Description

This PR modifies the `zotonic_mod_acl_user_groups` module to support visibility in rules, which is used to allow/restrict permissions on resources depending on the value of their `visible_for` property.

For example, let's imagine that one has a platform for moderated publishing of `article`s, where the desired outcome is that:
1. there are many teams (`acl_collaboration_group`s) where people collaborate on writing
2. articles can have 3 different states (`visible_for` values): "draft", "preview" or "shared"
3. any registered user can write (`insert`) new "draft" articles in teams (`acl_collaboration_group`s) that they are a member of
4. users can edit their own articles, but only if they are still "draft"s
5. users can add (`link`) comments to any "draft" article in their teams, regardless of who wrote it
6. editors can delete and edit any article in any team, including changing its state to "preview" or "shared"
7. any registered user can see (`view`) "preview" articles on the platform, regardless of team
8. anyone (including visitors) can `view` "shared" articles on the platform, regardless of team
9. articles remain in their team for their entire life-cycle

Then this could be represented with these rules:
image
image

_Note: in the screenshots the site replaces the numbers for the visibility with text, which can be done with template overriding._

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_mod_acl_user_groups/priv/templates/_admin_acl_rule_row_collab.tpl
+++ b/apps/zotonic_mod_acl_user_groups/priv/templates/_admin_acl_rule_row_collab.tpl
@@ -8,3 +8,10 @@
         <em>{_ All Categories _}</em>
     {% endif %}
 </td>
+<td>
+    {% if rule.visibility %}
+        {{ rule.visibility }}
+    {% else %}
+        <em>{_ Any visibility _}</em>
+    {% endif %}
+</td>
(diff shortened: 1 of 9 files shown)


pasqu4le:
@mworrell done 👍

mworrell:
@pasqu4le There seems to be some error in the parameter counting, could you check (might be a +/- 1 problem).

https://github.com/zotonic/zotonic/actions/runs/16420778967/job/48162245638?pr=4149#step:6:292

pasqu4le:
@mworrell I've changed the UI so that it shows both numbers and text for two visibility levels: public and private.

Not sure if it's worth adding more, as I think their meaning will depend heavily on individual website's logic, but let me know if I should add a couple more example values.

mworrell (approved):

PR #4150, opened 2025-07-23
smtp: added support to pass tls_options for the relay server.

abensoft-alvaro:
### Description

We use AWS SES as a relay server for Zotonic, since Erlang 26 the TLS connection could not be estabilished because of some certificate errors. The situation was improved with the use of `tls_certificate_check`, but we were still getting errors while trying to connect. Basically we need to pass the parameter `{middlebox_comp_mode, false}` as tls_option so that the connection can be established.

Instead of hardcoding this parameter, I've enabled a configuration option to pass any tls_option for the relay server.

### Checklist

- [X] documentation updated
- [ ] tests added
- [X] no BC breaks


mworrell (approved):

mworrell:
These options gets more and more esoteric, thanks!

PR #4151, opened 2025-07-29
Add ghostscript dependency & default dates

Amber-Cloud:
### Ghostscript dependency and default end date 

Added ghostscript dependency to shell.nix (for ImageMagick)
Added default end date == start date for date ranges


robvandenbogaard (commented):
Should we be able to see in the admin interface that a resource has no end date? I think there are cases where it is important to know.

Amber-Cloud:
@robvandenbogaard I reverted the date change, if adding the dep is ok, we just need to make sure we squash when we merge

PR #4153, opened 2025-08-08
Support languages with 3-letters language codes (0.x)

pasqu4le:
### Description

Fixes support for languages with 3-letters language codes (Zotonic 0.x)

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/modules/mod_translation/mod_translation.erl
+++ b/modules/mod_translation/mod_translation.erl
@@ -472,7 +472,7 @@ language_add(OldIsoCode, NewIsoCode, Language, FallbackIsoCode, IsEnabled, Conte
     set_language_config(Languages3, Context).
 
 -define(is_09(C), ((C) >= $0 andalso (C) =< $9)).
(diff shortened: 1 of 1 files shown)


mworrell:
If we start supporting (certain) iso639-2 language codes then we should also add that support to master. Otherwise we create incompatibilities.

Also be sure that we don't (accidentally) create conflicts between three letter path names and language codes.

Afaik there is code at places to recognize two letter language prefixes (with optional sub-language separated by a dash), that code might need to be checked as well.

CyBeRoni:
> @mworrell I've removed the other versions from the CI build, but it now seems to be failing on OTP 22.3 too 😕
> 
> From the runs' logs it seems that the issue is with the debian repos being unreachable(404), maybe it's a temporary issue?

No, this is quite permanent. Because buster (Debian 10) is now more than 3 releases old, it has been removed from the general package archives and it must now be used from https://archive.debian.org/. This requires changing the sources.list files.

PR #4155, opened 2025-08-14
Add ppo and ppu for papiamento and pipiamentu as language options

loetie:
### Description

Add papiamento/papiamentu options for sites that need to support both languages and therefor can't use pap

### DIFF
--- a/src/i18n/z_trans.erl
+++ b/src/i18n/z_trans.erl
@@ -254,6 +254,8 @@ default_language(Context) ->
 
 %% @doc check if the two/three letter code is a valid language
 -spec is_language(Language :: string() | binary()) -> boolean().
+is_language("ppo") -> true; %papiamento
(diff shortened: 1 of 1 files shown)


src/i18n/z_trans.erl
```diff
@@ -254,6 +254,8 @@ default_language(Context) ->
 
 %% @doc check if the two/three letter code is a valid language
 -spec is_language(Language :: string() | binary()) -> boolean().
+is_language("ppo") -> true; %papiamento
+is_language("ppu") -> true; %papiamentu
```
pasqu4le: I think that for this to work correctly we also need to add the same clauses for binary strings:
```suggestion
is_language("ppo") -> true; %papiamento
is_language(<<"ppo">>) -> true; %papiamento
is_language("ppu") -> true; %papiamentu
is_language(<<"ppu">>) -> true; %papiamentu
```

PR #4169, opened 2025-08-21
core: admin user can login even if unpublished

### LINKED ISSUE #4152: Unpublished admin user (id 1) means you can't login as admin
Sometimes people unpublish the admin user. I do not know why, but I assume it's because they think it should not be visible on the site and think it is "published" by mistake. In any case, this of course causes problems and I think it is reasonable to special-case it and prevent this problem.

## `Zotonic` version
All

## `OS` / `Browser` version
Any, Any

## Steps to reproduce
Unpublish the main admin user (id 1), try to log back in using it

## Current behavior
Login impossible, "user disabled"

## Expected behavior
Either:
 - Published state is ignored for admin user, or
 - Unpublishing admin user is not possible

mworrell:
### Description

Fixes #4152

Fix an issue where the admin user can not log on if the admin user's rsc (id 1) is unpublished.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_auth.erl
+++ b/apps/zotonic_core/src/support/z_auth.erl
@@ -229,6 +229,9 @@ unpublish_user_session(Context) ->
 
 %% @doc Check if the user is enabled, a user is enabled when the rsc is published and within its publication date range.
 -spec is_enabled( m_rsc:resource_id(), z:context() ) -> boolean().
+is_enabled(?ACL_ADMIN_USER_ID, _Context) ->
+    % The admin user is always enabled.
+    true;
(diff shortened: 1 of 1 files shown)


mmzeeman (approved):

mworrell:
Apparently you got lock out the admin - as Driebit found out on one of their sites.

CyBeRoni:
> I always thought it was a feature where you could disable the admin login.

That's fair, but that should not be a thing that is able to be done by any random (non-admin!) editor. So imo this should be a separate mechanism from simply unpublishing it.

PR #4175, opened 2025-08-26
mod_acl_user_groups: fix an issue where inserting unknown rsc id creates wildcards

### LINKED ISSUE #4162: ACL rule for not existing category results into all category acl rule
## `Zotonic` version
1.x and 0.x

## Steps to reproduce
Create an acl rule in schema for a resource with a category that does not exist.

## Current behavior
This creates an ACL rules that applies to all categories if the specified category does not exist. 

Rules are deleted when category is deleted (this seems to work fine). But if the schema is run again after a category is deleted the rule is added again.

## Expected behavior
Would expect the creating of the ACL rule to not happen if the category does not exist and log an error when this happens. 

Since this can create a situation that user groups get too much access if a category is removed and the schema is run.

mworrell:
### Description

Fix #4162

When inserting or updating an ACL rule, check if the referenced resources exist. If not then emit an error.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_acl_user_groups/src/models/m_acl_rule.erl
+++ b/apps/zotonic_mod_acl_user_groups/src/models/m_acl_rule.erl
@@ -230,19 +230,37 @@ actions(module, Context) ->
 
 
 update(Kind, Id, Props, Context) ->
-    ?LOG_DEBUG(
-        "ACL user groups update by ~p of ~p:~p with ~p",
-       [z_acl:user(Context), Kind, Id, Props]
-    ),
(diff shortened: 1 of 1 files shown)


mworrell:
@pasqu4le Do you agree with this change?

mworrell:
@pasqu4le Made the change, and a couple of other small cleanups wrt the usage of binaries for get_q calls.

pasqu4le (approved):

PR #4176, opened 2025-08-26
mod_ssl_letsencrypt: higher timeout on LE requests

mworrell:
### Description

The timeout of 15 seconds on LE requests is sometimes too short. Up it to 60 seconds, also make it a `-define` for easier adjustment in the future.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Good to know this sometimes takes a little bit longer.

PR #4181, opened 2025-08-29
mod_export: editable export columns with value expressions

mworrell:
### Description

Adds admin option to edit the fields to be exported.
These fields can be expressions, enabling model and filter data to be exported.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Very cool..

This probably needs some examples on how to use the template and expressions to export values.

PR #4186, opened 2025-09-03
Accept country and region codes for iso639-2 three letter languages

mworrell:
### Description

Also accept language codes like `pap-aw`

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
@loetie This one is needed for usage of `pap-cw` and `pap-aw` codes.

mworrell:
@loetie could you test with `pap-cw` ?

PR #4190, opened 2025-09-05
new: allow containerization in Podman

Pakti:
### Description

Allow containerization in Podman:

- add script `./start-podman.sh` to start Zotonic in a Podman container
- convert relative paths of Docker images to absolute (as required by Podman)
- grant root privileges to the `zotonic` user when running in Podman  -- Docker does that by default
- update documentation

### Checklist

- [X] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Installation
 ------------
 
 * Download Zotonic from the [**official website**](https://zotonic.com/download)
-  or start Zotonic from a [**Docker image**](https://zotonic.com/docs/1411/docker).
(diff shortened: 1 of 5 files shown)


Pakti:
I've tried to update the documentation despite not knowing how ReStructuredText/Sphinx works, but I couldn't verify it because make html can't find a directory. I assume it's meant to run as part of a full build, but building Zotonic locally seems to interfere with the container, so I skipped that.

mworrell:
Thank you, and don't worry about the docs, we are reworking the docs anyway and it looks good to me!

PR #4202, opened 2025-09-12
Fix cols_map after enabling json decoding at db driver level

mmzeeman:
### Description

Fix a problem where a props_json column was not handled correctly because the cell value currently is already decoded.

- Remove the jsxrecord decode
- Refactored the function so it is more clear what it does.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks


mworrell:
Let's keep the current behavior, for the binary/atom mix when decoding values.  Changing it might need some more inspection to how the JSON decoding is actually done, how to pass the "keys: atom" option, and decisions about what to do with nested keys in the JSON.  That all might give rise to atom-leaks when non-restricted JSONs are decoded.

mworrell (approved):

Issue #4194, opened 2025-09-08
Translated url for resource not showing the correct language


loetie:
## `Zotonic` version
https://github.com/zotonic/zotonic/releases/tag/0.90.0

## `OS` / `Browser` version
All

## Steps to reproduce
1. Use the translated page slugs for example English and Dutch
Assuming English is the default language. 

English url "/patato"
Dutch url "/aardappel"

2. View url directly for dutch (eg; www.website.com/aardappel) you get redirected to the english page

## Current behavior
The translated urls dont go to the corresponding language associated with the but the default language. 
It is possible to directly view it but that url will be www.website.com/nl/aardappel. 

## Expected behavior
The end user expects that the filled in url for language you get to the dutch version of the page if you use the above example. Also because the explanation in the admin below it shows it replaces /nl/page/number/slug 
It can be tricky same url is used for multiple language, then it would be a good idea to default to the default language. But if they are unique for the page then that makes no sense to do.

mworrell:
Do I understand that you set the config to _not_ include the language in the URL?

mworrell:
See #4203 for a possible implementation of the requested behavior for Zotonic 1.x.  @loetie Is this what you requested?

mworrell:
Closed with #4203

mworrell:
Ping @loetie

mworrell:
@loetie ping

PR #4203, opened 2025-09-12
mod_base: use translated page-path to set request language

mworrell:
### Description

Issue #4194

if a page is requested using a custom page path without a language, then this will look at the page-paths of the resource to set the language of the request.

Example:

Assume a resource with the page paths:
- `/over-ons` for `nl`
- `/about-us` for `en`

For the request of `/over-ons` the user will be redirected to `/nl/over-ons`.
For the request of `/about-us` the user will be redirected to `/en/about-us`

For the request of `/en/over-ons` the user will be redirected to `/en/about-us`.
For the request of `/nl/about-us` the user will be redirected to `/nl/over-ons`

If the language is in the url (`/en/...`) then that language will be preferred.

/cc @loetie Can you test the behavior for your use cases?  If it works well then we can copy a version of it to the 0.x for your customer.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_base/src/mod_base.erl
+++ b/apps/zotonic_mod_base/src/mod_base.erl
@@ -389,7 +389,7 @@ maybe_page_path_language(Id, RequestedPagePath, Context) ->
                     {ok, Id}
             end;
         _OtherPath ->
-            % Assume language was in the URL andthere was a rewrite for the language
+            % Assume language was in the URL and there was a rewrite for the language
             {ok, Id}
     end.
 


apps/zotonic_mod_base/src/mod_base.erl
```diff
@@ -353,6 +354,46 @@ observe_dispatch(#dispatch{path=Path}, Context) ->
...
+                                    end
+                            end;
+                        _ ->
+                            {ok, Id}
+                    end;
+                _QLang ->
+                    {ok, Id}
+            end;
+        _OtherPath ->
+            % Assume language was in the URL andthere was a rewrite for the language
```
pasqu4le: Nitpick:
```suggestion
            % Assume language was in the URL and there was a rewrite for the language
```

PR #4207, opened 2025-09-18
docs: move docs to erlang moduledoc and markdown files

mworrell:
### Description

This is a **major** change to the documentation.

Main purpose is to have the reference documentation as `-moduledoc` in the corresponding erlang files.

Changes:

 * Move reference documentation to Erlang files
 * Move all release notes to markdown files in doc/release-notes
 * Move all template tag documentation to doc/template-tags
 * Delete all other documentation.

The other documentation will be maintained on zotonic.com, in the CMS.

The zotonic.com site will automatically import the moduledoc, release notes and template tag documentation into its resources.

### Checklist

- [x] documentation updated
- [ ] tests added
- [ ] no BC breaks


mmzeeman:
🤯 Files changed (2061)

mmzeeman (approved):
Nice too see the number of requirements for building.

mworrell:
@vkatsuba Was is your opinion of moving the docs to the `-moduledoc` module attributes?

mworrell:
I would also like to add keywords to the files, am still uncertain if we should add them in the files (where we have the code) or add them inside Zotonic (were we have the formal definitions of all subjects).

apps/zotonic_core/src/models/m_acl.erl
```diff
@@ -18,6 +18,65 @@
...
+-moduledoc("
+The m\\_acl model gives access the id of the currently logged in user, and provides a mechanism to do basic access
+control checks.
+
+The following m\\_acl model properties are available in templates:
+
+| Property                                       | Description                                                                      |
+| ---------------------------------------------- | -------------------------------------------------------------------------------- |
+| user                                           | Returns the current user id. If not logged in, this returns `undefined`.         |
+| is\\\\_admin                                     | Check if the current user is alllowed to access the admin. Internally, this checks the `use, mod_admin_config` ACL. |
```
robvandenbogaard, line 30: In plain text view this table row looks just a bit unaligned, due to the double slashes; is the goal to have it look good both in raw view and rendered?

robvandenbogaard (approved):
Looks good, a big improvement in terms of being able to keep the documentation in sync (helped with the tooling for back and forth syncing of the documentation site). In some of the current texts I thought maybe we would like to add backtick quoting of inline literal code, but we can make those fixes more easily in iteration from now. 🙏
Maybe I'd be for including the user-guide too, but indeed, maybe tricky with media in there. I'm not totally convinced that it would be a bad idea, but we can always address that in a later iteration.

apps/zotonic_core/src/models/m_acl.erl
mworrell: Just when rendered - extra backslashes were added when we made it into an Erlang string.

We can't use triple quoted strings yet, as they are only added in OTP-27 and we support OTP-25+ at the moment.

mworrell: Maybe we can fix backslash problems when we can see the generated HTML on the site. In this case the properties might actually better with backticks code. Like `is_admin`

PR #4211, opened 2025-09-23
Full text search: use websearch_to_tsquery and normalize texts

mworrell:
### Description

Changes to full text and full text indexing:

- Normalize all indexed texts using `z_string:normalize/1`.  This transliterates most accented characters, Cyrillic, Hebrew and Arabic to ascii.
- Normalize search questions using `z_string:normalize/1`
- Use the PostgreSQL function `websearch_to_tsquery` for mapping a search text to a tsquery text. This function has been available since PostgreSQL 11.

Questions:

1. Should we "normalize" the indexed address and name information?
2. Do we want to transliterate Cyrillic, Arabic and Hebrew?  If not then we can use the "unaccent" function instead of "normalize".

### Copilot summary

This pull request introduces a new, extensible normalization mechanism for text values used in search queries and indexing. It adds a `normalize_value/4` function to the `z_search` module, which uses a new notification (`search_query_normalize_value`) to allow custom normalization logic. The PR updates various parts of the codebase to use this mechanism, ensuring consistent and configurable normalization of text across search and pivoting operations. Additionally, it improves the handling of UTF-8 and text case in several places.

**Core search normalization improvements:**

* Introduced the `normalize_value/4` function in `z_search.erl`, which normalizes text values for search and indexing by filtering UTF-8, removing diacritics, converting to lowercase, and supporting custom normalization via notifications. [[1]](diffhunk://#diff-40db7e87e8af0c17db8766be9742355fb25b9b86c3ad9a0e8d19af6dcd5b9d8fL46-R48) [[2]](diffhunk://#diff-40db7e87e8af0c17db8766be9742355fb25b9b86c3ad9a0e8d19af6dcd5b9d8fR245-R271)
* Defined the `search_query_normalize_value` notification record and callbacks, allowing modules to override or extend normalization logic for search terms. [[1]](diffhunk://#diff-17afdb86efbfaa27b2a60add0e04ba450cb7f25d8fee51786d32a696cadbe16aR879-R887) [[2]](diffhunk://#diff-341796a97940b4a8495809ccb972fef405bd79391c80c57afc936e4a35090f32R2998-R3035)

**Pivot resource and search pipeline updates:**

* Updated `z_pivot_rsc_job.erl` to use `z_search:normalize_value/4` for normalizing pivot titles and TSV text, and ensured that address/name fields are consistently lowercased. Improved title extraction to prefer English or the first available translation, and enhanced text cleanup and truncation logic. [[1]](diffhunk://#diff-66f5e4730ccaa9ae8f36bdcc5dbc7bc8f0d475925fe4eabb70758ee301bc2793L73-R74) [[2]](diffhunk://#diff-66f5e4730ccaa9ae8f36bdcc5dbc7bc8f0d475925fe4eabb70758ee301bc2793R232-R240) [[3]](diffhunk://#diff-66f5e4730ccaa9ae8f36bdcc5dbc7bc8f0d475925fe4eabb70758ee301bc2793L247-R262) [[4]](diffhunk://#diff-66f5e4730ccaa9ae8f36bdcc5dbc7bc8f0d475925fe4eabb70758ee301bc2793L395-R405) [[5]](diffhunk://#diff-66f5e4730ccaa9ae8f36bdcc5dbc7bc8f0d475925fe4eabb70758ee301bc2793R424-R437) [[6]](diffhunk://#diff-66f5e4730ccaa9ae8f36bdcc5dbc7bc8f0d475925fe4eabb70758ee301bc2793L475-R500)
* Modified `mod_search.erl` and `search_facet.erl` to use the new normalization function for query and facet processing, and improved handling of list and binary text inputs for UTF-8 safety. [[1]](diffhunk://#diff-3e0f8ef375ec36ea3e62da06391e5f1ad4c7738216c8c9c150f58482541fbf53R800-R826) [[2]](diffhunk://#diff-3e0f8ef375ec36ea3e62da06391e5f1ad4c7738216c8c9c150f58482541fbf53L831-R862) [[3]](diffhunk://#diff-3e0f8ef375ec36ea3e62da06391e5f1ad4c7738216c8c9c150f58482541fbf53L851-R880) [[4]](diffhunk://#diff-346b033c6dff93111616669b0ee8e44eb97f250df9a56008b1893ed1826fb5c4L599-R599) [[5]](diffhunk://#diff-346b033c6dff93111616669b0ee8e44eb97f250df9a56008b1893ed1826fb5c4L615-R615) [[6]](diffhunk://#diff-346b033c6dff93111616669b0ee8e44eb97f250df9a56008b1893ed1826fb5c4L731-R730) [[7]](diffhunk://#diff-346b033c6dff93111616669b0ee8e44eb97f250df9a56008b1893ed1826fb5c4L743-R742)

**General improvements and copyright:**

* Updated copyright years to 2026 in several files. [[1]](diffhunk://#diff-341796a97940b4a8495809ccb972fef405bd79391c80c57afc936e4a35090f32L2-R7) [[2]](diffhunk://#diff-66f5e4730ccaa9ae8f36bdcc5dbc7bc8f0d475925fe4eabb70758ee301bc2793L2-R6) [[3]](diffhunk://#diff-40db7e87e8af0c17db8766be9742355fb25b9b86c3ad9a0e8d19af6dcd5b9d8fL2-R6)

These changes make search normalization more robust, extensible, and consistent across the system, improving both search accuracy and maintainability.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_pivot_rsc_job.erl
+++ b/apps/zotonic_core/src/support/z_pivot_rsc_job.erl
@@ -244,19 +244,19 @@ pivot_resource_1(Id, Lang, Context) ->
                     KVs = #{
                         <<"pivot_tsv">> => Tsv,
                         <<"pivot_rtsv">> =>  Rtsv,
-                        <<"pivot_street">> => truncate(Street, 120),
-                        <<"pivot_city">> => truncate(City, 100),
-                        <<"pivot_postcode">> => truncate(Postcode, 30),
-                        <<"pivot_state">> => truncate(State, 50),
-                        <<"pivot_country">> => truncate(Country, 80),
-                        <<"pivot_first_name">> => truncate(NameFirst, 100),
-                        <<"pivot_surname">> => truncate(NameSurname, 100),
-                        <<"pivot_gender">> => truncate(Gender, 1),
+                        <<"pivot_street">> => z_string:to_lower(truncate(Street, 120)),
+                        <<"pivot_city">> => z_string:to_lower(truncate(City, 100)),
+                        <<"pivot_postcode">> => z_string:to_lower(truncate(Postcode, 30)),
+                        <<"pivot_state">> => z_string:to_lower(truncate(State, 50)),
+                        <<"pivot_country">> => z_string:to_lower(truncate(Country, 80)),
+                        <<"pivot_first_name">> => z_string:to_lower(truncate(NameFirst, 100)),
+                        <<"pivot_surname">> => z_string:to_lower(truncate(NameSurname, 100)),
+                        <<"pivot_gender">> => z_string:to_lower(truncate(Gender, 1)),
                         <<"pivot_date_start">> => DateStart,
                         <<"pivot_date_end">> => DateEnd,
                         <<"pivot_date_start_month_day">> => DateStartMonthDay,
                         <<"pivot_date_end_month_day">> => DateEndMonthDay,
-                        <<"pivot_title">> => truncate(Title, 100),
+                        <<"pivot_title">> => truncate(z_string:normalize(Title), 100),
                         <<"pivot_location_lat">> => LocationLat,
                         <<"pivot_location_lng">> => LocationLng
                     },
@@ -398,7 +398,8 @@ to_tsv(Text, Level, Args, StemmingLanguage) when is_binary(Text) ->
             {"tsvector('')", Args};
         TsvText ->
             N = length(Args) + 1,
-            Truncated = z_string:truncatechars(TsvText, ?MAX_TSV_LEN, <<>>),
+            TsvText1 = z_string:normalize(TsvText),
+            Truncated = z_string:truncate(TsvText1, ?MAX_TSV_LEN, <<>>),
             Args1 = Args ++ [Truncated],
             {["setweight(to_tsvector('pg_catalog.",StemmingLanguage,"', $",integer_to_list(N),"), '",Level,"')"], Args1}
     end.
@@ -417,17 +418,20 @@ to_integer(Text) ->
         Text1 -> z_convert:to_integer(Text1)
     end.
 
+%% @doc Preprocess a text before making in into a TSV. Ensure the text is valid utf8, replace newlines
+%% and tabs with a space, and trim the resuling text.
 -spec cleanup_tsv_text(binary()) -> binary().
 cleanup_tsv_text(Text) when is_binary(Text) ->
     Text1 = z_string:sanitize_utf8(Text),
-    Text2 = iolist_to_binary(re:replace(Text1, <<"[ ",13,10,9,"/-]+">>, <<" ">>, [global])),
+    Text2 = iolist_to_binary(re:replace(Text1, <<"[ ",13,10,9,"/–]+">>, <<" ">>, [global])),
     z_string:trim(Text2).
 
+%% @doc Truncate a string to a max length, map control characters to spaces
+%% and remove non-utf8 codepoints.
 -spec truncate(binary(), integer()) -> binary().
 truncate(S, Len) ->
     S1 = z_string:trim(S),
-    S2 = truncate_1(S1, Len, <<>>),
-    z_string:trim( z_string:to_lower(S2) ).
+    z_string:trim(truncate_1(S1, Len, <<>>)).
 
 truncate_1(_S, 0, Acc) ->
     Acc;
@@ -472,20 +476,25 @@ get_pivot_title(Id, Context) ->
     Title = m_rsc:p(Id, <<"title">>, z_language:default_language(Context), Context),
     Lang = z_language:default_language(Context),
     Title1 = z_trans:lookup_fallback(Title, Lang, Context),
-    z_string:trim(z_string:to_lower(Title1)).
+    z_string:trim(z_string:normalize(z_convert:to_binary(Title1))).
 
-%% @doc Fetch the first title from the record for sorting.
+%% @doc Fetch the English or first title from the record for sorting.
 -spec get_pivot_title(map()) -> binary().
 get_pivot_title(Props) ->
     Title = case maps:get(<<"title">>, Props, <<>>) of
         #trans{ tr = [] } ->
             <<>>;
-        #trans{ tr = [{_, Text}|_] } ->
-            Text;
-        T ->
-            T
+        #trans{ tr = [ {_,First} | _ ] = Tr } ->
+            case proplists:get_value(en, Tr) of
+                undefined -> First;
+                En -> En
+            end;
+        T when is_binary(T) ->
+            T;
+        _ ->
+            <<>>
     end,
-    z_string:trim(z_string:to_lower(Title)).
+    z_string:trim(z_string:normalize(Title)).
 
 
 %% @doc Return the raw resource data for the pivoter
--- a/apps/zotonic_core/src/support/z_pivot_rsc_job.erl
+++ b/apps/zotonic_core/src/support/z_pivot_rsc_job.erl
@@ -423,7 +423,7 @@ to_integer(Text) ->
 -spec cleanup_tsv_text(binary()) -> binary().
 cleanup_tsv_text(Text) when is_binary(Text) ->
     Text1 = z_string:sanitize_utf8(Text),
-    Text2 = iolist_to_binary(re:replace(Text1, <<"[ ",13,10,9,"/–]+">>, <<" ">>, [global])),
+    Text2 = iolist_to_binary(re:replace(Text1, <<"[ \r\n\t/–]+"/utf8>>, <<" ">>, [global])),
     z_string:trim(Text2).
 
 %% @doc Truncate a string to a max length, map control characters to spaces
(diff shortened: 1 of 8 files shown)


apps/zotonic_mod_search/src/mod_search.erl
```diff
@@ -783,8 +783,16 @@ search(_, _, _, _) ->
 trim(undefined, _Context) -> <<>>;
 trim(S, _Context) when is_binary(S) -> z_string:trim(S);
 trim(#trans{} = Tr, Context) -> trim(z_trans:lookup_fallback(Tr, Context), Context);
+trim(S, Context) when is_list(S) -> trim(characters_to_binary(S), Context);
 trim(S, Context) -> trim(z_convert:to_binary(S), Context).
 
+characters_to_binary(S) when is_list(S) ->
+    case unicode:characters_to_binary(S) of
+        {incomplete, B, _} -> z_string:sanitize_utf8(B);
+        {error, B, _} -> B;
```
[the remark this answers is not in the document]
mworrell: Fixed in later commit.

apps/zotonic_core/src/support/z_pivot_rsc_job.erl
```diff
@@ -417,17 +418,20 @@ to_integer(Text) ->
         Text1 -> z_convert:to_integer(Text1)
     end.
 
+%% @doc Preprocess a text before making in into a TSV. Ensure the text is valid utf8, replace newlines
+%% and tabs with a space, and trim the resuling text.
 -spec cleanup_tsv_text(binary()) -> binary().
 cleanup_tsv_text(Text) when is_binary(Text) ->
     Text1 = z_string:sanitize_utf8(Text),
-    Text2 = iolist_to_binary(re:replace(Text1, <<"[ ",13,10,9,"/-]+">>, <<" ">>, [global])),
+    Text2 = iolist_to_binary(re:replace(Text1, <<"[ ",13,10,9,"/–]+">>, <<" ">>, [global])),
```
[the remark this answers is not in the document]
mworrell: Made into binary, added /utf8 for the mdash;

mmzeeman (approved):
It is easy to change the way the transliteration/normalization is done. Should it be configurable?

mworrell:
Ok, we can first try this, and maybe later introduce a switch to choose the normalization.

mworrell:
@vkatsuba I am running into the problem that Київ transliterates to Kiiv.  Which makes it quite hard to find when searching for Kyiv. 

The cause is that we use a generic Cyrillic to Ascii transliteration, and not the specific Ukrainian one.

Do you have any ideas?  I am a bit hesitant to make a language dependent transliteration, though it is possible. Another option is to pull the full text index through "sounds as", my guess is that Kiiv and Kyiv will then match each other.

vkatsuba (approved):

vkatsuba:
> @vkatsuba Yes, we can add these common names in the transliterate functions, especially if we first have lowercased the strings. I will have a look at the routines in zotonic_stdlib.

Thanks!

mworrell:
The new transliteration with word mappings is in https://github.com/zotonic/z_stdlib/pull/111

apps/zotonic_mod_search/src/mod_search.erl
```diff
@@ -814,14 +839,18 @@ to_tsquery(Text, Context) when is_binary(Text) ->
...
 to_tsquery(Text, Context) when is_list(Text) ->
-    to_tsquery(z_convert:to_binary(Text), Context).
-
+    to_tsquery(characters_to_binary(Text), Context).
 
 to_tsquery_1(Text, Context) when is_binary(Text) ->
     Stemmer = z_pivot_rsc:stemmer_language(Context),
-    [{TsQuery}] = z_db:q("select plainto_tsquery($2, $1)", [z_pivot_rsc:cleanup_tsv_text(Text), Stemmer], Context),
+    Text1 = cleanup_text(Text),
+    [{TsQuery}] = z_db:q("select websearch_to_tsquery($2, $1)", [Text1, Stemmer], Context),
```
pasqu4le, line 860: @mworrell have you considered making different search query terms instead?

I'm thinking of letting users decide which of the `tsquery` building mechanisms to use based on their needs, both for flexibility as well as to avoid breaking changes.

We could for example keep `"text"` to perform plaintext queries (using `mod_search:to_tsquery`/`plainto_tsquery`) and `"webtext"` to run them with `websearch_to_tsquery` by adding another function, let's say `mod_search:to_websearch_tsquery`.

This would also allow sites to use different mechanisms in different places, e.g. one might want to use web search in the public search page but only use plaintext in a small search dialog.


On a smaller note (because I suspect not many other people have done this) I would also prefer that as it would not change the behaviour of `mod_search:to_tsquery`, which I've used elsewhere to try out alternative search logics, although I could update that as well.

I'm a bit more concerned about user experience however: existing sites will now behave differently on existing search setups as `websearch_to_tsquery` interprets some quotes and symbols as operators.

apps/zotonic_core/src/support/z_pivot_rsc_job.erl
```diff
@@ -472,20 +476,25 @@ get_pivot_title(Id, Context) ->
...
+                undefined -> First;
+                En -> En
+            end;
+        T when is_binary(T) ->
+            T;
+        _ ->
+            <<>>
     end,
-    z_string:trim(z_string:to_lower(Title)).
+    z_string:trim(z_string:normalize(Title)).
```
pasqu4le, line 501: I've used a similar approach to the one mentioned above to support accent-agnostic searches in some sites before: I made a new search term that unaccents the input query before handling it like any other `"text"` search query. 

However the issue with that is that I have to `unaccent` the other side of the equation too, be it a (custom) pivot or the content of pivot templates, to be able to use it.

I have some concerns in making that the default (I'm not sure how that affects [the parsing of text of `to_tsvector`](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-DOCUMENTS) and I suppose are usecases for differentiating between accents), but now I'm thinking: couldn't this also be left to the individual application/site?

What do you think about making a "postprocessing" notification for indexed text, so that one could pick transformations to apply before they are given to `to_tsvector` and saved in the DB?

I'm not against the normalization by default, but if you agree with the above then I'd do this with an observer in `mod_search` itself, so that it can be skipped/overridden.

mworrell: I am ok with an optional custom normalisation step, defaulting to the one in this merge request.  Of course that has to be done in the pivot, facet and search code.

apps/zotonic_mod_search/src/mod_search.erl
mworrell: It could be an option in the passed query term.  Much like we also pass operators, predicates etc. in the query terms.

mworrell: Actually, I suspect that most searches do not have any quotes, And if they have, then people might not be too surprised that the behavior will align more with what search engines do.

apps/zotonic_core/src/support/z_pivot_rsc_job.erl
mworrell: Shouldn't this be a configuration?

Also, maybe a special full text  field type for facets that does not do the normalization at all, but treats all text as-is.

apps/zotonic_mod_search/src/mod_search.erl
```diff
@@ -828,14 +853,18 @@ to_tsquery(Text, Context) when is_binary(Text) ->
...
     fixup_tsquery(z_convert:to_list(Stemmer), append_wildcard(Text, TsQuery)).
```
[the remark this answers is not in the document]
mworrell: @copilot Disable wildcard expressions if the last term is not a lexeme token.

apps/zotonic_core/src/support/z_pivot_rsc_job.erl
mworrell: I have added a convenience function `z_search:normalize_value/4` which does a notification, with a fallback to the `z_string:normalize/1` for text values.

apps/zotonic_mod_search/src/mod_search.erl
```diff
@@ -849,7 +874,11 @@ append_wildcard(_Text, <<>>) ->
...
 append_wildcard(Text, TsQ) ->
-    case is_wordchar(z_string:last_char(Text)) of
+    % Only append :* if both the input text ends with a word character
+    % and the tsquery ends with a lexeme token (i.e. ends with $').
+    % websearch_to_tsquery can return parenthesized expressions like
+    % ('foo' & 'bar') which would produce invalid syntax if :* is appended.
+    case is_wordchar(z_string:last_char(Text)) andalso binary:last(TsQ) =:= $' of
         true -> <<TsQ/binary, ":*">>;
         false -> TsQ
     end.
```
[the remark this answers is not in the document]
mworrell: @copilot apply changes based on [this feedback](https://github.com/zotonic/zotonic/pull/4211#discussion_r3093333010)

pasqu4le (approved):
Very nice, LGTM now 👍

PR #4219, opened 2025-09-30
db: Also handle the case where busy_pid is set

### LINKED ISSUE #4218: Function clause 500 error in z_db_pgsql:disconnect_1
Function clause 500 error in z_db_pgsql:disconnect_1

```
Reason: function_clause: 
     z_db_pgsql:disconnect_1({state,<0.20466.124>, ..
```

In the code I see that disconnect_1 can't handle the situation where a wait_pid is set.

mmzeeman:
### Description

Fix #4218 

Handles the case where `busy_pid` is set.
Fixes a problem with `fetch_con` call returning an answer which did not fit the `gen_server:handle_call` spec.
When `fetch_conn` fails because the worker is shutdown, now the correct error is returned.
Removed some unneeded `is_alive` checks.

### Checklist

- [ ] documentation updated
- [x] tests added
- [x] no BC breaks


mmzeeman:
@mworrell Do you know a way to test this error? BTW, why didn't dialyzer catch this?

PR #4229, opened 2025-10-23
mod_admin: refactor list view title, add include for showing extra flags for each resource

mworrell:
### Description

Refactor the list title of the normal and and media overview pages to a separate template.

This integrates with https://github.com/zotonic/zotonic_mod_presence to show who is currently editing a resource.

Accompanying request on zotonic_mod_presence: https://github.com/zotonic/zotonic_mod_presence/pull/1

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_mod_admin/priv/templates/_admin_overview_list_title.tpl
+++ b/apps/zotonic_mod_admin/priv/templates/_admin_overview_list_title.tpl
@@ -3,20 +3,5 @@
 {% endblock %}
 
 {% block title_flags %}
-    {% if id.is_protected %}
-        &nbsp; <span class="fa fa-lock text-muted" title="{_ Protected, not deletable _}"></span>
-    {% endif %}
-    {% if id == 1 or id.is_a.meta or id.content_group_id.name == 'system_content_group' %}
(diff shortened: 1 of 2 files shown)


apps/zotonic_mod_admin/priv/templates/_admin_overview_list_title.tpl
```diff
@@ -0,0 +1,22 @@
...
+            {{ id.name|default:_"system content" }}
+        </span>
+    {% elseif id.name %}
+        <span class="label label-default pull-right hidden-xs">
+            {{ id.name }}
+        </span>
+    {% endif %}
+    {% if m.identity[id].username as username %}
+        <span class="label label-info pull-right hidden-xs">{{ username | escape }}</span>
+    {% endif %}
```
pasqu4le: I'd say that these are flags too, what do you think about moving them to a new `_admin_overview_list_flags.tpl` in `mod_admin` itself?

It would get included by the `all catinclude` just below as well

mworrell: Yes, let's try

PR #4230, opened 2025-10-24
mod_development: add function tracing

mworrell:
### Description

Add a function tracing page to mod_development.

The authentication dispatch has now a flag to prevent tracing and process introspections. This is done by adding a flag `sensitive` to the controller options.

In the development tools it is now possible to set a trace on a module and (optionally) a function. The trace is then output to the browser.

TODO:

 - [x] Add config flag for non development environments to enable tracing
 - [x] Add `bin/zotonic` command to allow or disallow tracing

Example:

Screenshot 2025-10-29 at 18 13 56

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


mmzeeman (approved):
Good idea to put it here.

PR #4237, opened 2025-11-12
core: enable memo cache on all requests

mworrell:
### Description

This makes Zotonic use the memo and in-process cache more often.
Which is particularly useful for resource property fetching and ACL checks.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_acl.erl
+++ b/apps/zotonic_core/src/support/z_acl.erl
@@ -158,7 +158,7 @@ is_allowed_prop(Action, Object, Property, Context) when is_atom(Property) ->
     is_allowed_prop(Action, Object, atom_to_binary(Property, utf8), Context);
 is_allowed_prop(Action, Object, Property, Context) ->
     case z_memo:get({rsc_prop_visible, Action, Object, Property}, Context) of
-        {ok, IsAllowed} ->
+        IsAllowed when is_boolean(IsAllowed) ->
             IsAllowed;
         undefined ->
             IsAllowed = case z_notifier:first(#acl_is_allowed_prop{ action = Action, object = Object, prop = Property }, Context) of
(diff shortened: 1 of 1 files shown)


apps/zotonic_core/src/support/z_acl.erl
```diff
@@ -103,7 +103,7 @@ is_allowed(_Action, _Object, #context{ user_id=?ACL_ADMIN_USER_ID }) ->
 is_allowed(link, Object, Context) ->
     is_allowed(insert, #acl_edge{subject_id=Object, predicate=relation, object_id=Object}, Context);
 is_allowed(Action, Object, Context) ->
-    case maybe_allowed(Action, Object, Context) of
+    case maybe_allowed_memo(Action, Object, Context) of
```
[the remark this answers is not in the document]
mworrell: This is actually correct, it allows to skip some redundant checks.

mmzeeman (approved):

apps/zotonic_core/src/support/z_acl.erl
```diff
@@ -144,10 +157,17 @@ is_allowed_prop(_Action, _Object, _Property, #context{ user_id = ?ACL_ADMIN_USER
...
+            IsAllowed;
+        undefined ->
+            IsAllowed = case z_notifier:first(#acl_is_allowed_prop{ action = Action, object = Object, prop = Property }, Context) of
+                undefined -> true; % Note, the default behaviour is different for props!
+                true -> true;
+                false -> false
+            end,
+            z_memo:set({rsc_prop_visible, Action, Object, Property}, {ok, IsAllowed}, Context),
+            IsAllowed
     end.
```
pasqu4le, line 170: Nitpick, but IIUC with the keys used in `maybe_allowed_memo` it's necessary to wrap the result in a tuple with `ok` to be able to distinguish a cache miss from a hit with value `undefined`.

Here however, when a value is set, it's always a boolean, so that could be simplified away:
```suggestion
    case z_memo:get({rsc_prop_visible, Action, Object, Property}, Context) of
        IsAllowed when is_boolean(IsAllowed) ->
            IsAllowed;
        undefined ->
            IsAllowed = case z_notifier:first(#acl_is_allowed_prop{ action = Action, object = Object, prop = Property }, Context) of
                undefined -> true; % Note, the default behaviour is different for props!
                true -> true;
                false -> false
            end,
            z_memo:set({rsc_prop_visible, Action, Object, Property}, IsAllowed, Context)
    end.
```

mworrell: Good catch!

PR #4322, opened 2026-02-20
Replace Bootstrap 3 with Bootstrap 5 and compatibility shims

mworrell:
### TODO

 - [x] Rename selected theme localstorage key to non-admin name
 - [x] Store theme name as serialized JSON, so that it is compatible with the localStorage model
 - [x] Move theme init javascript to set document attributes to mod_base templates

### Description

This pull request introduces Bootstrap 5 assets to the `zotonic_mod_admin` module, enabling the admin interface to use Bootstrap 5 while preserving compatibility with the existing Bootstrap 3 build. The update includes source files, build scripts, compatibility shims, and appropriate licensing documentation.

The most important changes are:

**Bootstrap 3 fixes**

* Patches for CVEs reported against the Bootstrap 3.4.1 Javascript.

**Bootstrap 5 Integration and Build System:**
* Added Bootstrap 5 SCSS source files and a new `Makefile` for building admin styles, producing both regular and minified CSS outputs for the admin interface. (`priv/lib-src/admin-bootstrap5/bootstrap/scss`, `priv/lib-src/admin-bootstrap5/Makefile`, `priv/lib/css/admin-bootstrap5.css`, `priv/lib/css/admin-bootstrap5.min.css`) [[1]](diffhunk://#diff-a4618ef8f8190129bdd1b3fc8fa556b25b509ba7caca27e3c8adea45fb70b499R1-R153) [[2]](diffhunk://#diff-8e1f65b6c2564564f6eae85e22b6716c17fe924816cee9776eec452180bffbd4R1-R14)
* Created a custom SCSS entrypoint (`admin-bootstrap.scss`) that maps legacy Bootstrap 3 variables to Bootstrap 5 equivalents, imports Bootstrap 5, applies compatibility shims, and includes Zotonic admin overrides. (`priv/lib-src/admin-bootstrap5/admin-bootstrap.scss`)

**Compatibility and Documentation:**
* Added Bootstrap 3 compatibility shims to ensure legacy admin templates continue to work with the new Bootstrap 5 build. (`priv/lib-src/admin-bootstrap5/compat`)
* Updated module documentation to explain the inclusion of Bootstrap 5 assets, the build process, compatibility layers, and licensing information. (`README.md`)
* Included Bootstrap's MIT license and official README for compliance and reference. (`BOOTSTRAP-LICENSE`, `BOOTSTRAP-README.md`) [[1]](diffhunk://#diff-3187308d7d6bc7c9e8487d10f3f2f405ab3b9a3729e5440d79d652451b4fea2aR1-R21) [[2]](diffhunk://#diff-7a85df815bfeaf98b3b342687c5e2b7562691742f00d499a1998b712a08c301bR1-R246)

### Checklist

- [x] documentation updated
- [ ] tests added
- [x] no BC breaks


EdwardAbboud:
Code changes look good to me, also testing things locally went well. So far i encounter one issue which i'm still debugging the source of but i only encounter it on this branch which is the login form. I get some jQuery errors

```
Uncaught TypeError: $.widgetManager is not a function
```
and

```
Uncaught SyntaxError: Invalid or unexpected token (at livevalidation-1.3~z.dialog~jquery.loadmask~3009626650.js:7168:1)
```

I'll report more if i find the fix/source

apps/zotonic_mod_bootstrap/priv/lib-src/bootstrap/compat/_bootstrap3-compat.scss
```diff
@@ -0,0 +1,1213 @@
...
+
+.embed-responsive-16by9 { padding-bottom: 56.25%; }
+.embed-responsive-4by3 { padding-bottom: 75%; }
+.embed-responsive-21by9 { padding-bottom: 42.8571428571%; }
+.embed-responsive-1by1 { padding-bottom: 100%; }
+
+// Glyphicons (from Bootstrap 3)
+@font-face {
+  font-family: "Glyphicons Halflings";
+  src: url("../fonts/glyphicons-halflings-regular.eot");
```
karlsson: Replace ".." with "/lib/bootstrap" in all urls in the the font family.
Seems you get 404 Not Found otherwise when trying to load the fonts.

mworrell: Good catch!

mworrell:
@EdwardAbboud On which pages do you see those errors? Ah, I see, the login form. I'll check as well.

**Update** Checked on my test site, and there I don't see any errors.

mworrell:
@mmzeeman We store the selected theme in a key specific for the admin, do you think we should just make it a zotonic standard key?  So that we select the theme for both the front- and the backend at the same time.

apps/zotonic_mod_bootstrap/priv/lib-src/bootstrap/compat/_bootstrap3-compat.scss
karlsson: Works nice now, thanks!

mmzeeman:
> @mmzeeman We store the selected theme in a key specific for the admin, do you think we should just make it a zotonic standard key? So that we select the theme for both the front- and the backend at the same time.

That sounds fine to have a single setting for the admin and the site. We can always change it later. Most users probably don't care and set it to auto anyway.

mmzeeman (approved):
Everything looks fine.. I'll run this branch to see if anything pops up. Our site was already done in bootstrap 5.3. I'm also not seeing any glitches with that.

mworrell:
@mmzeeman For the the theme selection, I have moved the init of the document attributes to templates in `mod_base`.  It is now included in `_html_head.tpl` and `_html_head_admin.tpl`

apps/zotonic_mod_base/priv/templates/_html_head.tpl
```diff
@@ -1,2 +1,3 @@
 {% include "_html_head_cotonic.tpl" %}
 {% include "_html_head_code_highlight.tpl" %}
+{% include "_html_head_theme_init.tpl" %}
```
pasqu4le, line 3: It seems that this PR affects the looks of the `zotonic_site_status` too, whose style I think should be adjusted:
Image
Note: I have the dark theme in my system settings

pasqu4le: Minor comment for the ACL rules' page as well:
Image
deny rules no longer have the red background (at least on the dark theme) and there's some odd vertical lines in the table that didn't use to be there.

pasqu4le: Modal header's close button has poor contrast on dark theme:
Image

apps/zotonic_mod_base/priv/templates/_html_head_theme_init.tpl
```diff
@@ -0,0 +1,19 @@
+{# Sets `document[data-zotonic-theme]` to the user-selected theme, and `document[data-bs-theme]`
+ # to the resolved theme (which is either the user-selected theme or the system theme if the
+ # user-selected theme is "auto")
+ #}
+<script nonce="{{ m.req.csp_nonce }}">
+    (function() {
+        let theme = "auto";
+        try {
+            theme = JSON.parse(localStorage.getItem("zotonic-theme")) || "auto";
```
pasqu4le: I've tried to ignore the system settings and force the light theme by running:
```
localStorage.setItem("zotonic-theme", "light");
```
but this doesn't seem to make a difference.

Only then I noticed the `JSON.parse`, which AFAICT fails on a string.
Am I missing something? How does one select a theme? 🤔

mworrell: In the localStorage model we use JSON to store data. In that way we can store arbitrary data and not just single values.

```javascript
localStorage.setItem("zotonic-theme", '"light"');
```

BTW I've add a Zotonic configuration to fix the ui_theme.

apps/zotonic_mod_base/priv/templates/_html_head.tpl
mworrell: Except for those vertical lines in the `th`, these css issues should be better now.

PR #4431, opened 2026-05-31
[ACL] Expanded ACL rules in the database

pasqu4le:
### Description

_NOTE: this is marked as a draft as it's still in progress (see TODOs in the code), but I wanted to ask for some feedback._
_I've left all the intermediate commits in this PR, which might help with the review_

This PR proposes a change in the internals of the `acl_user_groups` mod, replacing the ETS table used for fast access to "expanded" rules with a combination of database tables and cache entries.

The rationale behind this is that in order to query the database for visible resources (as it happens in searches) the ETS table approach forces us to expand the rules and build a query with them, which ends up potentially having a lot of clauses and parameters (depending on the rules installed).

If instead ACL rules (rsc and collab) were "digested" and expanded in the database, then this could be queried using such expanded rules in `JOIN`/`EXIST`s clauses, which are generally better handled by the query planner, can take full advantage of indexes and thus should scale better with more complex ACL rulesets.

For non-SQL checks, the ETS table is faster than accessing the database every time, so I've added caching (based on `z_depcache`) as well.

@mworrell @mmzeeman the recent #4422 reminded me of this idea and I couldn't help myself from giving it a shot.
I've not tested it extensively yet, but it seems like this could be a working replacement.
I'd love to have your opinion and/or comments.

### Checklist

- [ ] documentation updated
- [ ] tests added
- [ ] no BC breaks


mmzeeman:
Interesting approach, instead of using Erlang as a ACL interpreter you move to SQL, and cache the result.

Last week I half jokingly said to @mworrell that another approach might be compiling the ACL rules to erlang and use the compiled modules to check access. 

One unique aspect of Zotonic is that it checks access on a property level, so it should be fast. The last couple of years I've been looking at approaches to make creating ACLs more flexible, auditable and fast. It is probably something other Erlang projects struggle with.

The auditablility is also important, which is hard to do if the access control decisions are coded in Erlang, because then you have to find somebody which knows about the web, access control and also knows Erlang.

The user group acl module is nice, but in practice we find that we need to add additional acl modules for flexibility, because the acl rules are kind of hard coded into it. This can make deciphering acl decisions hard.

I've looked at Open Policy Agent (https://www.openpolicyagent.org), and other logic based approaches (cuelang.org is very interesting). These are usually too slow for the fine grain access control we want. Another problem is that moving the access control to an entirely new language also doesn't help with audit-ability.

pasqu4le:
> The auditablility is also important

Agreed, I didn't do anything to improve that yet, but with the approach in this PR I was also thinking that we can track which ACL rule resulted in which "expanded" rule, thus trace back permissions for debugging purposes.

> The user group acl module is nice, but in practice we find that we need to add additional acl modules for flexibility [...] This can make deciphering acl decisions hard.

I wonder it we can make this more auditable/traceable as well 🤔 

ACL permissions are requested through notifications, which now only return either `true`, `false` or `undefined`.
However, if we accepted additional data back from the answering module, then we could trace the source of the decision.

In the case of the ACL user group module, this would point at the ID of the ACL rule in the database that resulted in a (non-`undefined`) answer. Other modules would have their own mechanisms and "tracing info".

For debugging purposes we could then also add a form in the "development" admin page, where given a user-action-object combination, we return said "tracing info".

PR #4447, opened 2026-06-23
[Media] Add sha-256 hash digest to file media for integrity checks

robvandenbogaard:
### Description

We already had a sha1 field to prepare for storing file digest hashes in medium records; these changes update the field to contain recommended SHA-256 hashes [1] [2]. Hashes are calculated using Zotonic's `z_crypto:hex_sha2_file/1` function after upload and preprocessing, just before the medium is inserted in `m_medium:replace_file_db/5`, and shown in the admin edit page of media resources.

As the existing `sha1` medium field was not used yet, and a db migration is included (automatically performed from `z_install:upgrade/3`) no backward compatibility breaks are expected.

[1] https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
[2] https://www.nist.gov/news-events/news/2022/12/nist-retires-sha-1-cryptographic-algorithm

### Checklist

- [ ] documentation updated
- [ ] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/install/z_install_update.erl
+++ b/apps/zotonic_core/src/install/z_install_update.erl
@@ -267,16 +267,6 @@ get_column_type(C, Table, Column, Database, Schema) ->
               and column_name = $4", [Database, Schema, Table, Column]),
     ColumnType.
 
-get_column_type_and_length(C, Table, Column, Database, Schema) ->
-    {ok, _, [{ColumnType, CharacterMaximumLength}]} = epgsql:equery(C, "
-            select data_type, character_maximum_length
-            from information_schema.columns
-            where table_catalog = $1
-              and table_schema = $2
-              and table_name = $3
-              and column_name = $4", [Database, Schema, Table, Column]),
-    {ColumnType, CharacterMaximumLength}.
-
 is_column_nullable(C, Table, Column, Database, Schema) ->
     {ok, _, [{IsNullable}]} = epgsql:equery(C, "
             select is_nullable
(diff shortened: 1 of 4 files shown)


apps/zotonic_core/src/install/z_install_update.erl
```diff
@@ -1092,6 +1106,29 @@ medium_size_bigint(C, Database, Schema) ->
...
+                {<<"character varying">>, 64} ->
+                   ok;
+                _ ->
+                    {ok,[],[]} = epgsql:squery(C, "alter table medium alter column digest type character varying(64)"),
+                    ok
+            end;
+        false ->
+            {ok, [], []} = epgsql:squery(C, "alter table medium add column digest character varying(64)"),
+            ok
+    end.
```
pasqu4le, line 1128: Checking and comparing with the other updates in this module I noticed that others do a `?LOG_NOTICE` when adding/removing a column, I think it would be nice to do this here too so to have that info in the logs.

Aside from that, as for the `digest` column itself, I have a few questions:
1. why did you go for a `character varying(64)`? After a quick search it seems that [postgres itself has sha256 functions](https://www.postgresql.org/docs/16/functions-binarystring.html) and that those return a `bytea`. If that can be used then I think it'd be preferable, as it would allow us to use related hashing functions in queries;
2. was the `digest` column already used in the past? I didn't see it before and I'm wondering if it wouldn't be ok to skip the check on its length when it is found to be there (especially if we go for a `bytea`)

robvandenbogaard: Primary reason to make it a `character varying(64)` is that the previously existing `sha1` field also was a `character varying`, and as `z_crypto:hex_sha2_file` outputs a hex encoded binary string, it was not directly on my radar to make it a binary field. Would be good for minimising storage requirements.
The field was not there yet, so indeed the check would not be necessary, though I like the check to not assume it (reasoning that the complete type includes the length).

robvandenbogaard: Covered in new commit.

robvandenbogaard:
@mworrell do you also agree on using a `bytea` field for the hash digest?

pasqu4le (approved):
LGTM now, but would also like @mworrell's opinion before merging 👌

mworrell:
I am not sure about a `bytea` for this, as with a text field we can:

* easily verify the hash (visually)
* add a partial index on the first hex digits, without an exploding index
* directly display the value (as we do in the medium details)
* easily use the value for integration with external systems that use the same hashing method

IMHO, the space efficiency is not very significant, considering we also store EXIF data and long filenames.

PR #4478, opened 2026-08-07
core: always deny anonymous users update actions

mworrell:
### Description

This fixes an issue where a wrongly configured ACL could allow (automated) anonymous users to insert/update content.

I also considered to set the max upload size to 0, but that would disable any upload of any image in an anonymous form, which we don't want.

If you do want to let an anonymous user create content, then add a "anonymous" person/bot with the correct access control to the system. Sudo as this person/bot to make the changes. As this can be gated by logic, it will prevent automated creation or updates of resources by anonymous bots.

---

This pull request updates the access control logic in `z_acl.erl` to strengthen restrictions on anonymous users and adds a new utility function for detecting anonymous contexts. The main changes are:

**Access Restrictions for Anonymous Users:**

* Anonymous users (those with `user_id = undefined`) are now explicitly denied all update actions in both `maybe_allowed/3` and `is_allowed_prop/4`. This ensures that only authenticated users can perform updates, improving security. [[1]](diffhunk://#diff-17bf4e593771a8d75b1224dcd2491411b38307331d2f74980267638bb97d0e2aR129-R131) [[2]](diffhunk://#diff-17bf4e593771a8d75b1224dcd2491411b38307331d2f74980267638bb97d0e2aR161-R162)

**Utility Function for Anonymous Detection:**

* Added a new function `is_anonymous/1` to check if the current context is for an anonymous (non-sudo) user. This function is now exported and can be used elsewhere in the codebase. [[1]](diffhunk://#diff-17bf4e593771a8d75b1224dcd2491411b38307331d2f74980267638bb97d0e2aR46) [[2]](diffhunk://#diff-17bf4e593771a8d75b1224dcd2491411b38307331d2f74980267638bb97d0e2aR350-R359)

**Other:**

* Updated copyright years to 2026.

### Checklist

- [ ] documentation updated
- [x] tests added
- [ ] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/support/z_acl.erl
+++ b/apps/zotonic_core/src/support/z_acl.erl
@@ -1,9 +1,9 @@
 %% @author Marc Worrell <marc@worrell.nl>
-%% @copyright 2010-2024 Marc Worrell
+%% @copyright 2010-2026 Marc Worrell
 %% @doc Access control for Zotonic.  Interfaces to modules implementing the ACL events.
 %% @end
 
-%% Copyright 2010-2024 Marc Worrell
+%% Copyright 2010-2026 Marc Worrell
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
@@ -43,6 +43,7 @@
          is_admin_editable/1,
          is_admin/1,
          is_sudo/1,
+         is_anonymous/1,
          sudo/1,
          sudo/2,
          anondo/1,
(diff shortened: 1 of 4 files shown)


apps/zotonic_core/src/support/z_acl.erl
```diff
@@ -122,9 +123,15 @@ maybe_allowed(_Action, _Object, #context{ acl = admin }) ->
 maybe_allowed(UpdateAction, _Object, #context{ acl_is_read_only = true }) when ?is_update_action(UpdateAction) ->
     % Read-only context - maybe from an OAuth token
     false;
-maybe_allowed(_Action, _Object, #context{user_id = ?ACL_ADMIN_USER_ID}) ->
+maybe_allowed(_Action, _Object, #context{ user_id = ?ACL_ADMIN_USER_ID}) ->
```
robvandenbogaard:
```suggestion
maybe_allowed(_Action, _Object, #context{user_id = ?ACL_ADMIN_USER_ID}) ->
```

robvandenbogaard (changes requested):

robvandenbogaard (approved):

mmzeeman (approved):

PR #4483, opened 2026-08-13
zotonic_launcher: fix running on OpenBSD without requiring bash

### LINKED ISSUE #4482: zotonic_launcher: fix running on OpenBSD without requiring bash
## `Zotonic` version
[1.0.0-rc.17](https://github.com/zotonic/zotonic.git)

## `OS` version
OpenBSD 7.9 amd64 (`/bin/sh` = pdksh v5.2.14)

## Description
* **Motivation**
  `zotonic_launcher` should run on OpenBSD without requiring bash, which is
  not part of the OpenBSD base system.
* **Proposal**
  1. Add `OpenBSD*)   machine=BSD;;` to the `uname` case in
     `apps/zotonic_launcher/bin/zotonic`.
  2. Replace the non-POSIX `>|` clobber redirect with `>`.

## Current behavior
* The `uname` case has no `OpenBSD*)` branch, so `machine=UNKNOWN:OpenBSD`
  and the `completion` command's `BASH_FILE` is never set. Combined with the
  unquoted `$BASH_FILE` tests (`[ -n $BASH_FILE ]` / `[ -f $BASH_FILE ]`), an
  empty variable expands to single-argument tests that are always true, so the
  branch is entered with an empty filename, causing redirect errors and a
  stray `bash -c "source "` call.
* The `>|` clobber redirect is not POSIX (bash/ksh-only).

## Expected behavior
* OpenBSD is recognized (`machine=BSD`), so `BASH_FILE` is set to `~/.bashrc`
  and the empty-variable bug is never triggered; the `completion` command
  works or degrades gracefully.
* The launcher uses POSIX syntax only (`>`, not `>|`) and runs under
  `#!/usr/bin/env sh`.

## Config
None. The `completion` feature intentionally generates a bash completion file
and invokes `bash -c`, so it still requires bash to be installed - that is by
design.

kasicass:
### Description

Fix #4482 

- Add OpenBSD*) branch to the uname case, so machine=BSD and the completion command gets a valid BASH_FILE on OpenBSD
- Replace the non-POSIX '>|' clobber redirect with '>'

### Checklist

- [x] no BC breaks

### DIFF
--- a/apps/zotonic_launcher/bin/zotonic
+++ b/apps/zotonic_launcher/bin/zotonic
@@ -19,6 +19,7 @@ case "${unameOut}" in
     Linux*)     machine=Linux;;
     Darwin*)    machine=Mac;;
     FreeBSD*)   machine=BSD;;
+    OpenBSD*)   machine=BSD;;
     DragonFly*) machine=BSD;;
     CYGWIN*)    machine=Cygwin;;
     MINGW*)     machine=MinGw;;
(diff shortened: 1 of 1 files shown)


mworrell:
@kasicass Thank you!

PR #4498, opened 2026-08-28
Dispatcher: sanitize UTF8 characters after unescaping (split) path

pasqu4le:
### Description

This fixes an issue where dispatching an invalid path creates a `#dispatch` notification with non-UTF8 characters in its `path`.

Example: the path  `<<"/page/12345/page-slug/%00http">>`, given to `z_sites_dispatcher:dispatch_path` will send a notification for a `#dispatch` with `path = <<47,112,97,103,101,47,49,50,51,52,53,47,112,97,103,101,45,115,108,117,103,47,0,104,116,116,112>>`.

This causes errors in modules such as `mod_alternative_uris` as those characters are unexpected.

The issue seems to occur because in `z_sites_dispatcher:do_dispatch_fail` we recombine a path from tokens with `tokens_to_path`, but those tokens can contain invalid utf8 characters, because they were first `unescape`d in `split_path` and not checked again afterwards.

### Checklist

- [ ] documentation updated
- [x] tests added
- [x] no BC breaks

### DIFF
--- a/apps/zotonic_core/src/models/m_rsc.erl
+++ b/apps/zotonic_core/src/models/m_rsc.erl
@@ -417,6 +417,7 @@ name_to_id_cat(Name, Cat, Context) ->
 %% resource.
 %% The page path is normalized by ensuring that the path starts with
 %% a single "/" and removing trailing "/" characters.
+%% Invalid UTF-8 and characters in the range 0..31 are rejected before querying.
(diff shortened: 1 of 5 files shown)


mworrell (approved):

mworrell:
We might also do a quick check, and refuse the path if it is not UTF-8, as we only make UTF-8 paths.

